PR #9, opened 2018-07-03
Avoid unnecessary files in dist / release files

raphaelstolt:
I took the liberty to exclude the `LICENSE.md` file as it isn't really needed in dist files. Your position may differ on that.


nunomaduro:
I agree with you. Thanks!

PR #12, opened 2018-07-05
Add option --path & --memory-limit

sangnguyenplus:
Example:
```
php artisan code:analyse --path=packages/comment/src --memory-limit=120M
```

I'm working on package development so I need to have the option to change path and some packages are very big so sometimes it has an error.

```
Fatal error: Allowed memory size of 10485760 bytes exhausted (tried to allocate 131072 bytes) in /Users/mac/workspace/cms/vendor/phpstan/phpstan/src/Analyser/NodeScopeResolver.php on line 675
```

`--memory-limit` will resolve it.

Thank you!

### DIFF
--- a/src/Console/CodeAnalyseCommand.php
+++ b/src/Console/CodeAnalyseCommand.php
@@ -13,9 +13,7 @@ declare(strict_types=1);
 
 namespace NunoMaduro\LaravelCodeAnalyse\Console;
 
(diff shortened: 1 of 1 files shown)


nunomaduro:
The latest version already have those options. Thanks anyway! Please don't forget that this package is under development. Is not stable yet.

PR #55, opened 2018-07-20
Go straight to binary instead of relying on composer parser

deleugpn:


ForzaSFerrari:
It works for me on Windows in Homestead. @deleugpn

PR #60, opened 2018-07-21
Fix styles

szepeviktor:
@nunomaduro For the previous two comments


nunomaduro:
@szepeviktor Thanks! Already got fixed in https://github.com/nunomaduro/larastan/pull/62.

PR #69, opened 2018-07-22
small change in readme

ForzaSFerrari:


szepeviktor:
Hello!

Just above the change:

> to install Larastan as a development dependency

nunomaduro:
@ForzaSFerrari Thanks for the proposition. 👍

As @szepeviktor said, Larastan should be only installed as dev dependency.

Issue #56, opened 2018-07-20
Xampp on Windows


ForzaSFerrari:
With the latest update, Larastan finally works in Homestead. Unfortunately, it doesn't work anymore with Xampp on Windows. The result, what I get is. 

```
'phpstan' is not recognized as an internal or external command,
operable program or batch file.
```

deleugpn:
I'm going to try to pick this up next week.

boomhq:
Hi,  i've found 2 solutions but i dont know what is the best way.
The bug is a regression from :  [COMMIT](https://github.com/nunomaduro/larastan/commit/08dcba3740bc24e5f1e8ac8e530170717c49478c#diff-fd1a7c19da3313d9dcfb80ffadb016bb
)
1st solution: Back to "vendor/bin" and work on LINUX/Windows

2nd solution: on CodeAnalyseCommand.php  on 

> private function command()

 replace with :

`   private function command(): string
    {
        $command = Artisan::phpBinary();
        return "$command phpstan analyse";
    }
`

On this case the PHP binary is alwways resolved.
Maybe you can help me with your opinion and i'm ready for PR

ForzaSFerrari:
@boomhq The first solution would not work with Laravel's Homestead. Look at #35 

Your second solution does not seem to work for me on Windows. Larastan does not do anything.

boomhq:
@ForzaSFerrari it's strange for me in classique laravel app (no homestead, valet ...)  on linux and windows with         `$process = new Process($this->cmd(), $this->laravel->basePath('vendor/phpstan/phpstan/bin'));`

And after replace` function command()` it's work fine.

Test on Windows10 PHP 7.1.7 and Debian 9 with PHP 7.1.18

staf:
Not only related to XAMPP, seems to be a general Windows thing as this also happens running with apache and php installed "manually". Using Windows 10 and php 7.2.

It seems to try to run phpstan globally and not the binary from the vendor folder. I tried to install phpstan globally but that ended up with larastan "doubling" the path name.

Fresh Laravel app installed in ` C:\Code\larastan` results in the error `Path C:\Code\larastan\C:\Code\larastan\app does not exist`. when running the script with phpstan installed globally to allow actually running the command.

boomhq:
@staf  you can start larastan with `php artisan code:analyse --path=app`

staf:
[...]

PR #110, opened 2018-08-02
Use cross-platform value for default paths input option.

hordijk:
As mentioned in https://github.com/nunomaduro/larastan/issues/56#issuecomment-407047456 running  larastan on Windows doesn't work well without specifying the directory to analyze. 

If the directory is not specified it uses: `base_path('app')` defined in `NunoMaduro\Larastan\Console\OptionsResolver`.  The `cmd` function in `NunoMaduro\Larastan\Console\CodeAnalyseCommand` will add a basepath again (see code snippet below), unless the the path is empty or starts with DIRECTORY_SEPARATOR.  This works properly on linux due to /  and will fail on Windows since C: doesn't match the DIRECTORY_SEPARATOR.

Using 'app' as default value should work on both platforms.
Another solution could be to verify the supplied path starts with the actual basepath?

```php
 $pathsValue = $this->option('paths');
if (is_array($pathsValue)) {
	$pathsValue = current($pathsValue);
}

$paths = array_map(
	function ($path) {
		return starts_with($path, DIRECTORY_SEPARATOR) || empty($path) ? $path : $this->laravel->basePath(
			trim($path)
		);
	},
	explode(',', $pathsValue)
);
```

### DIFF
--- a/src/Console/OptionsResolver.php
+++ b/src/Console/OptionsResolver.php
@@ -74,7 +74,7 @@ final class OptionsResolver
 
         $definition->addOption(
             new InputOption(
-                'paths', 'p', InputOption::VALUE_REQUIRED, 'Paths with source code to run analysis on', base_path('app')
+                'paths', 'p', InputOption::VALUE_REQUIRED, 'Paths with source code to run analysis on', 'app'
             )
         );
 


nunomaduro:
Thanks @hordijk! I will close this, since the PR #128 also fixes this issue.

PR #115, opened 2018-08-08
Respect the exit code from phpstan and exit with the same.

j3j5:
Right now, larastan always exits with `0` regardless of the errors phpstan finds. This change exits the process with the same exit code as phpstan did, allowing a better integration of larastan on build services.

PR #117, opened 2018-08-09
Exit with process exit code on failure

pleckey:
If the process was not successful (non-zero exit code), return that exit code from the artisan call.

This allows `larastan` to be run as part of scripts that require non-zero exit codes on failure, like git pre-commit hooks to block commit unless `phpstan` passes.


pleckey:
Duplicate of #115

PR #118, opened 2018-08-11
$request->input returns array|string|null, string expected

deleugpn:
As stated in #112, Laravel's method signature for `Illuminate\Foundation\Http\FormRequest::input()` is `array|string|null`. From Laravel's perspective, this will be used together with the method `rules` which allows the developer to state the requirements for an input field (e.g. `array` or `string`).

This pull request instructs phpstan to consider the return type as dynamic instead. Since this is my first-ever contribution to a phpstan extension, I have a couple of concerns to highlight: 

- The method actually comes from `trait Illuminate\Http\Concerns\InteractsWithInput`. Should the extension target the trait or only concrete classes?
- This pull request only targets `FormRequest`, however `Illuminate\Http\Request` will have the same issue as it also uses the `InteractsWithInput` trait. Should there be one extension for each class or is there a way to have a single extension that targets both?

Any feedback is well appreciated and thanks for this wonderful tool, both larastan and phpstan!

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -36,3 +36,8 @@ services:
         tags:
             - phpstan.broker.dynamicStaticMethodReturnTypeExtension
 
+    -
+        class: NunoMaduro\Larastan\ReturnTypes\FormRequestExtension
+        tags:
+            - phpstan.broker.dynamicMethodReturnTypeExtension
+
--- a/src/ReturnTypes/FormRequestExtension.php
+++ b/src/ReturnTypes/FormRequestExtension.php
@@ -50,4 +50,3 @@ final class FormRequestExtension implements DynamicMethodReturnTypeExtension
         return new MixedType;
     }
 }
-
--- a/src/ReturnTypes/FormRequestExtension.php
+++ b/src/ReturnTypes/FormRequestExtension.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of Larastan.
+ *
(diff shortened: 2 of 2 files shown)


deleugpn:
Here's a MVP of the problem: https://phpstan.org/r/708f863bb376f24259086d30ee8be72e

By doing any string manipulation, PHPStan will complain that you might be manipulating string with a _possible_ array. Or, if you also use it in an array function, PHPStan will tell you that you're using a _possible_ string where an array is expected. By changing from `array|string|null` to `mixed`, PHPStan will not complain anymore.

ondrejmirtes:
Would it be possible to statically read the rules and find out whether the method returns array|null OR string|null?

deleugpn:
@ondrejmirtes I would say no, unless you have a valid Laravel container. The Rules method is invoked via the Container and can have any dependency in it's signature. Also, any class implementing the Rule interface can act as a ruler, which makes it harder to figure out which output will be.

Also, from the HTTP definitions, I don't think it's possible to build anything other than array or string. Since laravel will convert empty string / missing keys into `null`, that gives us only those 3 scenarios: array, string or null.

ondrejmirtes:
Maybe you should think about using BenevolentUnionType of array/string/null instead of mixed. That means that if you use it as something completely else (like `stdClass`), PHPStan will tell you, but it won't complain in most situations normal UnionType complains.

nunomaduro:
Thanks! 🏄‍♂️

nunomaduro:
@deleugpn I am open for more extensions like that.

PR #127, opened 2018-08-20
Windows Compatibility

schursin:
Latest release doesn't work correctly on Windows. So...

- Non working Windows related hack was removed.
- Process call code was refactored. Now it's cross-platform.


schursin:
breaks Homestead. closed

PR #128, opened 2018-08-21
Windows Compatibility

### LINKED ISSUE #56: Xampp on Windows

schursin:
Latest release doesn't work correctly on Windows. So...

- Non working Windows related hack was removed.
- Process call code was refactored. Now it's cross-platform.

closes #56 and not breaks #55

### DIFF
--- a/src/Console/CodeAnalyseCommand.php
+++ b/src/Console/CodeAnalyseCommand.php
@@ -121,25 +121,13 @@ final class CodeAnalyseCommand extends Command
         );
 
         $params = [
-            $this->command(),
+            Artisan::phpBinary(),
+            'phpstan',
+            'analyse',
             implode(' ', $paths),
             $options,
         ];
 
         return implode(' ', $params);
-    }
-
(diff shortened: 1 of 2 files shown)


nunomaduro:
I do confirm that this is working on Ubuntu. Can someone check on Mac OSX?

@hordijk This works for you right?

schursin:
@nunomaduro year, it works on macOS.

image

nunomaduro:
And you have tested that on non bash windows right ?

schursin:
@nunomaduro Yes, right.

nunomaduro:
Thanks!

hordijk:
The fix in OptionsResolver.php is the same as #110. I can confirm it works fine for me.

PR #137, opened 2018-09-12
Filter out non existing classes from mixins

hartman:
Mixins declared by @mixin and @see might not always be loaded classes. Exclude them from the list of mixins when they cannot be instantiated.


nunomaduro:
Thank you!

PR #145, opened 2018-09-25
Add support for request helper

j3j5:
Add support for the `request()` helper.

This is my first significant contribution to phpstan/larastan so all feedback is welcome.

### DIFF
--- a/src/ReturnTypes/Helpers/RequestExtension.php
+++ b/src/ReturnTypes/Helpers/RequestExtension.php
@@ -1,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of Larastan.
+ *
+ * (c) Julio Foulquie <jfoulquie@gmail.com>
+ *
+ *  For the full copyright and license information, please view the LICENSE
+ *  file that was distributed with this source code.
+ */
+
+namespace NunoMaduro\Larastan\ReturnTypes\Helpers;
+
+use function count;
+use PHPStan\Type\Type;
+use PHPStan\Analyser\Scope;
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\MixedType;
+use PHPStan\Type\ObjectType;
+use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+
+/**
+ * @internal
+ */
+final class RequestExtension implements DynamicFunctionReturnTypeExtension
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function isFunctionSupported(FunctionReflection $functionReflection): bool
+    {
+        return $functionReflection->getName() === 'request';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getTypeFromFunctionCall(
+        FunctionReflection $functionReflection,
+        FuncCall $functionCall,
+        Scope $scope
+    ): Type {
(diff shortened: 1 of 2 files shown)


szepeviktor:
Looks very nice. Does it work?
BTW Larastan's code is checked by PHPStan ;-)

szepeviktor:
Thank you for your contribution!

src/ReturnTypes/Helpers/RequestExtension.php
```diff
@@ -0,0 +1,59 @@
...
+        FuncCall $functionCall,
+        Scope $scope
+    ): Type {
+        // No arguments, it returns a request object from the container
+        if (count($functionCall->args) === 0) {
+            return new ObjectType(\Illuminate\Http\Request::class);
+        }
+
+        if (isset($functionCall->args[0]->value) && $functionCall->args[0]->value instanceof Array_) {
+            return new ArrayType(new MixedType(), new MixedType());
```
nunomaduro: Shouldn't this be only `return new ArrayType()` ?

src/ReturnTypes/Helpers/RequestExtension.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of Larastan.
+ *
+ * (c) Julio Foulquie <jfoulquie@gmail.com>
```
nunomaduro: Sorry, but this should contain the name of the package owner: `Nuno Maduro <enunomaduro@gmail.com>`.

j3j5: I'll change it! I wasn't sure what was the licensing policy for fully new files.

nunomaduro: @j3j5 Thanks!

src/ReturnTypes/Helpers/RequestExtension.php
j3j5: Looking at the constructor I didn't think it was possible, but I just tested it and it works. What are the values of `Type $keyType` and `Type $itemType` when you construct the object w/o arguments?

j3j5:
I've pushed to the original branch, not sure how GH will handle now that the PR is already merged. Shall I submit another one?

nunomaduro:
@j3j5 I think you can submit a new PR.

src/ReturnTypes/Helpers/RequestExtension.php
nunomaduro: When $key is an `array` the request helper returns the result of the method `only`:
```
        if (is_array($key)) {
            return app('request')->only($key);
        }
```

And the only method returns an `array`:
```php
    /**
     * Get a subset containing the provided keys with values from the input data.
     *
     * @param  array|mixed  $keys
     * @return array
     */
    public function only($keys)
```

j3j5: Yes, that I understand, that's why I returned a `PHPStan\Type\ArrayType` in the first place. But looking at the source code of that class, I interpreted (wrongly it seems) that the constructor needed 2 arguments.

```php
public function __construct(Type $keyType, Type $itemType)
{
	if ($keyType->describe(VerbosityLevel::value()) === '(int|string)') {
		$keyType = new MixedType();
	}
	$this->keyType = $keyType;
	$this->itemType = $itemType;
}
```
I looked at PHPStan's source code and I copypasted that one, thinking it made sense.

j3j5: LOL, this is what I was talking about https://travis-ci.org/nunomaduro/larastan/jobs/433160174#L650-L659

PR #148, opened 2018-09-25
Check for copyright messages

szepeviktor:


szepeviktor:
The test fails as we have a non-nunomaduro copyright in the repo!

nunomaduro:
@szepeviktor Thanks for the proposal. What it does?

nunomaduro:
@szepeviktor Just tested locally! Nice!

PR #156, opened 2018-10-04
Hide composer suggestions

szepeviktor:


nunomaduro:
Thanks buddy @szepeviktor !

PR #161, opened 2018-10-06
Handle "auth()" helper function

### LINKED ISSUE #44: Call to an undefined method user()
```php
auth()->user()
```

No error on
```php
Auth::user()
```

> Call to an undefined method Illuminate\Contracts\Auth\Factory|Illuminate\Contracts\Auth\Guard::user()

I think the code above is also often used in many different projects.

ForzaSFerrari:
Fixes #44

### DIFF
--- a/src/ReturnTypes/Helpers/AuthExtension.php
+++ b/src/ReturnTypes/Helpers/AuthExtension.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of Larastan.
+ *
+ * (c) Nuno Maduro <enunomaduro@gmail.com>
+ *
+ *  For the full copyright and license information, please view the LICENSE
+ *  file that was distributed with this source code.
+ */
+
+namespace NunoMaduro\Larastan\ReturnTypes\Helpers;
+
+use function count;
+use PHPStan\Type\Type;
+use PHPStan\Analyser\Scope;
+use PHPStan\Type\ObjectType;
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Reflection\FunctionReflection;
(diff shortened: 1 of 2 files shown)


src/ReturnTypes/Helpers/AuthExtension.php
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getTypeFromFunctionCall(
+        FunctionReflection $functionReflection,
+        FuncCall $functionCall,
+        Scope $scope
+    ): Type {
+        if (count($functionCall->args) === 0) {
+            return new ObjectType(\Illuminate\Contracts\Auth\Guard::class);
```
nunomaduro: This should be: `return new ObjectType(\Illuminate\Contracts\Auth\Factory::class);`

src/ReturnTypes/Helpers/AuthExtension.php
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        return new ObjectType(\Illuminate\Contracts\Auth\Factory::class);
```
nunomaduro: This should be: `return new ObjectType(\Illuminate\Contracts\Auth\Guard::class);`.

ForzaSFerrari:
I'm not sure because then I get this.
`Call to an undefined method Illuminate\Contracts\Auth\Factory::user().`

nunomaduro:
@ForzaSFerrari That's because the AuthManager is not covered by Larastan yet. Let's leave this Pull Request open until I fix it. ☺️

UksusoFF:
@nunomaduro any news here?

nunomaduro:
@UksusoFF No time to work in this package. Fell free to help.

PR #170, opened 2018-10-22
Fix internal error caused by non-existing provider name

aso824:
I found an internal error in Larastan in my app, appeared because I don't have "users" provider in my auth config - renamed it to "client" (and added "admin"). Method `sendToPipeline` expects string, and null is given from `$config->get`.


nunomaduro:
Thanks @aso824!

PR #178, opened 2018-10-24
Make it possible to customize the default config file path

### LINKED ISSUE #134: Configuration file for users
I would like to make a suggestion about a generated larastan configuration file. Even if it's just an idea, the configuration file could contain the following:

- your preferred error format,
- the paths that should be tested,
- the rule level that the user prefers,
- possibly the memory limit set by the user,
- if he likes the progress bar or just want to see the results,

My thought behind this idea is just to use the 'php artisan code:analyse' command instead of attaching several sub commands at the end.

That configuration file should be then, in the usually laravel structure for configuration files.

johanvanhelden:
I found it cumbersome having to define my own configuration file every time. So I made it configurable.

So before this PR I had to run:
`artisan code:analyse --configuration=$PWD/phpstan.neon`

The solution is adding a customizable Laravel config file that you can publish by running:
`artisan vendor:publish --tag=larastan-config`

And now I simply change the default:
`'configuration' => base_path('vendor/nunomaduro/larastan/extension.neon'),`

To:
`'configuration' => base_path('phpstan.neon'),`

Now I can just run:
`artisan code:analyse`

I made it backward compatible, so people without the `larastan.php` config file do not run into issues.

Fixes #134

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,11 @@ You may use [Composer](https://getcomposer.org) to install Larastan as a develop
 composer require --dev nunomaduro/larastan
 ```
 
+Publish the configuration file:
+```bash
+php artisan vendor:publish --tag=larastan-config
+```
+
 ### Usage in Laravel Applications
 
 Once you have installed Larastan, you may start analyzing your code using the `code:analyse` Artisan command.
--- a/config-laravel/larastan.php
+++ b/config-laravel/larastan.php
@@ -1,0 +1,13 @@
+<?php
+
+return [
+    /*
+    |--------------------------------------------------------------------------
+    | Configuration
+    |--------------------------------------------------------------------------
+    |
+    | This option allows you to set the path to the phpstan configuration file.
+    |
+    */
+    'configuration' => base_path('vendor/nunomaduro/larastan/extension.neon'),
+];
--- a/src/Console/OptionsResolver.php
+++ b/src/Console/OptionsResolver.php
@@ -67,7 +67,7 @@ final class OptionsResolver
             ->setDefault(base_path('vendor/autoload.php'));
 
         $definition->getOption('configuration')
-            ->setDefault(__DIR__.'/../../extension.neon');
+            ->setDefault($this->defaultConfiguration());
 
         $definition->getOption('memory-limit')
             ->setDefault(self::DEFAULT_MEMORY_LIMIT);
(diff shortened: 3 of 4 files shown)


nunomaduro:
@johanvanhelden Please next expose your idea before coding it. It's better.

johanvanhelden:
@nunomaduro That does not work automatically. The custom configuration is not automatically detected in my application.

If I place a `phpstan.neon.dist` in the root of my project, it still looks at the default file inside your package because that file is hardcoded.

Just to be clear: I've got a Laravel application with a custom config file. I am not making a package.

nunomaduro:
But you can specify a `configuration`option.

Such: `artisan code:analyse --configuration="phpstan.neon"`.

johanvanhelden:
Yes, I know, and that is not user friendly to do that each and every time because it is a lot to type out. So that's why I made it configurable via a Laravel config file.

You can simply define the path to your custom configuration 1 time, and you can simply run `artisan code:analyse`. That is way less typing and remembering and it's easier for other people who work on our projects as well.

nunomaduro:
@johanvanhelden Ok. Could you develop a POC where you could define the phpstan configuration inside that same php file?

nunomaduro:
Parameters such as ignoreErrors etc?

johanvanhelden:
I am not sure if that is even possible. I know PHPStan would support the passing of a level, but I am not sure on the other things. But why should all those options be in a PHP config file?

Because that would also involve new documentation that would be Larastan specific, and right now, people can just look at the PHPStan documentation on how to override things and create their own custom config. 

And it's not just `ignoreErrors`, you would have to support all parameters. Like `level` and `excludes_analyse`. And that is a lot of work for something that just works as it is if people can just look at the PHPStan documentation. I'm not even sure how that would be done.

In my opinion, this PR is the best of both worlds. People can use the regular, documented, PHPStan configuration syntax and just set the path to their configuration file without having to append anything to the `artisan code:analyse` command.

nunomaduro:
@johanvanhelden You are right. We shouldn't support phpstan options in a configuration file. I am to tired today. 😅

I would suggest a Pull Request that automatically detects a `phpstan.neon` or a `phpstan.neon.dist` in the root of the project. I really would like to avoid the php configuration file.

johanvanhelden:
@nunomaduro Auto detection sounds even better and would of course also solve my issue. If you would be willing to merge that, I'll work on that tomorrow and make a PR.

nunomaduro:
Yes! Thanks for being so cool. 👍

szepeviktor:
> Yes, I know, that is what I said. And that is, in my opinion, not user friendly to do that each and every time because it is a lot to type out. So that's why I made it configurable via a Laravel config file.

Try adding a composer script:

```json
    "scripts": {
        "analyze": "artisan code:analyse --configuration=$(pwd)/phpstan.neon",
        "lint": "parallel-lint --blame --exclude vendor/ .",
        "sniff": "vendor/bin/phpcs",
    }
```

and issue `composer run-script analyze`

johanvanhelden:
Will be making a new PR with autodetection.

PR #179, opened 2018-10-24
Add support for url() helper method

### LINKED ISSUE #177: url() helper returns UrlGenerator|string
- Larastan Version: 0.3.4
- `--level` used: 6

### Description:
When using the `url()` helper provided by Laravel, it returns an instance of `Illuminate\Contracts\Routing\UrlGenerator` when the `$path` parameter is `null` (which it is by default). But when providing a string to the `url()` method, it always returns a `string`.

Passing the return result of the url() method to a function that requires a string, raises an error:
```
Parameter #2 $url of method Illuminate\Notifications\Messages\SimpleMessage::action() expects string, Illuminate\Contracts\Routing\UrlGenerator|string given.
```

### Laravel code where the issue was found:
```php
$redirectTarget = url('http://somewebsite.com');
```

devfrey:
Fixes #177 

Since there were no unit tests covering the existing return type helpers, I gave it my best shot to write my own tests. Please let me know what you think.

**Notes**
This implementation only covers usage like this:
```php
$generator = url();
```
In theory, passing `null` as the first argument (directly or indirectly) would also result in a `UrlGenerator` instance, but I was not able to implement a check for that.


devfrey:
What's up with StyleCI's strange conventions for imports?

(analyses: https://github.styleci.io/analyses/q2RoxB)

Korri:
Nice PR! For the Tests though, you can get away with a bit less mocking (which is always better to prevent changes in PHPStan from breaking the tests), have a look at how we did it here: https://github.com/lxrco/phpstan-laravel/blob/master/tests/HelpersReturnTypeExtensionTest.php

devfrey:
> Nice PR! For the Tests though, you can get away with a bit less mocking

I see the same amount of mocked objects in the test you referenced. Am I missing something? The only (big) difference is the use of a data provider, which I didn't find suitable because there are different assertions for both return types. (In case of ObjectType it also performs an assertion on the object type, which is not the case for the string return type).

szepeviktor:
And maybe the copyright need changing: https://travis-ci.org/nunomaduro/larastan/jobs/445860198#L674

devfrey:
@szepeviktor I've added the copyright notice. Perhaps consider mentioning that requirement in  [Contributing](https://github.com/nunomaduro/larastan/blob/master/CONTRIBUTING.md)?

szepeviktor:
Okay.
Now all should be green!

nunomaduro:
It’s easy, I will work on an example tonight ok? 🤙

devfrey:
> @devfrey Just added tests for the `view` and `response` helpers. Can you add tests for the `url` extension?

Sure thing!

devfrey:
I've added new tests!

nunomaduro:
Thanks for this @devfrey !

PR #181, opened 2018-10-24
Run laravel test in production

szepeviktor:
@nunomaduro What is your opinion?


nunomaduro:
I will be able to check issues and pull requests after Friday afternoon. 🤙

nunomaduro:
@szepeviktor Technically Larastan will be used in Laravel **with** dev dependencies. I would prefer install dev dependencies.

szepeviktor:
Yes, you're right.

I thought there may be problems without Larastan's dev-deps.

PR #188, opened 2018-10-29
[Work in progress] Feature testing

Korri:
This is a work in progress, I need help finishing it.

This would allow to create test cases for all the parts of Larastan, but there are a few issues that must or should be fixed:

- Larastan start PHPStan in another process, so no code coverage can be reported.
- [This](https://github.com/nunomaduro/larastan/commit/7e6fef41039817160f377e55641d4d9010b21c77#diff-300c6013fb9bc1bc02b8fbd9fb9ac2b0R15) is not something that can be merge, but I don't know how to make it work.

Some parts of the code can feel a bit hachish to, but I'm sure it can be improved, I just don't know the internals of PHPStan/Larastan well enought.

EDIT: I think coverage might be OK to merge, still better then no tests, but the second one definitely needs a solution. Looking a bit more at PHPStan code I still see no way to make it work unless [relative path includes work](https://github.com/phpstan/phpstan/issues/662).

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -17,7 +17,9 @@ define('LARAVEL_START', microtime(true));
 
 if (file_exists($applicationPath = __DIR__.'/../../../bootstrap/app.php')) { // Applications
     $app = require $applicationPath;
-} else if (file_exists($applicationPath = getcwd().'/../../../../bootstrap/app.php')) { // Local Dev
+} elseif (file_exists($applicationPath = getcwd().'/../../../../bootstrap/app.php')) { // Local Dev
+    $app = require $applicationPath;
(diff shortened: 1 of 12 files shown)


nunomaduro:
@Korri Thanks for proposing this! Just to make it sure: Have you considered that `orchestra/testbench` already contains a fake application inside of it?

Korri:
@nunomaduro You're right its much lighter like that, the two issues still stand though.

nunomaduro:
I have made a new pull request with the contents of this pull request and some more stuff: https://github.com/nunomaduro/larastan/pull/190.

Thank you so much for your work!

PR #194, opened 2018-10-31
Config auto detection

johanvanhelden:
As we discussed yesterday in https://github.com/nunomaduro/larastan/pull/178, this is the PR for configuration auto-detection.

This will allow the user to place a custom configuration file in the root of the application, and Larastan will automatically pick that one up instead of the hard-coded vendor file.

So you no longer have to run:
`artisan code:analyse --configuration=$PWD/phpstan.neon`

But now, you can simply run:
`artisan code:analyse`

I've also added some extra documentation on setting up a custom configuration.


nunomaduro:
Thanks for this! 👍

PR #199, opened 2018-11-06
Update README.md

AdrianHL:
Small fix so the goal of the excluded message pattern match the expected string


szepeviktor:
Thank you.
Could you escape the dot in all lines?

AdrianHL:
Done @szepeviktor

szepeviktor:
Thank you.

PR #211, opened 2018-11-25
fixes #210. Auth::user() return type is nullable

### LINKED ISSUE #210: Auth::user() can return null
- Larastan Version: v0.3.10
- `--level` used:

### Description:
`php artisan code:analyse` says
> If condition is always true.

However, the return type for `Auth::user()` should be nullable https://github.com/laravel/framework/blob/5.7/src/Illuminate/Support/Facades/Auth.php#L10

### Laravel code where the issue was found:
```
if (Auth::user()) {
            return redirect('/home');
}
return view('welcome');
```

mr-feek:
https://github.com/nunomaduro/larastan/issues/210

### DIFF
--- a/src/ReturnTypes/AuthExtension.php
+++ b/src/ReturnTypes/AuthExtension.php
@@ -59,7 +59,7 @@ final class AuthExtension implements DynamicStaticMethodReturnTypeExtension
             ->get('config');
 
         if ($userModel = $config->get('auth.providers.users.model')) {
-            return new ObjectType($userModel);
+            return TypeCombinator::addNull(new ObjectType($userModel));
         }
 
         return ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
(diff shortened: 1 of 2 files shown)


tests/Features/ReturnTypes/AuthExtension.php
```diff
@@ -9,7 +9,7 @@
 
 class AuthExtension
 {
-    public function testUser(): User
+    public function testUser(): ?User
```
mr-feek: This isn't a proper regression test, but I'm not familiar with the testing strategy used here and haven't had a chance to dig into it.

src/ReturnTypes/AuthExtension.php
```diff
@@ -59,7 +59,7 @@ public function getTypeFromStaticMethodCall(
             ->get('config');
 
         if ($userModel = $config->get('auth.providers.users.model')) {
-            return new ObjectType($userModel);
+            return TypeCombinator::addNull(new ObjectType($userModel));
```
nunomaduro: When the user is not logged, the `User::auth()` returns `null`?

mr-feek: Correct, here is an example in the default `SessionGuard` that laravel provides: https://github.com/laravel/framework/blob/5.7/src/Illuminate/Auth/SessionGuard.php#L115-L117

nunomaduro:
Perfect. Thanks!

PR #240, opened 2019-02-15
Providers not bootstrapped in the Lumen Framework

click-kb:


szepeviktor:
Thank you for your addition!

What do you think? Could we also test a simple lumen app?
like https://github.com/nunomaduro/larastan/blob/master/tests/laravel-test.sh

szepeviktor:
https://github.com/nunomaduro/larastan/pull/242

Issue #244, opened 2019-02-15
Fix Lumen error


szepeviktor:
```
 ------ ----------------------------------------------------------------------
  Line   app/Providers/AuthServiceProvider.php
 ------ ----------------------------------------------------------------------
  33     Cannot call method viaRequest() on Illuminate\Auth\AuthManager|null.
 ------ ----------------------------------------------------------------------
```

Can anybody land me a hand how to solve this one in Lumen?
https://github.com/laravel/lumen/blob/master/app/Providers/AuthServiceProvider.php#L33

Now we test Lumen too in CI.

szepeviktor:
@goodevilgenius ?

goodevilgenius:
@szepeviktor Not sure. It seems to think that `$app['auth']` can be `Illuminate\Auth\AuthManager|null`. `AuthManager` has a `viaRequest` method, but of course `null` doesn't, which seems to be the problem.

But from my reading of the code, it's not possible for `$app['auth']` to be `null`.

The binding for `auth` is registered at https://github.com/laravel/lumen-framework/blob/87e7dcccda1b83ef0953253d03c056575308439e/src/Application.php#L270-L272

Or more precisely https://github.com/illuminate/auth/blob/master/AuthServiceProvider.php#L35-L42

So, the question is, why does it think it could be `null` as well?

szepeviktor:
@ondrejmirtes Could it be a phpstan bug?

ondrejmirtes:
@szepeviktor I don't know what `$app` is and who is defining the type of `$app['auth']` - maybe some dynamic method return type extension hooked onto `offsetGet`?

szepeviktor:
`$app` is `\Illuminate\Contracts\Foundation\Application|\Illuminate\Foundation\Application`

https://github.com/laravel/framework/blob/5.7/src/Illuminate/Support/ServiceProvider.php#L9-L14

It is not clear to me how it can be an [array] also.

goodevilgenius:
In Lumen, it's actually `Laravel\Lumen\Application`. It doesn't extend, `Illuminate\Foundation\Application`, because `Illuminate\Foundation` is *only* in Laravel, and not Lumen.

It works as an array because `Illuminate\Foundation\Application` and `Laravel\Lumen\Application` extend `Illuminate\Container\Container`, which implements `ArrayAccess`.

[...]

PR #245, opened 2019-02-15
Lumen testing needs registration

szepeviktor:


szepeviktor:
End up waiting for #244

Issue #248, opened 2019-02-19
Cannot call method store() uploaded file from request


claystation:
- Larastan Version: 0.3.15
- `--level` used: max

### Description
Larastan does not recognize the `store()` method on an uploaded file. It gives the error:
`Cannot call method store() on array<Illuminate\Http\UploadedFile>|Illuminate\Http\UploadedFile|null`

### Laravel code where the issue was found

```php
$request->file('file')->store('images', 'public');
```

PR #249, opened 2019-02-19
Added test and extension for file method on Request.

claystation:
Solves issue #248

This is my first PR, please add constructive feedback instead of just closing. Help me help you.


src/ReturnTypes/RequestFileExtension.php
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getTypeFromMethodCall(
+        MethodReflection $methodReflection,
+        MethodCall $methodCall,
+        Scope $scope
+    ): Type {
+        return new MixedType;
```
ondrejmirtes: Why mixed? Isn't there a more specific type we can use?

claystation: Return type from `file()` is `\Illuminate\Http\UploadedFile|\Illuminate\Http\UploadedFile[]|array|null`. As far as I could find in the list of types, this looked like the only fit? Or have i overlooked something.

ondrejmirtes: Definitely not - please watch this talk about in-depth look at PHPStan's typesystem: https://www.youtube.com/watch?v=85Aq2rWpGQE

We should differentiate when it returns the object, when it returns the array and when it returns null:

```
new ObjectType(\Illuminate\Http\UploadedFile|\Illuminate\Http\UploadedFile::class)
new ArrayType(new IntegerType(), new ObjectType(\Illuminate\Http\UploadedFile::class))
new NullType()
```

I guess that it returns the object when an argument is present and it returns the array when there's not argument and the method is called like this: `$request->file()`

Not sure when it returns null.

claystation: I think it returns null when `$default` is not specified (default for `$default` is null) and the `$target` is not a Collection or Array? I will look into the youtube video and update the PR. Thanks!

claystation: In this scenario, accessing the `file()` method which accesses the `data_get()` method, the `$target` will never be null. It will always be an array. So would this new implementation suffice?

claystation:
How do i configure my IDE so that it sorts the import statements in the order as StyleCI wants them? IntelliJ auto sorts them alphabetically, but StyleCI does not seem to like that.

src/ReturnTypes/RequestFileExtension.php
claystation: Nevermind, it will definitely return null when the `$key` does not exist in the `$target` variable and `$default` is not set to anything but null. How would one go about specifying this as we have no influence over the input array?

ondrejmirtes: Can we recognize it statically? Is it defined somewhere that the key must be there? Or should we switch the return types to array|null + object|null?

claystation: I think we can recognise it statically?

Paths: 
- `$key` input null, returns whole array (which may be empty).
- `$key` existent returns UploadedFile
- `$key` defined but non existent in input array returns `$default` which has a default value of null.

But, how would one check for this? Also, return type can be array + object|null right?

ondrejmirtes: Where's the `$default`? It's the second parameter of the method?

ondrejmirtes: And how do you find out that `$key` is existent?

claystation: Yes its the second parameter:
`public function file($key = null, $default = null)` [InteractsWithInput.php](https://github.com/laravel/framework/blob/5.7/src/Illuminate/Http/Concerns/InteractsWithInput.php#L375)

And the `data_get()` method will check that for you [here](https://github.com/laravel/framework/blob/5.7/src/Illuminate/Support/helpers.php#L475).

ondrejmirtes: So that's easy - check `$methodCall->args[1]` - if it does not exist, consider it NullType, if it exists, use `$scope->getType($methodCall->args[1]->value)`.

claystation: > check `$methodCall->args[1]` - if it does not exist, consider it NullType
That is not true, because even if `args[1]` is not defined, it can still return an `UploadedFile`.

claystation: So actually if `args[1]` is not defined, it is actually a `UploadedFile|null` type.

ondrejmirtes: Yes, that's what I was saying. You need to `TypeCombinator::union(new ObjectType(UploadedFile), $defaultType)` and return that.

claystation: Ah i misunderstood you, now that i've read it twice i fixed it. This gives me an error on the test tho actually portraying the original problem but narrowed down the types. `Cannot call method store() on Illuminate\Http\UploadedFile|null.`. Which would make sense, if you do not verify your request at first? But if you do, and of course static code analysis has no knowledge of that, there wouldn't be a problem.

ondrejmirtes: I think this error is fine. It shows up only on level 7 - the strictest level.

claystation: Well, i guess i should start adding defaults to my project then. Thanks for the support and feedback!

ondrejmirtes: Cool - I'm glad there's now one more person that can write PHPStan extensions :)

claystation: I will definitely keep an eye on the issues here :)

tests/Features/ReturnTypes/RequestFileExtension.php
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types=1);
+
+namespace Tests\Features\ReturnTypes;
+
+use Illuminate\Http\Request;
+use Illuminate\Http\UploadedFile;
+
+class RequestFileExtension
+{
+    public function testRequestFileStore()
```
nunomaduro: Hey, just missing here the return type. Check other tests.

tests/Features/ReturnTypes/RequestFileExtension.php
```diff
@@ -0,0 +1,23 @@
...
+    {
+        $request = new Request();
+        $request->file('image', new UploadedFile('', ''))->store('/path');
+    }
+
+    public function testRequestFileArrayStore()
```
nunomaduro: Hey, just missing here the return type. Check other tests.

tests/Features/ReturnTypes/RequestFileExtension.php
nunomaduro: Hey, just missing here the return type. Check other tests.

tests/Features/ReturnTypes/RequestFileExtension.php
nunomaduro: Hey, just missing here the return type. Check other tests.

nunomaduro:
Fell free to re-open this pull request once it's ready.

claystation:
Ok i never figured it was closed and not merged. Anyway i fixed it. Still needed?

canvural:
Hi @claystation This is fixed by #695

You can try the latest version of Larastan

claystation:
Cool cool cool! Keep this pr closed then!

PR #260, opened 2019-03-30
Ignore error from issue #244

szepeviktor:


nunomaduro:
Hey buddy, the solution for this should be code an extension and not ignoring the error.

szepeviktor:
Yes, I wish I could understand Larastan.

nunomaduro:
Let's do it for now.

Issue #291, opened 2019-08-05
Crash with variadic model scope functions (Undefined offset: 0)


wi1dcard:
# Variadic Model Scope Issue

- Larastan Version: `v0.3.18`
- PHPStan Version: `0.11.12`
- `--level` used: 5 (default).

### Description

Short version:

Larastan crashes while analyzing invoking model's dynamic scope functions with variadic parameters. For instance:

```php
# app/User.php
class User extends Authenticatable
{
    // ...

    public function scopeOfTypes($query, ...$types)
    {
        return $query->whereIn('type', $types);
    }
}

# In controllers or console commands:
User::query()->ofTypes(0, 1, 2);
```

Output:

```
 ------ -------------------------------------------------------------- 
  Line   app/Console/Commands/QueryScopeCommand.php                    
 ------ -------------------------------------------------------------- 
         Internal error: Undefined offset: 0                           
         Run PHPStan with --debug option and post the stack trace to:  
         https://github.com/phpstan/phpstan/issues/new                 
 ------ --------------------------------------------------------------
```

### How to Reproduce the Problem

> For quickly reproducing, I create a repo [here](https://github.com/wi1dcard/larastan-variadic-model-scope) which includes minimal codes and steps. You can find what I did step by step in the commit history.

1. `git clone git@github.com:wi1dcard/larastan-variadic-model-scope.git`
2. `composer install`
3. `php artisan code:analyse --debug`

```
   ErrorException  : Undefined offset: 0

  at /Users/wi1dcard/GitHub/larastan-variadic-model-scope/vendor/phpstan/phpstan/src/Rules/FunctionCallParametersCheck.php:119
    115|                                if (!$parametersAcceptor->isVariadic() || count($parameters) === 0) {
    116|                                        break;
    117|                                }
    118| 
  > 119|                                $parameter = $parameters[count($parameters) - 1];
    120|                                $parameterType = $parameter->getType();
    121|                                if (!($parameterType instanceof ArrayType)) {
    122|                                        break;
    123|                                }
```

These outputs seem to point to a PHPStan bug. However, after working on this issue 2 days, I start to think this may cause by Larastan:

### More Information that I found so far

1. Removing [src/Methods/Pipes/ModelScopes.php#L44][larastan-src-1] could solve this issue.

[...]

PR #298, opened 2019-08-17
Macro implements MethodReflection instead of BuiltinMethodReflection

ondrejmirtes:
This is fix for PHPStan 0.11.14. This PR changes what interface `Macro` implements, and some additonal cleanups stem from that.

This is closer to how PHPStan extensions should work. It's better to implement `MethodReflection` in a class directly rather than pass it through PhpMethodReflectionFactory.

I'm sorry about these BC breaks, it didn't occur to me that some extensions would be using the `BuiltinMethodReflection` interface. I want to have much clearer backwards compatibility boundary in the future, probably by tagging some classes and interfaces with `@api`. Everything else will be considered internal.

Right now, the rule of thumb is that backwards compatibility covers what's mentioned in the README.

### DIFF
--- a/src/Contracts/Methods/PassableContract.php
+++ b/src/Contracts/Methods/PassableContract.php
@@ -13,7 +13,6 @@ namespace NunoMaduro\Larastan\Contracts\Methods;
 use PHPStan\Broker\Broker;
 use PHPStan\Reflection\ClassReflection;
 use PHPStan\Reflection\MethodReflection;
-use PHPStan\Reflection\Php\PhpMethodReflectionFactory;
 use Illuminate\Contracts\Container\Container as ContainerContract;
 
 /**
@@ -97,9 +96,4 @@ interface PassableContract
      * @return \PHPStan\Broker\Broker
      */
     public function getBroker(): Broker;
-
-    /**
-     * @return \PHPStan\Reflection\Php\PhpMethodReflectionFactory
-     */
-    public function getMethodReflectionFactory(): PhpMethodReflectionFactory;
 }
--- a/src/Methods/Extension.php
+++ b/src/Methods/Extension.php
@@ -17,7 +17,6 @@ use NunoMaduro\Larastan\Concerns;
 use PHPStan\Reflection\ClassReflection;
 use PHPStan\Reflection\MethodReflection;
 use PHPStan\Reflection\BrokerAwareExtension;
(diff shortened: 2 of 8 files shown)


szepeviktor:
@nunomaduro What do you think. I think it's working.

ondrejmirtes:
I'm also fixing #291 here.

nunomaduro:
Blindly trusting you on this. 😅

nunomaduro:
@ondrejmirtes Shouldn't we bump the required phpstan version?

patrickbrouwers:
@nunomaduro I'm getting the following exceptions after updating: 

```
 Whoops\Exception\ErrorException  : Uncaught InvalidArgumentException: Expected string, array of parts or Name instance in vendor/nikic/php-parser/lib/PhpParser/Node/Name.php:234
Stack trace:
#0 vendor/nikic/php-parser/lib/PhpParser/Node/Name.php(26): PhpParser\Node\Name::prepareName(Object(Closure))
#1 vendor/nunomaduro/larastan/src/Methods/Pipes/BuilderLocalMacros.php(61): PhpParser\Node\Name->__construct(Object(Closure))
#2 vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(163): NunoMaduro\Larastan\Methods\Pipes\BuilderLocalMacros->handle(Object(NunoMaduro\Larastan\Methods\Passable), Object(Closure))
#3 vendor/nunomaduro/larastan/src/Methods/Pipes/ModelScopes.php(67): Illuminate\Pipeline\Pipeline->Illuminate\Pipeline\{closure}(Object(NunoMaduro\Larastan\Methods\Passable))

  at vendor/nikic/php-parser/lib/PhpParser/Node/Name.php:234
    230|         } elseif ($name instanceof self) {
    231|             return $name->parts;
232|         }
    233|
  > 234|         throw new \InvalidArgumentException(
    235|             'Expected string, array of parts or Name instance'
    236|         );
237|     }
    238|
```

Seems to be caused by stuff like:

```
Route::bind('myModel', static function (string $uuid): MyModel {
    return MyModel::onlyTrashed()->where('uuid', $uuid)->firstOrFail();
});
```

ondrejmirtes:
Not necessary to bump PHPStan version.

I'm gonna look into the exception, maybe I screwed up.

ondrejmirtes:
I haven't realized that `ReflectionFunction` constructor also accepted closuures, not just strings.

PR #303, opened 2019-08-21
Test CI on PHP 7.4

szepeviktor:


szepeviktor:
I am very happy that 7.4 is more strict

> ErrorException: Trying to access array offset on value of type int

.travis.yml
```diff
@@ -3,9 +3,7 @@ language: php
 sudo: false
 
 php:
-  - 7.1
```
nunomaduro: I think we should still test against older versions no?

szepeviktor: Yes, this PR is a test. I was not going to wait for other version to run.

szepeviktor:
phpstan tests do run on PHP 7.4 https://travis-ci.org/phpstan/phpstan/jobs/573557274

.travis.yml
```diff
@@ -3,9 +3,7 @@ language: php
...
-  - 7.2
-  - 7.3
+  - '7.4snapshot'
```
nunomaduro: Can you update the pull request just adding the `7.4snapshot`?

szepeviktor: Yes, if you solve the above exception.

szepeviktor: @kylekatarnls Do you feel like investigating this one?

kylekatarnls: OK, I will take a look. It may be not harder than a `if (!is_int($foo))` but I will check to be sure.

kylekatarnls: I will base it on my branch for carbon-macro to see if there is no PHP 7.4 incompatibility in my changes neither.

kylekatarnls: Hum, a bit harder than I thought, I think it's something to fix in Laravel:

Laravel do:
`array_unshift($parameters, $command);` putting `$command` (code:analyse) at the top of the parameters with a `0` index (don't see why) while Symfony expect only string keys for `$parameters`

kylekatarnls: Error seems to have been fixed in Symfony:
https://github.com/symfony/symfony/pull/32806/files

szepeviktor: Very well done, thanks for the investigation!!

composer.json
```diff
@@ -12,6 +12,7 @@
     "require": {
         "php": "^7.1.3",
         "symfony/process": "^4.2",
+        "symfony/console": "dev-master",
```
kylekatarnls: You can't do that in the "prod" require. I suggest to change on .travis.yml:
https://github.com/nunomaduro/larastan/pull/306/files#diff-354f30a63fb0907d4ad57269548329e3

kylekatarnls: It will also need to change the minimum-stability and maybe add an alias but the whole thing should done in Travis-CI only, and symfony/console will just be properly update when Symfony 4.4 will be released.

composer.json
```diff
@@ -12,7 +12,7 @@
     "require": {
         "php": "^7.1.3",
         "symfony/process": "^4.2",
-        "symfony/console": "dev-master",
+        "symfony/console": "4.3.x-dev",
```
kylekatarnls: It's 4.4 according to the branch of the commit:
https://github.com/symfony/symfony/pull/32806/commits/5f451e6f7074976aae60a65ba144fff684b7b48c

szepeviktor: Apparent success :)

szepeviktor:
Next problem: _In ServiceProvider.php line 97: Trying to access array offset on value of type null_

https://travis-ci.org/nunomaduro/larastan/builds/576530284

composer.json
kylekatarnls: I guess it have been copied/cherry-picked 1 minor down.

nunomaduro:
Duplicate of https://github.com/nunomaduro/larastan/pull/306.

Issue #295, opened 2019-08-15
Improvement: remove orchestra dependency


nunomaduro:
The addition of the `orchestra` dependency cause too many issues. https://github.com/nunomaduro/larastan/issues/259 https://github.com/nunomaduro/larastan/pull/247 and many others.

**Goal**: 

1. Remove the dependency `orchestra/testbench` here.
2. Add the dependency `orchestra/testbench` to the suggest on composer, and update the readme explaining if you are using `larastan` for packages you need to install this dependency.

> After this feature we need to release a major version.

mfn:
Ah, interesting, I was just wondering today why I could not run larastan with Laravel 6.0.

My failing build https://travis-ci.org/rebing/graphql-laravel/jobs/574566750 at the end says:
> nunomaduro/larastan v0.3.21 requires orchestra/testbench ^3.6

But I've to be honest, the composer dependency output is SO convoluted, I really don't know what I'm looking for and whether this is really the cause; just a guess for now.

mfn:
> 2\. Add the dependency `orchestra/testbench` to the **suggest** on composer

But doesn't _this_ package require orchestra for its own tests?

Shouldn't it be rather moved to `require-dev`?

ondrejmirtes:
@mfn It isn't just for tests. It sets up the application for Laravel package development.

mfn:
> It sets up the application for Laravel package development.

Oh, sorry. I was not really aware of it. I only know this package for "providing a Laravel compatible test runner for the package".

I'll see myself out before adding too much noise here ;)

szepeviktor:
@mfn Your comments are welcome.

martijngastkemper:
I've made a PR (#304) with an alternative solution. Not removing the requirement, but add additional versions to the constraint.

PR #304, opened 2019-08-25
Require appropriate orchestra/testbench version

martijngastkemper:
In #295 the proposed solution is remove `orchestra/testbench` from as required dependency and add it as a suggestion. I think this is a better solution because in most cases users won't notice anything of the testbench version stuff. Composer will figure out the right version. 

People will get wonky Composer errors when non compatible PHPUnit versions are required. But that's also the case when the described solution is in place.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "php": "^7.1.3",
         "symfony/process": "^4.2",
         "phpstan/phpstan": "^0.11.15",
-        "orchestra/testbench": "^3.6",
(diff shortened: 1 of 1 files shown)


szepeviktor:
Seems like Larastan is not compatible with `composer require "orchestra/testbench:~3.6.0"`

nunomaduro:
I would prefer remove this dependency: https://github.com/nunomaduro/larastan/issues/295.

PR #306, opened 2019-08-25
PHP 7.4 compatibility

kylekatarnls:


nunomaduro:
Can you update the pr?

kylekatarnls:
Done. As @szepeviktor noticed here: https://github.com/nunomaduro/larastan/pull/303#issuecomment-524656377

There is an other problem to solve in/around ServiceProvider.php.

kylekatarnls:
This incompatibility is not yet fixed by Laravel, it needs:
https://github.com/laravel/framework/pull/29482 + https://github.com/driesvints/framework/pull/1

kylekatarnls:
Meanwhile we may add an empty view config in tests so `$this->app->config['view']` would be an array and the errors would not be raised.

kylekatarnls:
Fixed. So no matter Lumen will be compatible to PHP 7.4, we can check than larastan itself is now.

nunomaduro:
Thanks!

PR #317, opened 2019-10-03
Fix for auth guard method calls on auth helper and facade

### LINKED ISSUE #309: Call to an undefined method Illuminate\Contracts\Auth\Factory::xxx()
- Larastan Version: 0.3.21
- `--level` used: default/not set

### Description

>  53     Call to an undefined method `Illuminate\Contracts\Auth\Factory::check()`.  
>  53     Call to an undefined method `Illuminate\Contracts\Auth\Factory::user()`.  
>  80     Call to an undefined method `Illuminate\Contracts\Auth\Factory::login()`.   

```php
auth()->check(); 
auth()->user();
auth()->login();
```

This is only for example, I get this error everywhere where I use `auth()->xxx()`.

canvural:
Fixes #309 

Also adds tests for `user`, `check` and `id` method calls on both auth helper and facade.

I'm not %100 sure if this does not break anything else (auth related) but according to my tests, it looks fine.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -39,6 +39,11 @@ services:
             - phpstan.broker.dynamicStaticMethodReturnTypeExtension
 
     -
+        class: NunoMaduro\Larastan\ReturnTypes\AuthManagerExtension
+        tags:
+            - phpstan.broker.dynamicMethodReturnTypeExtension
+
+    -
         class: NunoMaduro\Larastan\ReturnTypes\RequestExtension
         tags:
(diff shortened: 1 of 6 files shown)


szepeviktor:
Thank you for your work!

@canvural Could it be that your commit were made with a non-GitHub email from aequitas-group.pl?

canvural:
@szepeviktor Ah, yes. I was working on my work laptop. Forgot to change the address. Is that gonna be a problem?

szepeviktor:
Not for this repo, maybe for you.

PR #319, opened 2019-10-05
Model scope call after relation returns Builder

canvural:
Currently code like this `$this->hasOne(User::class)->active();` results in error saying `Call to an undefined method Illuminate\Database\Eloquent\Relations\HasOne::active()`. This PR fixes this.

`Relation` class in Laravel forwards unknown calls to `Builder`. And `Builder` can call scope methods on the model. So it is safe to call model scopes from the relation, like above.

There was one issue that I couldn't solve. As far as my understanding goes for the Reflection and PHPStan, there is no way to know the calling context from the reflection. 

So in this implementation, I'm treating every unknown call as a scope method call, if the method doesn't also exist in `Builder`

I'm open to suggestions if anyone knows how to get the actual scope method reflection from inside the `ModelScopeAfterRelations` pipe.

### DIFF
--- a/src/Methods/Kernel.php
+++ b/src/Methods/Kernel.php
@@ -70,6 +70,7 @@ final class Kernel
                     Pipes\BuilderLocalMacros::class,
                     Pipes\BuilderDynamicWheres::class,
                     Pipes\RedirectResponseWiths::class,
+                    Pipes\ModelScopeAfterRelations::class,
                 ]
             )
             ->then(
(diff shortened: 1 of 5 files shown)


nunomaduro:
Thanks!

nunomaduro:
@canvural Let me know if you would like to maintain this project.

canvural:
@nunomaduro I don't have too much experience with writing PHPStan extensions. I don't know if I'd be the right person.

I'm just fixing the errors that I found when running Larastan on my project. And I have some free time over the next 2 or 3 weeks. I'll continue to fix the issues I found. As for the maintaining let's see after this time.

Issue #324, opened 2019-10-08
Wrong tests and possible wrong implementation


canvural:
In `ModelExtension` test class we have [tests](https://github.com/nunomaduro/larastan/blob/master/tests/Features/Methods/ModelExtension.php#L23-L49) that are using some Builder methods on a model and expect the return type to be the model itself.

I'm thinking this is wrong. I also checked this in a different project in tinker. `(new User)->whereId(1)` statements like this, returns `Illuminate\Database\Eloquent\Builder`

Am I missing something or this is actually wrong? I can fix this but wanted to ask first.

cc @szepeviktor, @nunomaduro

nunomaduro:
The `Illuminate\Database\Eloquent\Builder` contains a magic method that proxies all calls to the original model. That's why.

canvural:
@nunomaduro But at the end, the return type shouldn't be the model itself. You have to call `get` to return the model itself.

canvural:
You can also test it in a Laravel project by doing `dd(User::whereId(1));` This dumps the Builder instance, not the model.

nunomaduro:
How about supporting relations methods on that chain? Today is possible because we modify the return type to a model:

```php
User::whereId(1)->companies();
```

canvural:
@nunomaduro That's not a valid Laravel code.

PR #325, opened 2019-10-10
Handle magic calls on model and query builder correctly

canvural:
Ok. This turned out to be a little bigger PR than I imagined :sweat_smile: But the outcome is good! I tested it on my big project. And I could go to level 4 from level 3 without changing anything.

List of changes:

- `Model` class' `__call` method is mimicked. If it's not `increment` or `decrement` calls are forwarded to `\Illuminate\Database\Eloquent\Builder`
- `\Illuminate\Database\Eloquent\Builder` `__call` method is mimicked. With one exception. Local macros are not supported. Otherwise, all unknown calls are forwarded to `\Illuminate\Database\Query\Builder` class. But return type is still `\Illuminate\Database\Eloquent\Builder`
- Added support for calling `get` on the query builder. This will return `Collection` class and elements of the collection correctly typehinted with the model. 
- Return type of query builder methods is fixed. Previously `User::where('foo', 'bar')` was returning the model itself. See #324 This is fixed now
- Return type of model local scopes is correctly handled now.
- Added correct return types for model retrieval and creation methods. Ex. `find`, `firstOrCreate` etc.
- Added more test cases for everything.

There might be an increased number of errors after these changes. But this is because there were false negatives before. For example Larastan would think `User:whereFoo('bar')->someRelation()` is a correct code and returning model instance. But that example is not a valid Laravel code in reality. So I guess any of this can't be considered as breaking change. And all the existing (which I didn't change) tests are passing, so existing functionality is preserved.

I'm open to any comments and questions. About implementation or other things.

### DIFF
--- a/src/Methods/EloquentBuilderForwardsCallsExtension.php
+++ b/src/Methods/EloquentBuilderForwardsCallsExtension.php
@@ -22,6 +22,7 @@ use PHPStan\Reflection\MethodReflection;
 use PHPStan\Reflection\PassedByReference;
 use PHPStan\Reflection\BrokerAwareExtension;
 use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Reflection\Dummy\DummyMethodReflection;
 use PHPStan\Reflection\MethodsClassReflectionExtension;
 use PHPStan\Reflection\Native\NativeParameterReflection;
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
@@ -64,7 +65,7 @@ final class EloquentBuilderForwardsCallsExtension implements MethodsClassReflect
             return true;
         }
 
-        return $this->getBuilderReflection()->hasNativeMethod($methodName);
+        return true;
     }
 
     public function getMethod(ClassReflection $classReflection, string $methodName): MethodReflection
@@ -78,6 +79,11 @@ final class EloquentBuilderForwardsCallsExtension implements MethodsClassReflect
 
         if (in_array($methodName, $this->passthru)) {
             return $this->getBroker()->getClass(Builder::class)->getNativeMethod($methodName);
+        }
+
+        // Could be a model scope
+        if (! $this->getBuilderReflection()->hasNativeMethod($methodName)) {
+            return new DummyMethodReflection($methodName);
         }
 
         $methodReflection = $this->getBuilderReflection()->getNativeMethod($methodName);
--- a/src/Methods/EloquentBuilderForwardsCallsExtension.php
+++ b/src/Methods/EloquentBuilderForwardsCallsExtension.php
@@ -13,21 +13,19 @@ declare(strict_types=1);
 
 namespace NunoMaduro\Larastan\Methods;
 
+use PHPStan\Type\StringType;
+use PHPStan\Type\CallableType;
 use NunoMaduro\Larastan\Concerns;
 use Illuminate\Database\Query\Builder;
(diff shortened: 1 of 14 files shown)


szepeviktor:
Too big to fail.

src/Methods/ModelForwardsCallsExtension.php
```diff
@@ -0,0 +1,134 @@
...
+    private $modelRetrievalMethods = ['first', 'find', 'findMany', 'findOrFail'];
+
+    /** @var string[] */
+    private $modelCreationMethods = ['make', 'create', 'forceCreate', 'findOrNew', 'firstOrNew', 'updateOrCreate', 'fromQuery'];
+
+    /**
+     * @return ClassReflection
+     * @throws \PHPStan\Broker\ClassNotFoundException
+     */
+    protected function getBuilderReflection(): ClassReflection
```
nunomaduro: This can be private?

canvural: Changed. :+1:

src/Methods/EloquentBuilderForwardsCallsExtension.php
```diff
@@ -0,0 +1,96 @@
...
+
+        if ($methodName === 'macro') {
+            return true;
+        }
+
+        if (in_array($methodName, $this->passthru)) {
+            return true;
+        }
+
+        return true;
```
CyberiaResurrection: @canvural , why do you have the same action in both branches of the if statement on line 64?

canvural:
@nunomaduro Can we merge this and make a new `0.4.2` release? Or do you want this in a new release? `0.5` maybe ?

szepeviktor:
@canvural Please do release v0.4.2

src/Methods/EloquentBuilderForwardsCallsExtension.php
canvural: This method tries to mimic the magic method `__call` in the Eloquent Builder class. `if` checks are because of that. And the last `return true` is because model scopes can be called on the builder, and they can have any name. So we need to return true always.

Of course, we can just return true, and not do the checks. But in my opinion this reads better, and close to the `__call` in the builder class.

CyberiaResurrection: Fair enough, @canvural - maybe add your reply as a comment in that code?

canvural:
@nunomaduro Is there anything left to do here? Can we merge?

canvural:
Ok, so it's up to you. We can't merge because of the `continuous-integration/styleci/push Expected` status. :roll_eyes:

bobbybouwmann:
@nunomaduro Can we force merge this, since we're only waiting for the style ci?

nunomaduro:
Done.

bobbybouwmann:
@nunomaduro Do you have an ETA for the release? :D

canvural:
I was thinking maybe we can test it a little, with (`dev-master` in composer), and then make a release if everything seems fine.

But I can make the release 0.4.3, if it's ok with @nunomaduro

spaceemotion:
I will try the changes as soon as I am able (still today). Will report back in the issues once I got around to it.

bobbybouwmann:
Thanks for the awesome work guys!

nunomaduro:
@bobbybouwmann Share some love RT @canvural his work: https://twitter.com/can__vural/status/1186564559261388800.

PR #329, opened 2019-10-16
Use AnnotationExtension as first extension

### LINKED ISSUE #251: [False positive] Property does not accept Illuminate\Support\Facades\Gate
- Larastan Version: 0.3.15
- `--level` used:5

### Description
```
Property App\xxxxx::$gate (Illuminate\Contracts\Auth\Access\Gate) 
does not accept Illuminate\Support\Facades\Gate.
```

According to the `@property`-Hints this method returns an instance of `\Illuminate\Contracts\Auth\Access\Gate` which PHPStan/Larastan seems to ignore:

https://github.com/laravel/framework/blob/c821cbf6f20fca4c50aa9d8fb08f71425afc07b8/src/Illuminate/Support/Facades/Gate.php#L19

```php
class xxxxx
{
    /**
     * @var \Illuminate\Contracts\Auth\Access\Gate
     */
    private $gate;


    public function __construct(Request $request)
    {
        $this->gate = Gate::forUser($request->user());
    }
}
```

canvural:
When phpstan loads the extension,`AnnotationExtension` is executed at last.

This PR, adds our own extension which just mimics the `AnnotationExtension` and runs it **first**.

This allows to first check the class annotations, and if a magic method is documented there, it'll use the return type and parameters from there.

I think this is ok in our case. Laravel has some useful annotations for classes that defines methods by magic `__call` or `__callStatic` functions.

Fixes #251

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -19,6 +19,11 @@ parameters:
 
 services:
     -
+        class: NunoMaduro\Larastan\Methods\AnnotationExtension
+        tags:
+            - phpstan.broker.methodsClassReflectionExtension
+
+    -
         class: NunoMaduro\Larastan\Methods\ModelForwardsCallsExtension
         tags:
             - phpstan.broker.methodsClassReflectionExtension
--- a/src/Methods/AnnotationExtension.php
+++ b/src/Methods/AnnotationExtension.php
@@ -1,0 +1,52 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of Larastan.
+ *
+ * (c) Nuno Maduro <enunomaduro@gmail.com>
+ *
+ *  For the full copyright and license information, please view the LICENSE
+ *  file that was distributed with this source code.
+ */
+
+namespace NunoMaduro\Larastan\Methods;
+
+use NunoMaduro\Larastan\Concerns;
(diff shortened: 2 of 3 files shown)


szepeviktor:
we fix it later https://github.styleci.io/analyses/X0Ydbj

canvural:
I can fix the style issues. That's not the issue. I was just waiting to see if StyleCI gonna do something :smile:  Because of the `continuous-integration/styleci/push Expected` thing. I guess this should not be `Required`

szepeviktor:
> I guess this should not be `Required`

StyleCI is not the greatest product in the world. (from a sustainability point of view)

szepeviktor:
@nunomaduro Could you lend us a click?

canvural:
Travis build is failing for some unrelated reason :man_shrugging: And it seems I can't trigger a rebuild. (without committing something new)

szepeviktor:
> (without committing something new)

I call these commit _bunny commits_

szepeviktor:
> https://storage.googleapis.com/travis-ci-language-archives/php/binaries/ubuntu/16.04/x86_64/php-7.3.tar.bz2

It was not possible to download PHP 7.3

szepeviktor:
🛫 restarted

PR #330, opened 2019-10-17
Add extension for abort_if

### LINKED ISSUE #116: abort_if not treated like if statements by type checker
- Larastan Version: 0.2.4
- `--level` used: max

### Description:
Currently `abort_if` does not seem to get treated similarly to if, anything designated as nullable is still considered nullable after `abort_if`

This results in an error when querying relations on the object:
```
Cannot call method theRelation() on mixed.
```

### Workaround/example:
Current workaround for this is to use an if statement with a regular `abort`:

```php
// Doesn't pass checks
$foo = ModelName::where(['foo' => 'bar'])->first();
abort_if(!$foo, 404);

print_r($foo->theRelation());

// Passes checks
$foo = ModelName::where(['foo' => 'bar'])->first();
if (!$foo) {
  abort(404);
}

print_r($foo->theRelation());
```

canvural:
Fixes #116

### DIFF
--- a/src/Types/AbortIfFunctionTypeSpecifyingExtension.php
+++ b/src/Types/AbortIfFunctionTypeSpecifyingExtension.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace NunoMaduro\Larastan\Types;
+
+/*
+ * This file is part of Larastan.
+ *
+ * (c) Nuno Maduro <enunomaduro@gmail.com>
+ *
+ *  For the full copyright and license information, please view the LICENSE
+ *  file that was distributed with this source code.
+ */
+
+use PHPStan\Analyser\Scope;
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\TypeSpecifier;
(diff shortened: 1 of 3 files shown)


nunomaduro:
Perfect, thanks!

src/Types/AbortIfFunctionTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,51 @@
...
+        return $functionReflection->getName() === 'abort_if' && $context->null();
+    }
+
+    public function specifyTypes(
+        FunctionReflection $functionReflection,
+        FuncCall $node,
+        Scope $scope,
+        TypeSpecifierContext $context
+    ): SpecifiedTypes {
+        return $this->typeSpecifier->specifyTypesInCondition($scope, $node->args[0]->value, TypeSpecifierContext::createFalsey());
```
ondrejmirtes: This will crash if someone calls `abort_if` without any argument. The extension should account for any wrong usage, returning empty `new SpecifiedTypes()`.

nunomaduro: cc @canvural

PR #333, opened 2019-10-17
Check argument count in abort_if extension

canvural:
As per the comment from @ondrejmirtes, this PR adds check for the argument count. `abort_if` requires at least 2 arguments, so it checks if it's less than 2.

### DIFF
--- a/src/Types/AbortIfFunctionTypeSpecifyingExtension.php
+++ b/src/Types/AbortIfFunctionTypeSpecifyingExtension.php
@@ -41,6 +41,10 @@ final class AbortIfFunctionTypeSpecifyingExtension implements FunctionTypeSpecif
         Scope $scope,
         TypeSpecifierContext $context
     ): SpecifiedTypes {
+        if (count($node->args) < 2) {
+            return new SpecifiedTypes();
+        }
+
         return $this->typeSpecifier->specifyTypesInCondition($scope, $node->args[0]->value, TypeSpecifierContext::createFalsey());
     }
 


src/Types/AbortIfFunctionTypeSpecifyingExtension.php
```diff
@@ -41,6 +41,10 @@ public function specifyTypes(
         Scope $scope,
         TypeSpecifierContext $context
     ): SpecifiedTypes {
+        if (count($node->args) < 2) {
+            return new SpecifiedTypes();
```
nunomaduro: What will happen here? Thanks for all your work so far.

szepeviktor: Could it be the handling of invalid call to `abort_if()`?

canvural: This is to prevent crashing this extension.

This will return empty `SpecifiedTypes` But later `phpstan` will give the user an error like `Expected 2 arguments 0 given`

canvural: For [this](https://github.com/nunomaduro/larastan/pull/330#discussion_r335929487) comment

nunomaduro: Confirmed, the number of arguments has to be at least 2: ```function abort_if($boolean, $code, $message = '', array $headers = [])```.

nunomaduro:
@graham-campbell Do you know why styleci seems blocked? 👍

szepeviktor:
Is @GrahamCampbell his username?

szepeviktor:
BTW [my profession](https://github.com/szepeviktor/debian-server-tools/blob/master/CV.md) is to make applications like styleci run 💯 % of the time

PR #341, opened 2019-10-23
Adds support for resolving type from self and static

canvural:


szepeviktor:
What a fanatic achievement!

Issue #338, opened 2019-10-22
User code is getting executed


canvural:
- Larastan Version: 0.4.3
- `--level` used: 5

I have a project with Laravel Passport installed.

And when I analyze it with larastan it gives me an error like this `Internal error: Key path "file:///var/www/html/storage/oauth-public.key" does not exist or is not readable`  This is an exception from either Passport or League\OAuth2 package.

@CyberiaResurrection could this be because of your change to use `ClassMapGenerator` ?

CyberiaResurrection:
@canvural , I'm not sure off the top of my head - been in other things recently.  Does pain persist with the last version before #318 landed?

szepeviktor:
Actually phpstan and larastan are not totally **static** analyzers.

A fully static analyzer should use the PHP AST extension and not load class files.

canvural:
Ok I have some more information. @CyberiaResurrection yes issue is there also in previous versions. 0.4.0, 0.4.1 

But strange thing is I only get this error when `level` is 3 or higher. In `level` 2 everything is fine.

CyberiaResurrection:
@canvural , the obvious question would seem to be what does level 3 do (on version 0.4.0 or whatever) that level 2 does not?

nunomaduro:
cc @ondrejmirtes

canvural:
Ok. I tracked down the error to [this line](https://github.com/nunomaduro/larastan/blob/master/src/Methods/Pipes/Auths.php#L61). We are using the actual Laravel container to resolve the implementation of the interface. And yeah, this is bad :smile: 

Maybe we can just return the interface here.

CyberiaResurrection:
@canvural , @nunomaduro , and others who know a lot more than I do:

Would it be possible to do a "dry" resolve?  Rather than the whole three-ring circus, go far enough to dig out the resolved object's class name?

canvural:
No, it's dynamic.

 Here, it's better to just use the interface. Since the all actual implementations implement this interface.

canvural:
@CyberiaResurrection do you wanna take this? We just need to change [this line] (https://github.com/nunomaduro/larastan/blob/master/src/Methods/Pipes/Auths.php#L61) to use the interface. And update the tests.

CyberiaResurrection:
ok, @canvural , I'll have a bang at this.  Does anyone have any objections to me doing what I can to push unit test coverage up while I'm at it?

canvural:
[...]

PR #342, opened 2019-10-24
Use Guard contract directly rather than trying to resolve it.

CyberiaResurrection:
Trying to resolve an issue @canvural tripped over in #338 - user code running when it shouldn't be.


canvural:
What about [this] (https://github.com/nunomaduro/larastan/blob/master/tests/Features/ReturnTypes/AuthExtension.php#L22) test?

CyberiaResurrection:
@canvural , https://github.com/nunomaduro/larastan/blob/8017aa4f6436618172158e118f58936e43973470/tests/Features/ReturnTypes/Helpers/AuthExtension.php ?

canvural:
Yes, you changed the test for `auth` helper function. Doesn't the `Auth` facade test needs changing?

CyberiaResurrection:
Good catch - thank you.

szepeviktor:
Larastan flies with a never before seen speed.

# ✈️ 

@nunomaduro

PR #343, opened 2019-10-24
Squelch all resolution exceptions

CyberiaResurrection:
Keep nibbling away at #338.

### DIFF
--- a/src/Concerns/HasContainer.php
+++ b/src/Concerns/HasContainer.php
@@ -13,11 +13,9 @@ declare(strict_types=1);
 
 namespace NunoMaduro\Larastan\Concerns;
 
-use ReflectionException;
 use Illuminate\Container\Container;
-use Psr\Container\NotFoundExceptionInterface;
-use Illuminate\Contracts\Container\BindingResolutionException;
 use Illuminate\Contracts\Container\Container as ContainerContract;
+use Throwable;
 
 /**
  * @internal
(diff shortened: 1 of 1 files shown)


CyberiaResurrection:
Speaking of cleaning up unused imports, @canvural , would you say no to another PR that hooks up Scrutinizer checks for unused imports and cleans up what's already there?

CyberiaResurrection:
The immediate use case would be to catch omissions like this.  The general idea behind the approach is to divide labour between silicon and synapse, and automate what can(should?) be automated.

nunomaduro:
> The immediate use case would be to catch omissions like this. The general idea behind the approach is to divide labour between silicon and synapse, and automate what can(should?) be automated.

Can you re-phrase? Didn't understood.

CyberiaResurrection:
@nunomaduro , I'll try.

The immediate reason is to catch omissions like what I made in this pull req originally.

The longer-term reason is to reduce the effort required on the maintainers' part to review incoming patches - thus making whatever remaining review easier, more timely and more likely to happen.  I'm definitely _not_ saying "turn all the Scrutiniser checks on and up to 11 IMMEDIATELY".

CyberiaResurrection:
@canvural , @nunomaduro , discussions about Scrutiniser config changes aside, is this PR good to merge up?

nunomaduro:
Yeah - probably I am not getting everything because I am in a conference. But I still don't understand the user case for this. Why we need this?

canvural:
resolve method uses the actual container from the users application. And resolving some interface through the container may sometimes have side effects. Look at issue #338

So by catching all the exceptions, we make sure the analyses doesn't crash.

nunomaduro:
That's not a good solution. We should understand, instead, why are we resolving things from the container that can fail. Not that, if resolving things from the container a static analysis time, they may also fail at runtime for the user. 

Let's make sure we address this issue properly.

canvural:
OK so I'll close this and let's discuss it at #338

PR #347, opened 2019-10-25
Fixes method calls on facade fakes

### LINKED ISSUE #281: False positives with Facade Fakes
- Larastan Version: 0.3.17
- `--level` used: 1

### Description

When you use Facade fakes it gives false positives.

For example following code

```php
Event::fake([UserActivated::class]);
...
Event::assertDispatched(UserActivated::class);
```

gives `Call to an undefined static method Illuminate\Support\Facades\Event::assertDispatched().`

This one is tricky. I'm not sure how it can be solved.

Currently I solved it by adding `#Call to an undefined static method Illuminate\\Support\\Facades\\[a-zA-Z0-9\\_]+::assert[a-zA-Z0-9_]+#` to my `ignoreErrors`

canvural:
Fixes #281

### DIFF
--- a/src/Methods/Pipes/Facades.php
+++ b/src/Methods/Pipes/Facades.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace NunoMaduro\Larastan\Methods\Pipes;
 
 use Closure;
+use Illuminate\Support\Str;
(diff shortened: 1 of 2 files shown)


szepeviktor:
Thank you.

nunomaduro:
Good job!

PR #361, opened 2019-11-05
Adds support for resolving types of model relations accessed as property

canvural:
This PR adds support for specifying types for model relations accessed as model properties.

`$user->posts` for example, returns a collection of posts.

There is one thing though. I had to remove `Illuminate\Database\Eloquent\Model` from the `universalObjectCratesClasses` for this to work. So this can result in some new errors for the users. Because now all the property accesses will be checked by PHPStan. But as  I said, this is also required for this extension to work.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -1,7 +1,6 @@
 parameters:
     scopeClass: NunoMaduro\Larastan\Analyser\Scope
     universalObjectCratesClasses:
-        - Illuminate\Database\Eloquent\Model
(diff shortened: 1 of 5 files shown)


nunomaduro:
Awesome job. Thanks!

nunomaduro:
@canvural Don't hesitate in adding your github sponsors page into the funding github.

szepeviktor:
https://github.com/nunomaduro/larastan/blob/master/.github/FUNDING.yml

PR #362, opened 2019-11-07
Fix return type of mock, partialMock and spy methods in TestCase

BertvanHoekelen:
This PR add support for the return type of the `mock`, `partialMock` and `spy` methods in `Illuminate\Foundation\Testing\TestCase`.

Without it you get the following error when calling a method.
`Call to an undefined method  Mockery\MockInterface::someMethod()`

Or when `phpstan/phpstan-mockery` is installed and you declare it as a property:
`Tests\FooTest::$mock (App\Foo&Mockery\MockInterface) does not accept Mockery\MockInterface.`

### DIFF
--- a/tests/Features/ReturnTypes/TestCaseExtension.php
+++ b/tests/Features/ReturnTypes/TestCaseExtension.php
@@ -6,26 +6,48 @@ namespace Tests\Features\ReturnTypes;
 
 use App\User;
 use Illuminate\Foundation\Application;
-use PHPStan\Testing\TestCase;
 
-class TestCaseExtension extends TestCase
+class TestCaseExtension
 {
-    public function testTestCaseExtension(): void
+    public function testMockMethod(): void
     {
-        $testCase = new TestTestCase();
+        (new TestTestCase())->testMockMethod();
+    }
 
-        $testCase->testMockingMethod('mock');
-        $testCase->testMockingMethod('partialMock');
-        $testCase->testMockingMethod('spy');
+    public function testPartialMockMethod(): void
+    {
+        (new TestTestCase())->testPartialMockMethod();
+    }
+
+    public function testSpyMethod(): void
+    {
+        (new TestTestCase())->testSpyMethod();
     }
 }
 
 class TestTestCase extends \Illuminate\Foundation\Testing\TestCase
 {
-    public function testMockingMethod($method): void
+    public function testMockMethod(): void
     {
-        if (method_exists($this, $method)) {
-            $mock = $this->{$method}(User::class);
+        if (method_exists($this, 'mock')) {
+            $bla = 'mock';
+            $mock = $this->{$bla}(User::class);
+            $mock->accounts();
+        }
+    }
+
+    public function testPartialMockMethod(): void
+    {
+        if (method_exists($this, 'partialMock')) {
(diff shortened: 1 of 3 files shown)


szepeviktor:
Thank you for your contribution. And specially for the tests!

tests/Features/ReturnTypes/TestCaseExtension.php
```diff
@@ -0,0 +1,38 @@
...
+
+declare(strict_types=1);
+
+namespace Tests\Features\ReturnTypes;
+
+use App\User;
+use Illuminate\Foundation\Application;
+use PHPStan\Testing\TestCase;
+
+class TestCaseExtension extends TestCase
```
canvural: There is no need to extend the `TestCase` here. We don't really use it.

tests/Features/ReturnTypes/TestCaseExtension.php
```diff
@@ -0,0 +1,38 @@
...
+        $testCase->testMockingMethod('spy');
+    }
+}
+
+class TestTestCase extends \Illuminate\Foundation\Testing\TestCase
+{
+    public function testMockingMethod($method): void
+    {
+        if (method_exists($this, $method)) {
+            $mock = $this->{$method}(User::class);
```
canvural: You are calling the methods here dynamically. So when we test statically we can't determine if the extension is really working or not. If I remove the extension from `extension.neon` tests are still passing.

tests/Features/ReturnTypes/TestCaseExtension.php
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Tests\Features\ReturnTypes;
+
+use App\User;
+use Illuminate\Foundation\Application;
+use PHPStan\Testing\TestCase;
+
+class TestCaseExtension extends TestCase
+{
+    public function testTestCaseExtension(): void
```
canvural: Let's have different methods here for each method we are testing. Like `testSpy`, `testMock` etc.

tests/Features/ReturnTypes/TestCaseExtension.php
BertvanHoekelen: You're absolutely right, I had to change it because tests were failing on old Laravel version. (no partialMock method). I will create another PR that will test if disabling extensions results in an error so that is covered as well.

tests/Features/ReturnTypes/TestCaseExtension.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        (new TestTestCase())->testSpyMethod();
+    }
+}
+
+class TestTestCase extends \Illuminate\Foundation\Testing\TestCase
+{
+    public function testMockMethod(): void
+    {
+        if (method_exists($this, 'mock')) {
```
canvural: Do we need this `method_exists` checks?

BertvanHoekelen: I can remove the method_exists check for mock and spy. But for `partialMock ` I have to check if the method exists on `TestCase` because the testsuite of 7.1 install an older version of Laravel where the method doesn't exists. See [travis build](https://travis-ci.org/nunomaduro/larastan/builds/608719982?utm_source=github_status&utm_medium=notification)

PR #363, opened 2019-11-07
Verify analyse fails when extension is disabled.

BertvanHoekelen:
When working on https://github.com/nunomaduro/larastan/pull/362 test's succeeded while disabling the extension didn't change anything.

To make reviews easier without manually enabling/disabling the extension, I have added a test verifying the extension is really working or not.

I also added a test that checks if the extension is registered in the `extensions.neon` file.

### DIFF
--- a/tests/ExtensionsTest.php
+++ b/tests/ExtensionsTest.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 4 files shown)


nunomaduro:
@BertvanHoekelen Can you fix the conflicting files and redo this pull request?

PR #364, opened 2019-11-08
Return custom builder if method exists on builder

BertvanHoekelen:
When chaining other query builder methods on custom builder scopes you get: `Method Illuminate\Support\Collection::METHOD() invoked with 0 parameters,  1-2 required.` or something similar. This PR makes sure that the custom builder is returned when the scope is called.

I also added extra tests: verify `Builder` is returned for local scopes, verify model is returned for creation/retrieval methods and the default return collection.


canvural:
Thank you!

PR #365, opened 2019-11-08
Added model creation, model retrieval and get method eloquent builder tests

BertvanHoekelen:
Add extra tests for EloquentBuilderExtension


canvural:
We already have [tests](https://github.com/nunomaduro/larastan/blob/master/tests/Features/Methods/ModelExtension.php#L165-L179) like this. So, I don't see the benefit here.

BertvanHoekelen:
Ok, I can remove `testAModelRetrievalMethodReturnsModel`. But the other ones actually provide test coverage. As you can see in my last commit there was already an error with earlier Laravel versions.

canvural:
Like I said, we already have tests like this. The previous error was because phpdoc block of `make` is fixed in newer Laravel versions. But it still fails with 5.8 We can create a new issue for that. I'm closing this one.

PR #367, opened 2019-11-08
Allow nullable model relation

BertvanHoekelen:
This PR fixes errors like `Call to function is_null() with DummyModelWithDocBlock will always evaluate to false.` when the property is explicitly marked as nullable by the phpdoc comment.

### DIFF
--- a/src/Properties/ModelRelationsExtension.php
+++ b/src/Properties/ModelRelationsExtension.php
@@ -28,8 +28,6 @@ use PHPStan\Type\IterableType;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\UnionType;
-use PHPStan\Type\NullType;
-
 
 /**
  * @internal
@@ -42,6 +40,10 @@ final class ModelRelationsExtension implements PropertiesClassReflectionExtensio
     public function hasProperty(ClassReflection $classReflection, string $propertyName): bool
     {
         if (! $classReflection->isSubclassOf(Model::class)) {
+            return false;
+        }
+
+        if ($this->hasClassProperty($classReflection, $propertyName)) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


szepeviktor:
Thank you.

> when the property is explicitly marked as nullable by the phpdoc comment.

Could add a test exactly like that?

szepeviktor:
Oh! You've just did.

```php
/**
 * @property \Tests\Features\Properties\DummyModel|null $belongsToRelation
 */
```

nunomaduro:
@canvural Can you check this one?

BertvanHoekelen:
Yes Laravel always returns `null` when the relation is empty, so I can see why always returning the Union type is a good idea and might be the solution.

But I think that having errors like `Call to function is_null() with DummyModelWithDocBlock will always evaluate to false.` for relationships that cannot be null is really useful. Because you explicitly must add `nullable` to the field in the migration.

canvural:
If you have annotation for the relation in your class, `ModelRelationsExtension` extension will not even run. So you can customize it however you want.

So I think we can just return `UnionType` everytime with `NullType` added. And this should only apply to single relations. For example `belongsTo`, `hasOne` etc.  If `hasMany` relation is empty, an empty collection is returned. So that's fine now.

BertvanHoekelen:
I have annotations for the properties in my class, which are automatically generated by the ide-helper. I really don't want to add annotations for the methods because the methods themselves don't return null but a relation object.

I also want larastan to report errors for properties that can't be null, so I am closing this PR and will continue to use my own fork.

canvural:
Why would you need to add annotation for the relation method? If you already have annotation for the relation property as `Model|null` the extension would not even run and use the `Model|null` type.

BertvanHoekelen:
Disabling [these](https://github.com/nunomaduro/larastan/pull/367/files#diff-63083a7620ca9ffcf86d6ef41e9af3c3R80-R87) lines result in the following error: `tests/Features/Properties/ModelRelationsExtension.php:79:Ternary operator condition is always true.` So the extension is definitely running and not using the `Model|null` phpdoc.

BertvanHoekelen:
I now see your point. The extension shouldn't have run. I have changed the code to not run the extension when the property is defined in the phpdoc, however shouldn't this be the task of phpstan?

canvural:
PHPStan has a specific order of how it runs the extensions. `AnnotationsPropertiesClassReflectionExtension` extension runs after any 3rd party extensions. So our `ModelRelationsExtension` was running before it, so overriding it with the annotations was not possible. I fixed this in #369 So I'm closing this PR in favor of that.

But thank you for your time and effort to bring up this issue!

PR #387, opened 2019-12-22
Update composer.json

GrahamCampbell:


szepeviktor:
No more Laravel 5 support? 😿

GrahamCampbell:
> No more Laravel 5 support? 😿

Laravel 5 is (totally, or almost totally) no longer maintained (even 5.5 LTS), and people should upgrade to Laravel 6 LTS as soon as possible.

Laravel 5 support is not being dropped anyway, it will be available in the old version of the library, in the same way laravel 5 is available in the old version of the laravel framework, as it were. ;)

nunomaduro:
Yeah - @GrahamCampbell is right.

PR #391, opened 2019-12-22
Removes Artisan command

canvural:


canvural:
I'm not sure why Scrutinizer tests are failing.

nunomaduro:
Just removed the service, we don't use it anyway.

szepeviktor:
Scrutinizer on PHP 7.3 is able to install packages dues to unsolvable version constrains https://scrutinizer-ci.com/g/nunomaduro/larastan/inspections/6805790a-6eac-42b7-9474-226f1359cc78/configs

gazben:
@szepeviktor Can we get a summary why was this feature removed?

szepeviktor:
It had no additional value.
It had started PHPStan with a simple config file.
Now you have to create that file: https://github.com/nunomaduro/larastan/blob/master/README.md#-getting-started-in-3-steps

mfn:
Also: it did interfered with expectations of how phpstan itself works (this was most notably for users which had an existing phpstan with a config, then added larastan and then it didn't work due to certain environment setup regarding file paths was different => all fixed == gone now)

PR #392, opened 2019-12-22
Proper env var name in CI

szepeviktor:
`LARAVEL_VERSION`

and the colon: https://getcomposer.org/doc/03-cli.md#require


nunomaduro:
Nice catch!

PR #394, opened 2019-12-23
Simplify replacing type logic in ModelTypeHelper

canvural:
This PR simplifies the logic in `ModelTypeHelper` according to the @ondrejmirtes's comment [here](https://github.com/phpstan/phpstan/issues/2766#issuecomment-568594782).


nunomaduro:
Add a line on the changelog. And then you can merge it.

Issue #396, opened 2019-12-27
Access to an undefined property for custom accessors


johanvanhelden:
- Larastan Version: 0.5.0
- `--level` used: 5

### Description

I have a model with a few attributes that are not being recognized. Both "normal" database column attributes and accessors.

Here is an issue with normal columns:
```
 ------ ----------------------------------------------------------------- 
  Line   Actions/Subscriber/CreateSubscriber.php                          
 ------ ----------------------------------------------------------------- 
  35     Access to an undefined property App\Models\Subscriber::$uuid.    
  36     Access to an undefined property App\Models\Subscriber::$secret.  
 ------ ----------------------------------------------------------------- 
```

And here are some issues with accessors:
```
 ------ ---------------------------------------------------------------------------------- 
  Line   Http/Controllers/SubscriberController.php                                         
 ------ ----------------------------------------------------------------------------------       
  84     Access to an undefined property App\Models\Subscriber::$is_confirmed.             
  94     Access to an undefined property App\Models\Subscriber::$manage_subscription_url.  
 ------ ---------------------------------------------------------------------------------- 
```

These issues started showing up right after I upgraded to 0.5.0 from 0.4.3. They show up everywhere I use them in the codebase, so they are not only showing up in the examples above, but these should give some solid examples.

## Actions/Subscriber/CreateSubscriber.php  
```php
/**
 * Perform the action.
 *
 * @param array $data
 */
public function execute(array $data)
{
    $subscriber = new Subscriber($data);

    $subscriber->uuid = Str::uuid()->toString();
    $subscriber->secret = hash_hmac('sha256', Str::random(40), config('app.key'));

    $subscriber->save();

    $this->confirmationAction->execute($subscriber);
}
```

## Http/Controllers/SubscriberController.php
```php
/**
 * Executes the request to update a subscriber.
 *
 * @param string           $uuid
 * @param string           $secret
 * @param UpdateRequest    $request
 * @param UpdateSubscriber $action
 *
 * @return \Illuminate\Http\RedirectResponse
 */
public function update(string $uuid, string $secret, UpdateRequest $request, UpdateSubscriber $action)
{
    $subscriber = $this->getSubscriber($uuid, $secret);

    if (!$subscriber->is_confirmed) {
        return $this->handleNotConfirmed($subscriber);
    }

    $action->execute($subscriber, $request->validated());

    $subscriber->refresh();

    return redirect($subscriber->manage_subscription_url);
}
```

[...]

PR #401, opened 2019-12-27
Support for model accessors

canvural:
This PR adds support for Eloquent [model accessors](https://laravel.com/docs/6.x/eloquent-mutators#defining-an-accessor).


szepeviktor:
Related to #396 ?

PR #403, opened 2020-01-01
Fixed README note about max level

GrahamCampbell:
Level 8 is the current max: https://github.com/phpstan/phpstan-src/blob/master/conf/config.level8.neon.


nunomaduro:
Thanks!

PR #404, opened 2020-01-01
Fixed incorrect minimum php version

GrahamCampbell:
Because this package depends on the illuminate 6/7 components, it cannot be installed on PHP 7.1. The real minimum version is 7.2.0.


GrahamCampbell:
This PR actually changes nothing, since the php require line could technically be deleted. It's confusing having it lower than what is actually going to be allowed though, so we should change it.

nunomaduro:
Thanks!

PR #405, opened 2020-01-01
WIP: larastan actually requires the entire framework

GrahamCampbell:
### Primarily opened this PR for discussion, rather than consideration for merge.

This has the consequence of breaking Lumen support...

---

So, the reason for this PR is I have PHPStan installed using `bamarni/composer-bin-plugin`, which places it in a different place, much like if I were to download the phar file. I can add other phpstan extensions just fine, into this separate area, but not larastan. It crashes with:

```
Note: Using configuration file /data/phpstan.neon.dist.
Class 'Dotenv\Repository\Adapter\EnvConstAdapter' not found
```

The separate composer.json file is:

```json
{
    "require": {
        "phpstan/phpstan": "^0.12.3",
        "phpstan/extension-installer": "^1.0.3",
        "nunomaduro/larastan": "^0.5.0"
    }
}
```


nunomaduro:
On the installation process, we have the following note: `Using Larastan for analysing Laravel packages? You may need to install orchestra/testbench.`.

If you are installating `larastan` in a enviroment without a the laravel framework, you may need to install `orchestra/testbench` with the same version of your lumen app.

szepeviktor:
> much like if I were to download the phar file.

Since 0.12 PHPStan is distributed as a PHAR file 🎉

GrahamCampbell:
> Since 0.12 PHPStan is distributed as a PHAR file 🎉

I suppose, I meant, distributed outside of composer. ;)

GrahamCampbell:
> On the installation process, we have the following note

We should probably note this requirement for installs via `bamarni/composer-bin-plugin`, too.

PR #407, opened 2020-01-02
Added note about bamarni/composer-bin-plugin

GrahamCampbell:


szepeviktor:
plus one typo in similarly

README.md
```diff
@@ -32,7 +32,7 @@ Larastan was created by [Nuno Maduro](https://github.com/nunomaduro), got artwor
 composer require --dev nunomaduro/larastan
 ```
 
-> Using Larastan for analysing Laravel packages? You may need to install `orchestra/testbench`.
+> Using Larastan for analysing Laravel packages? You may need to install `orchestra/testbench`. Similarly, if you have PHPStan and Larastan installed separetely, together, using `bamarni/composer-bin-plugin`.
```
mfn: Not native English, here, what does this mean? :)
> Similarly … separetely, together, using…

I mean I've an idea what it tries to say… 🤷‍♀️

GrahamCampbell: PHPStan and Larastan have to be installed together, but both separately from the main composer.json.

nunomaduro: @GrahamCampbell I would like to keep the readme very very clean. Do you think the amount of people using `bamarni/composer-bin-plugin` is a considerable high that we need have this note here?

szepeviktor: As I've learned very small number of users read the README :)

nunomaduro:
I don't think the amount of people using this actually worth a note on the readme. Thanks anyway @GrahamCampbell !

Issue #409, opened 2020-01-07
Custom builder for relation methods


BertvanHoekelen:
- Larastan Version: 0.5
- `--level` used: 5

### Description
Calling custom builder methods on a relationship works in Laravel but larastan shows the following error: `Method Illuminate\Database\Eloquent\Model::uncategorized() invoked with 1 parameters, 0 required.`. 

### Laravel code where the issue was found for example:

```php
$uncategorizedTopics = $user->topics()->uncategorized(true)->get();
```

I would love to contribute but I have no idea where to start.
My idea is that whenever a `Relation` class is returned from a function it should check the class given in the relationship for a custom builder and return an `IntersectionType` of the relations class (e.g. `HasMany`) and the custom builder.

Changing the test `EloquentBuilderExtension` to the following should test it I believe.
```php
<?php

declare(strict_types=1);

namespace Tests\Features\ReturnTypes;

use Illuminate\Database\Eloquent\Builder;
use Illuminate\Database\Eloquent\Model;
use Illuminate\Database\Eloquent\Relations\HasMany;

class EloquentBuilderExtension
{
    public function testModelScopeReturnsBuilder(): Builder
    {
        return TestScopesModel::query()
            ->foo('piet');
    }

    public function testCustomBuilderMethodReturnsBuilder(): CustomBuilder
    {
        return TestScopesModel::query()
            ->type('piet');
    }

    public function testRelationReturnsCustomBuilder(): CustomBuilder
    {
        return TestScopesModel::testScopes();
    }
}

class TestScopesModel extends Model
{
    public function scopeFoo(string $foo) : Builder
    {
        return $this->where(['foo' => $foo]);
    }

    public function testScopes(): HasMany
    {
        return $this->hasMany(TestScopesModel::class);
    }

    /**
     * @param \Illuminate\Database\Query\Builder $query
     *
     * @return \Tests\Features\ReturnTypes\CustomBuilder
     */
    public function newEloquentBuilder($query)
    {
        return new CustomBuilder($query);
    }
}

class CustomBuilder extends Builder
{
    public function type(string $type): void
    {
        $this->where(['type' => $type]);
    }
}

```

szepeviktor:
Very nice bug report.

canvural:
Hi,

Thanks for the report!

The error you got is a little weird. It thinks `uncategorized` method is on `Model` class. Can you give a full example of the code?

[...]

PR #410, opened 2020-01-08
Return builder after relations

BertvanHoekelen:
As discussed in #409. 

I had to put it before the `Mixins` pipe. Otherwise the `Builder` mixin on `Relation` would handle it and stop the pipe from executing.

The only thing that doesn't work is `(new User())->accounts()->whereFoo('bar');`. I didn't want to copy the code from `BuilderDynamicWheres` so let's discuss how to handle that.


canvural:
Thank you! It's great.

But I was wrong in my comment [here](https://github.com/nunomaduro/larastan/issues/409#issuecomment-571968706). `$user->accounts()->where('foo', 'bar')` should return the relation, not the builder. Same for `$user->accounts()->whereIn('id', [1, 2, 3])` I tested this in Tinker in a live project, and indeed the relation is returned not the builder.

Can you fix that?

BertvanHoekelen:
Done!

Do you have any suggestions on how we can handle those dynamic wheres? Or should we leave it for now and create an issue as feature request?

src/Methods/Pipes/BuilderAfterRelations.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * @param class-string $builderClass
+     */
+    private function getMethodReflectionForBuilder(PassableContract $passable, string $builderClass): ?EloquentBuilderMethodReflection
+    {
+        $eloquentBuilder = $passable->getBroker()->getClass($builderClass);
+
+        if (! $eloquentBuilder->hasNativeMethod($passable->getMethodName())) {
```
canvural: You can do an additional check here like `! Str::startsWith($passable->getMethodName(), 'where')` (not sure of the boolean logic. You can check) So we will allow the dynamic wheres also.

BertvanHoekelen: Yeah that won't work. You will then get the following error:
 `Method Illuminate\\Database\\Eloquent\\Model::whereFoo() invoked with 1 parameter, 0 required.`. 

If you take a look at [`BuilderDynamicWheres`](https://github.com/nunomaduro/larastan/blob/master/src/Methods/Pipes/BuilderDynamicWheres.php) you will see that there is much more needed to support those dynamic where's. 

So maybe we should create a `BuilderHelper` where we can return method reflection for that dynamic where. And call that in both `BuilderDynamicWhere` as `BuilderAfterRelations`. And while writing this I think we should add support for [local macros](https://github.com/nunomaduro/larastan/blob/master/src/Methods/Pipes/BuilderLocalMacros.php) on the relationships. Which we can also add to that `BuilderHelper`

canvural: Ok, you can give that a go :+1:

BertvanHoekelen: I think this is a nice solution. `BuilderLocalMacros` is depending on the model which [as discussed](https://github.com/nunomaduro/larastan/issues/409#issuecomment-571961110) I don't have access to.

canvural:
Looks good! Thank you!

Issue #285, opened 2019-07-10
Is a docker image planned?


Einenlum:
Is it planned at some point to have a docker image, as [phpstan does](https://github.com/phpstan/docker-image)?

ondrejmirtes:
It does not make sense to be provided by 3rd party in this case. If you want Larastan inside Docker, inherit the PHPStan image and install Larastan inside it.

szepeviktor:
Yes, that seems to be easy https://github.com/phpstan/docker-image/blob/master/Dockerfile#L39

szepeviktor:
```dockerfile
FROM phpstan/phpstan

RUN composer global require nunomaduro/larastan
```

No! It is not that easy as Larastan needs to be inside your Laravel project!

ondrejmirtes:
You can mount it as a volume.

szepeviktor:
Larastan does not start without a Laravel project.

https://github.com/nunomaduro/larastan/blob/master/composer.json#L47-L52

There is no PHAR distribution of Larastan.

deleugpn:
I just hit this one, it's quite a bummer. I'm trying to run larastan on a project from outside the project itself. The goal is to gain velocity. If I run it from inside the project itself, I need to wait for the entire project Docker to build (which takes longer).
Larastan doesn't seem to work as a standalone project at all.

canvural:
Hi @deleugpn 

How are you running it? Did you mount your project as a volume inside the docker? Did you check the PHPStan [docs](https://github.com/phpstan/phpstan#install-phpstan-extensions) about Docker?

deleugpn:
@canvural I tried the phpstan Docker solution by running

```
docker run --rm -v $(pwd):/app phpstan/phpstan analyse
```

but I get the following error:

```
[05:10 AM]-[root@docker]-[/home/customergauge/analytics-api]-[git larastan]
# docker run --rm -v $(pwd):/app phpstan/phpstan analyse
Note: Using configuration file /app/phpstan.neon.
Bootstrap file /composer/vendor/phpstan/phpstan/../../nunomaduro/larastan/bootstrap.php does not exist.
```

------------------------------------------------------------------------------------------------

I also tried running from a complete clean folder the following: 

```
composer require nunomaduro/larastan
./vendor/bin/phpstan analyse -c /path/to/project/phpstan.neon
```

But I get an error saying 

```
[...]

PR #416, opened 2020-01-15
Allow Larastan to work from PHPStan Docker image

deleugpn:
While trying to run Larastan from a Docker image (https://github.com/nunomaduro/larastan/issues/285), I was getting the error 

```
docker run --rm -v $(pwd):/app phpstan/phpstan analyse
Note: Using configuration file /app/phpstan.neon.
Bootstrap file /app/../../nunomaduro/larastan/bootstrap.php does not exist.
```

The reason for this is because the Docker image provided by phpstan has phpstan installed globally. When trying to navigate the folder structure, we end up looking for the bootstrap file in the global composer folder instead of the local one. By changing it to `currentWorkingDirectory`, I managed to successfully run Larastan from the PHPStan Docker. I also managed to run it from within the project itself as a test case and it seems to be working fine as well.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -13,7 +13,7 @@ parameters:
         - '#Property App\\Http\\Middleware\\TrustProxies::\$headers \(string\) does not accept default value of type int#'
         # Ignore error from issue #244
         - '#Cannot call method viaRequest\(\) on Illuminate\\Auth\\AuthManager\|null\.#'
-    bootstrap: %rootDir%/../../nunomaduro/larastan/bootstrap.php
+    bootstrap: %currentWorkingDirectory%/vendor/nunomaduro/larastan/bootstrap.php
     reportUnmatchedIgnoredErrors: false
 
 services:


deleugpn:
I just learned that the extension neon files seems to behave like a recursive merge. By adding `bootstrap: %currentWorkingDirectory%/vendor/nunomaduro/larastan/bootstrap.php` to my own project, the problem was also fixed. I imagine I'm being able to override larastan's configuration by doing that, which makes this PR not necessary to Larastan itself.

I'll still leave it open so that the maintainers can decide whether they want to consider this a good default value or not, but if anybody lands here in the future, it's nice to know that you can control this on your own project.

canvural:
Hi,

Thanks for the PR! I'm not sure if it is a good idea to change the config file for this.

cc @ondrejmirtes 

Can you review this? Is this a good solution? Or users should create a custom Dockerfile and install the extension like it's written in PHPStan [docs](https://github.com/phpstan/phpstan#install-phpstan-extensions)?

ondrejmirtes:
Hi, this change doesn't make much sense. It would obviously break Larastan for people that run PHPStan from some unexpected directory.

The paths in the config file should be resolved deterministically. Currently, if you omit the absolute file, e.g. use `bootstrap: bootstrap.php`, it's resolved as `dirname('extension.neon') . '/bootstrap.php`. Which I'd argue should always be the case.

The user can override the config with what they want anyway.

But I'd like to see the Dockerfile @deleugpn is using because I suspect they're not following the recommended way of installing extensions to a T.

deleugpn:
@ondrejmirtes I'm not using any Dockerfile. The whole point for me is not to maintain my own Dockerfile. I want to use the raw image provided by phpstan itself. Larastan is not providing any Docker image either.

ondrejmirtes:
That’s the problem. So you running PHPStan from a Docker image does not have any benefits. You need to have phpstan/phpstan and nunomaduro/larastan as part of the same Composer installation, in the same composer.json.

Why are you running PHPStan from a Docker image anyway? I don’t see many reasons to continue doing so...

deleugpn:
@ondrejmirtes this is essentially my goal:

image

I don't want the test suite to break because phpstan broke or I don't want phpstan to break because the test suite broke. Keeping it decoupled will allow me to slowly introduce and expand static analyses into the organization (we have 40 services to do this).
Given this constraint, I wanted to try and avoid having to build my Docker image in both AWS CodeBuild mostly because of time. The Docker image takes 7 minutes to build, if I can use plain phpstan Docker image, we can get static analyse feedback onto GitHub in less than 1 minute.

ondrejmirtes:
You can run PHPStan directly along your tests and just do `|| true`. That way it will be forgiving for errors. Or you can use [the baseline feature](https://medium.com/@ondrejmirtes/phpstans-baseline-feature-lets-you-hold-new-code-to-a-higher-standard-e77d815a5dff) which is the recommended way of doing things. I consider PHPStan something that should pass in the build (otherwise it's dangerous to use the code in production), but you can grandfather in existing issues, and only have new ones coming from changed and new code reported.

I also don't think the Docker image would take 7 minutes to build. You can just download the existing PHPStan image with `FROM` directive, and add `composer global require nunomaduro/larastan`. That will not take Docker 7 minutes to complete.

ondrejmirtes:
Also, you can build the image once, push it to AWS ECR and just use it 😊

PR #427, opened 2020-01-27
Fix #426

szepeviktor:
@canvural Use `args[0]->value` after checking it.


szepeviktor:
Why is it failing?

canvural:
The reason why it's failing seems to be unrelated to these changes.

PR #430, opened 2020-01-28
fix: array access on container

### LINKED ISSUE #429: Does Lumen have an internal error?
or are we unable to understand Lumen's code?

(recently in CI)
```
 ------ ---------------------------------------------- 

  Line   Providers/AuthServiceProvider.php             

 ------ ---------------------------------------------- 

  33     Offset 'auth' does not exist on               

         Illuminate\Contracts\Foundation\Application.  

 ------ ---------------------------------------------- 
```

https://github.com/laravel/lumen/blob/master/app/Providers/AuthServiceProvider.php#L33

canvural:
Fixes #429

### DIFF
--- a/tests/Features/ReturnTypes/ContainerArrayAccessTest.php
+++ b/tests/Features/ReturnTypes/ContainerArrayAccessTest.php
@@ -1,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Features\ReturnTypes;
+
+use Illuminate\Auth\AuthManager;
+use Illuminate\Contracts\Foundation\Application;
+
+class ContainerArrayAccessTest
+{
+    /** @var Application */
+    protected $app;
+
+    public function testAuth(): AuthManager
+    {
+        return $this->app['auth'];
+    }
+}
--- a/src/Analyser/Scope.php
+++ b/src/Analyser/Scope.php
@@ -47,6 +47,8 @@ class Scope extends BaseScope
             $type = \Mockery::mock($type);
             $type->shouldReceive('isOffsetAccessible')
                 ->andReturn(TrinaryLogic::createYes());
+            $type->shouldReceive('hasOffsetValueType')
+                ->andReturn(TrinaryLogic::createYes());
         }
 
         return $this->getContainer()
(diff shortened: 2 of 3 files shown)


tests/Features/ReturnTypes/ContainerArrayAccessTest.php
```diff
@@ -0,0 +1,19 @@
...
+
+use Illuminate\Auth\AuthManager;
+use Illuminate\Contracts\Foundation\Application;
+
+class ContainerArrayAccessTest
+{
+    /** @var Application */
+    protected $app;
+
+    public function testAuth(): ?AuthManager
```
nunomaduro: `auth` always exists on the container. I would prefer the test to be:

```
public function testAuth(): AuthManager
```

canvural: It's because of [this](https://github.com/nunomaduro/larastan/blob/master/src/Properties/ReflectionTypeContainer.php#L43) line. I can change it to `false`. It doesn't seem to affect anything else. But not sure.

nunomaduro: If that lines specifies what can be resolved out of the container, a null can't be relsolved I guess.

canvural: Changed.

tests/Features/ReturnTypes/ContainerArrayAccessTest.php
```diff
@@ -0,0 +1,19 @@
...
+
+use Illuminate\Auth\AuthManager;
+use Illuminate\Contracts\Foundation\Application;
+
+class ContainerArrayAccessTest
+{
+    /** @var Application */
+    protected $app;
+
+    public function testAuth(): AuthManager
```
nunomaduro: Can you also add a test just bellow, that resolves $this->app['fooooooooo']; ? I am curious about the result.

canvural: It would throw `BindingResolutionException` exception.

nunomaduro: Don't merge this yet. I want to run some tests.

PR #431, opened 2020-01-28
Add extension for app return type, including test cases.

troelsselch:
I have added a simple return type extension for the `app()` helper.
Fully qualified classes will be resolved to the class itself. Unknown types, e.g. 'sentry' or 'db', will be resolved to `MixedType`.

This will allow checking for method calls directly on app(), so that e.g.

```php
/** @var Handler $handler */
$handler = app(Handler::class);
$handler->capture($exception);
```

can be simplified to

```php
app(Handler::class)->capture($exception);
```
and still be verified by larastan.

### DIFF
--- a/src/ReturnTypes/Helpers/AppExtension.php
+++ b/src/ReturnTypes/Helpers/AppExtension.php
@@ -15,6 +15,7 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
 use PHPStan\Type\ErrorType;
+use PHPStan\Type\NeverType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
 use Throwable;
@@ -58,6 +59,6 @@ class AppExtension implements DynamicFunctionReturnTypeExtension
             return new ObjectType($expr->class->toString());
         }
 
-        return new ErrorType();
+        return new NeverType();
     }
 }
--- a/src/ReturnTypes/Helpers/AppExtension.php
+++ b/src/ReturnTypes/Helpers/AppExtension.php
@@ -15,8 +15,6 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
 use PHPStan\Type\ErrorType;
-use PHPStan\Type\MixedType;
-use PHPStan\Type\NullType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
 use Throwable;
@@ -47,7 +45,7 @@ class AppExtension implements DynamicFunctionReturnTypeExtension
                 $resolved = $this->resolve($expr->value);
 
                 if (is_null($resolved)) {
-                    return new NullType();
+                    return new ErrorType();
                 }
 
                 return new ObjectType(get_class($resolved));
@@ -60,6 +58,6 @@ class AppExtension implements DynamicFunctionReturnTypeExtension
             return new ObjectType($expr->class->toString());
         }
 
-        return new MixedType();
+        return new ErrorType();
     }
 }
--- a/src/ReturnTypes/Helpers/AppExtension.php
+++ b/src/ReturnTypes/Helpers/AppExtension.php
@@ -26,7 +26,7 @@ class AppExtension implements DynamicFunctionReturnTypeExtension
 
     public function isFunctionSupported(FunctionReflection $functionReflection): bool
     {
-        return $functionReflection->getName() === 'app';
+        return $functionReflection->getName() === 'app' || $functionReflection->getName() === 'resolve';
     }
 
     public function getTypeFromFunctionCall(
--- a/src/ReturnTypes/Helpers/AppExtension.php
+++ b/src/ReturnTypes/Helpers/AppExtension.php
@@ -1,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+namespace NunoMaduro\Larastan\ReturnTypes\Helpers;
+
+use PhpParser\Node\Expr\ClassConstFetch;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Name\FullyQualified;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\MixedType;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\Type;
+
+class AppExtension implements DynamicFunctionReturnTypeExtension
+{
+    public function isFunctionSupported(FunctionReflection $functionReflection): bool
+    {
+        return $functionReflection->getName() === 'app';
+    }
+
+    public function getTypeFromFunctionCall(
+        FunctionReflection $functionReflection,
+        FuncCall $functionCall,
+        Scope $scope
+    ): Type {
+        $className = $this->getClassName($functionCall);
+
+        if ($className && class_exists($className)) {
+            return new ObjectType($className);
+        }
+
+        return new MixedType();
+    }
+
+    /**
+     * Returns a fully qualified path to a class, or null if it is not a class.
+     */
+    private function getClassName(FuncCall $functionCall): ?string
+    {
+        if (count($functionCall->args) === 0) {
+            return null;
+        }
+
+        /** @var ClassConstFetch $value */
+        $value = $functionCall->args[0]->value;
+
+        if (! $value->class instanceof FullyQualified) {
+            return null;
+        }
+
+        return $value->class->toString();
+    }
+}
--- a/src/ReturnTypes/Helpers/AppExtension.php
+++ b/src/ReturnTypes/Helpers/AppExtension.php
@@ -16,6 +16,7 @@ use PHPStan\Reflection\FunctionReflection;
 use PHPStan\Type\DynamicFunctionReturnTypeExtension;
 use PHPStan\Type\ErrorType;
 use PHPStan\Type\MixedType;
+use PHPStan\Type\NullType;
 use PHPStan\Type\ObjectType;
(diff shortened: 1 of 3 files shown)


nunomaduro:
I don't see a reason for having a `MixedType` here. You should resolve things out of the container, and return the type of the resolved value.

src/ReturnTypes/Helpers/AppExtension.php
```diff
@@ -0,0 +1,57 @@
...
+        return new MixedType();
+    }
+
+    /**
+     * Returns a fully qualified path to a class, or null if it is not a class.
+     */
+    private function getClassName(FuncCall $functionCall): ?string
+    {
+        if (count($functionCall->args) === 0) {
+            return Application::class;;
```
canvural: Please remove one of the semicolons.

src/ReturnTypes/Helpers/AppExtension.php
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        /** @var ClassConstFetch $value */
+        $value = $functionCall->args[0]->value;
+
+        if (! $value->class instanceof FullyQualified) {
+            return null;
+        }
+
+        return $value->class->toString();
```
canvural: We can use the `HasContainer` trait and resolve the `$value` out of the container here. Also maybe a try/catch block around it, because the container can throw an exception if it can resolve the given value. In that case, we can return `ErrorType` maybe.

src/ReturnTypes/Helpers/AppExtension.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        if (count($functionCall->args) === 0) {
+            return Application::class;;
+        }
+
+        /** @var ClassConstFetch $value */
+        $value = $functionCall->args[0]->value;
+
+        if (! $value->class instanceof FullyQualified) {
+            return null;
```
canvural: If it is a `FullyQualified` name I think we can directly return this with a `new ObjectType` We don't need to resolve it out of the container.

src/ReturnTypes/Helpers/AppExtension.php
```diff
@@ -0,0 +1,57 @@
...
+        FuncCall $functionCall,
+        Scope $scope
+    ): Type {
+        $className = $this->getClassName($functionCall);
+
+        if ($className && class_exists($className)) {
+            return new ObjectType($className);
+        }
+
+        return new MixedType();
```
canvural: We shouldn't use `MixedType`. We can know the return type if we resolve it from the container.

troelsselch:
Thanks for the feedback. I have cleaned up the code and try to resolve `String_` values from the container.

I still have two cases with `MixedType()`:

* if $expr is not instance of `ClassConstFetch`, and
* if $expr->class is instance of Name

In the first case I cannot see another option.
In the latter case I could process the `$expr->class` and `$expr->name` and try to resolve that in the container. However, that requires more reflection checking in case `class` is e.g. `self`.

What are your thoughts on this?

src/ReturnTypes/Helpers/AppExtension.php
```diff
@@ -0,0 +1,58 @@
...
+        if (count($functionCall->args) === 0) {
+            return new ObjectType(Application::class);
+        }
+
+        /** @var Expr $expr */
+        $expr = $functionCall->args[0]->value;
+
+        if ($expr instanceof String_) {
+            try {
+                return new ObjectType(get_class($this->resolve($expr->value)));
```
canvural: This is not quite right. `resolve` can return `null` We should also consider that here.

src/ReturnTypes/Helpers/AppExtension.php
```diff
@@ -0,0 +1,58 @@
...
+        Scope $scope
+    ): Type {
+        if (count($functionCall->args) === 0) {
+            return new ObjectType(Application::class);
+        }
+
+        /** @var Expr $expr */
+        $expr = $functionCall->args[0]->value;
+
+        if ($expr instanceof String_) {
```
canvural: Why only strings? What about `app(1)`? I know it is a silly example. But I think we should try to resolve everything out of the container and handle the error cases (returning null or throwing exceptions) This also mimics the actual implementation of the `app` better.

src/ReturnTypes/Helpers/AppExtension.php
```diff
@@ -0,0 +1,58 @@
...
+            try {
+                return new ObjectType(get_class($this->resolve($expr->value)));
+            } catch (Throwable $exception) {
+                return new ErrorType();
+            }
+        }
+
+        if ($expr instanceof ClassConstFetch && $expr->class instanceof FullyQualified) {
```
canvural: If we gonna resolve everything out of the container this will be unnecessary.

Or if you'd like, you can move this check above the `resolve` part. So if it's fully qualified class it will be returned directly, if not we will resolve.

I'm not sure which is better.

tests/Features/ReturnTypes/Helpers/AppExtension.php
```diff
@@ -0,0 +1,29 @@
...
+    public function testObjectType(): ApplicationResolver
+    {
+        return app(ApplicationResolver::class);
+    }
+
+    public function testMixedTypeNoArgument(): Application
+    {
+        return app();
+    }
+
```
canvural: Maybe we can add another test case here. For example, checking `app('auth')` returns `AuthManager`

src/ReturnTypes/Helpers/AppExtension.php
troelsselch: I am check the type, since `$this->resolve` takes a string, and `Expr` does not have a toString method. If there is a simple way I can pass any `Expr` to the resolve method, please let me know.

src/ReturnTypes/Helpers/AppExtension.php
troelsselch: If it is possible to simply resolve the $expr, then I'll get rid of this. (Unless there is a performance win by using the qualified name.)

src/ReturnTypes/Helpers/AppExtension.php
canvural: Yeah, sorry. You are right. This looks correct. :+1:

src/ReturnTypes/Helpers/AppExtension.php
```diff
@@ -0,0 +1,58 @@
...
+use PHPStan\Type\Type;
+use Throwable;
+
+class AppExtension implements DynamicFunctionReturnTypeExtension
+{
+    use HasContainer;
+
+    public function isFunctionSupported(FunctionReflection $functionReflection): bool
+    {
+        return $functionReflection->getName() === 'app';
```
canvural: `resolve` helper function is also identical to `app`. Maybe we can also add it to here?

src/ReturnTypes/Helpers/AppExtension.php
```diff
@@ -0,0 +1,65 @@
...
+            } catch (Throwable $exception) {
+                return new ErrorType();
+            }
+        }
+
+        if ($expr instanceof ClassConstFetch && $expr->class instanceof FullyQualified) {
+            return new ObjectType($expr->class->toString());
+        }
+
+        return new MixedType();
```
nunomaduro: We should always try to avoid the Mixed type.

src/ReturnTypes/Helpers/AppExtension.php
```diff
@@ -0,0 +1,65 @@
...
+
+        /** @var Expr $expr */
+        $expr = $functionCall->args[0]->value;
+
+        if ($expr instanceof String_) {
+            try {
+                $resolved = $this->resolve($expr->value);
+
+                if (is_null($resolved)) {
+                    return new NullType();
```
canvural: I think `app` never returns null in normal usage. So here we should return also `ErrorType`. Because if we are here it means nothing is resolved from the container.

src/ReturnTypes/Helpers/AppExtension.php
canvural: Is it safe to return `ErrorType` here also? Because if we are here this far, it means nothing is resolved from the container. In normal usage, an exception would be thrown.

nunomaduro: Exactly. That is the goal of static analysis. Prevent runtime errors 👍

src/ReturnTypes/Helpers/AppExtension.php
troelsselch: You can get a valid `null` from the container if you bind a closure to the container:

```
        app()->bind('null_value', function () {
            return;
        });
        dd(app('null_value'));
```

src/ReturnTypes/Helpers/AppExtension.php
```diff
@@ -0,0 +1,65 @@
...
+                }
+
+                return new ObjectType(get_class($resolved));
+            } catch (Throwable $exception) {
+                return new ErrorType();
```
nunomaduro: Can we pass the `exception->getMessage()` to the ErrorType constructor?

src/ReturnTypes/Helpers/AppExtension.php
nunomaduro: This needs to be removed.

src/ReturnTypes/Helpers/AppExtension.php
canvural: We can't. Actually `ErrorType` is just like `MixedType`. It extends it.

src/ReturnTypes/Helpers/AppExtension.php
canvural: Actually it might be ok here. An `ErrorType` or `NeverType` would be the same here. Because if we are here this far it means nothing is resolved from the container. 

And there is no alternative to do here.

canvural: `ErrorType` can make more sense semantically.

So if you can change this here and rebase the branch, I think we are good to go with this :+1:

troelsselch:
I have rebase and updated to be using ErrorType and NeverType as suggested.

canvural:
Thank you @troelsselch ! Good first contribution!

PR #435, opened 2020-01-31
feature: Infer model property types

canvural:
This PR adds ability to infer model property types by scanning the migration files of the application. :fire:

### DIFF
--- a/src/Properties/SchemaAggregator.php
+++ b/src/Properties/SchemaAggregator.php
@@ -1,0 +1,398 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of Larastan.
+ *
+ * (c) Nuno Maduro <enunomaduro@gmail.com>
+ *
+ *  For the full copyright and license information, please view the LICENSE
+ *  file that was distributed with this source code.
+ */
+
+namespace NunoMaduro\Larastan\Properties;
+
+use PhpParser;
+
+final class SchemaAggregator
+{
+    /** @var array<string, SchemaTable> */
+    public $tables = [];
+
+    /**
+     * @param array<int, PhpParser\Node\Stmt> $stmts
+     */
+    public function addStatements(array $stmts): void
+    {
+        foreach ($stmts as $stmt) {
+            if ($stmt instanceof PhpParser\Node\Stmt\Class_) {
+                $this->addClassStatements($stmt->stmts);
+            }
+        }
+    }
+
+    /**
+     * @param array<int, PhpParser\Node\Stmt> $stmts
+     */
+    private function addClassStatements(array $stmts): void
+    {
+        foreach ($stmts as $stmt) {
+            if ($stmt instanceof PhpParser\Node\Stmt\ClassMethod
+                && $stmt->name->name === 'up'
+                && $stmt->stmts
+            ) {
+                $this->addUpMethodStatements($stmt->stmts);
+            }
+        }
+    }
+
+    /**
+     * @param PhpParser\Node\Stmt[] $stmts
+     */
+    private function addUpMethodStatements(array $stmts): void
+    {
+        foreach ($stmts as $stmt) {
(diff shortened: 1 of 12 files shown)


szepeviktor:
What a feature!

canvural:
I did not test this in a Lumen project, or on a Laravel package. If someone can do so, it'd be great!

tests/Features/Properties/ModelPropertyExtension.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    /** @var User */
+    private $user;
+
+    public function testPropertyReturnType(): int
+    {
+        return $this->user->id;
+    }
+
+    public function testDateReturnType(): ?Carbon
```
nunomaduro: Why there is optional return type?

tests/Features/Properties/ModelPropertyExtension.php
```diff
@@ -0,0 +1,37 @@
+<?php
```
nunomaduro: Would be cool have more tests here.

src/Properties/SchemaAggregator.php
```diff
@@ -0,0 +1,398 @@
...
+    }
+
+    /**
+     * @param string                $tableName
+     * @param string                $argName
+     * @param PhpParser\Node\Stmt[] $stmts
+     *
+     * @throws \Exception
+     */
+    private function processColumnUpdates(string $tableName, string $argName, array $stmts): void
```
nunomaduro: It's ok to maintain all of this?

tests/Features/Properties/ModelPropertyExtension.php
canvural: Because `created_at` is nullable.

src/Properties/SchemaAggregator.php
canvural: I think yes. I tested the code on many different Laravel projects. It is working well. Also, psalm Laravel plugin uses this. So if there is any issue we can help each other.

tests/Features/Properties/ModelPropertyExtension.php
canvural: What kind of tests? I couldn't think of anything different. There are tests mainly for normal return types, date casts (Carbon) and casting according to `$casts` property of the model.

I didn't want to write a test for each type (integer, text, json etc.) Cause mainly it's the same code.

Also outside of these cases, I tested the whole branch in many different Laravel projects.

muglug:
@canvural well done for adding!

It's a little weird to see `(c) Nuno Maduro` on code I wrote, though...

canvural:
@muglug Yeah, you are actually right. The only reason I added that is we have the condition in the [Travis](https://github.com/nunomaduro/larastan/blob/master/.travis.yml#L34) file.

I don't know what would be the better solution in this case. I'd definitely want to give you credit. @nunomaduro what do you say?

muglug:
I think that’s quite a strange condition to have, especially for a package that relies on a lot of other people’s open-source contributions.

nunomaduro:
@canvural Let's remove those headers from every file.

nunomaduro:
@muglug It's not that strange, symfony uses this convention a lot. Anyway, I don't like. Let's remove it from every file.

canvural:
@muglug I see you just [recently added](https://github.com/psalm/laravel-psalm-plugin/pull/20) MIT license to psalm Laravel plugin. (I guess just because of this pr) So I guess we can add the MIT license notice to those 3 files in our repo.

muglug:
I think it may have made sense before license templates were so usable, and before everyone started using Composer. Now it’s trivially easy to find out who authored a given line of code, and also who’s most likely to be able to support that code.

muglug:
> So I guess we can add the MIT license notice to those 3 files in our repo.

I'm fine with that – and if you had a note to tag `@muglug` on any changes you make to those files, I'd have the option to make similar changes in my repo.

PR #438, opened 2020-02-01
feat: support for changing used date class

### LINKED ISSUE #421: now() helper & CarbonImmutable
- Larastan Version: 6.11.0
- `--level` used: 5

### Description

With the new `Date` Facade, we're now able to configure Laravel to always use `CarbonImmutable`.

```php
final class CalendarServiceProvider extends ServiceProvider
{
    public function register()
    {
        Date::useMonthsOverflow();

        Date::useYearsOverflow();

        Date::use(CarbonImmutable::class);
    }
}
```

However, the docblock for the helper `now()` (as well as the Date facade itself) says that the return type will be `Illuminate\Support\Carbon`, which extends `Carbon\Carbon` and not `Carbon\CarbonImmutable`. This causes an error with the following message:

`Parameter #x $date of method \My\Class::myMethod() expects Carbon\CarbonImmutable, Carbon\Carbon given.

```php

function myFunction(\Carbon\CarbonImmutable $date)
{
}

myFunction(now());
```

canvural:
Fixes #421 

I did not find an easy way to test this. I tested it in a separate Laravel project and it works. I tried putting the `Data::use(...)` inside the `getEnvironmentSetup` in `FeaturesTest` But that didn't work.

One idea is to create a new test case just for this, and a service provider. Put `Data::use(...)` in the service provider. And register the service provider in the test with `getPackageProviders` But that seemed a really long shot :stuck_out_tongue: 

I'm open to any ideas.

### DIFF
--- a/src/ReturnTypes/Helpers/NowAndTodayExtension.php
+++ b/src/ReturnTypes/Helpers/NowAndTodayExtension.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of Larastan.
+ *
+ * (c) Nuno Maduro <enunomaduro@gmail.com>
+ *
+ *  For the full copyright and license information, please view the LICENSE
+ *  file that was distributed with this source code.
+ */
+
+namespace NunoMaduro\Larastan\ReturnTypes\Helpers;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\DynamicFunctionReturnTypeExtension;
+use PHPStan\Type\ObjectType;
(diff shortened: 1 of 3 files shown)


src/ReturnTypes/Helpers/NowAndTodayExtension.php
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getTypeFromFunctionCall(
+        FunctionReflection $functionReflection,
+        FuncCall $functionCall,
+        Scope $scope
+    ): Type {
+        return new ObjectType(get_class(now()));
```
nunomaduro: You are a smart guy. <3

crynobone:
Using define environment via annotation seems the right use-case for this: <https://packages.tools/testbench/basic/define-environment.html#using-annotation>

PR #439, opened 2020-02-02
feat: add generics for Collection

canvural:
This PR adds stubs for the Laravel `Collection`.This will allow us to infer types in the collections for example returned by relationship. Like, `$user->accounts->first()` would return `Account`

Combined with the #435 this can be really good.

This should be merged after #435, and I want to do more testing.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -1,6 +1,10 @@
 parameters:
     stubFiles:
+        - stubs/Enumerable.stub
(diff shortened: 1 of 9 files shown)


nunomaduro:
@canvural Do we need something on the core of Laravel 7 for this?

canvural:
@nunomaduro No, we don't.

PR #318, opened 2019-10-05
Extend service provider search

CyberiaResurrection:
My attempt to implement a suggestion implied in https://github.com/nunomaduro/larastan/issues/273#issuecomment-503291317 , namely getting a list of directories to check for service providers, directly from Composer data.

@szepeviktor , I wasn't sure exactly how to test the load-from-extra-search-directories without polluting production use with test code (such as a dummy service provider).  In lieu of a way to do that cleanly, I added a test that verified my changes didn't break the status quo on Larastan itself.

PR #440, opened 2020-02-03
Fix service provider detection

CyberiaResurrection:
Back in Oct 2019, my original changes (#318 )  to introduce service provider detection implicitly assumed all such providers were concrete.

My own work since has falsified that assumption, so I'm tightening the requirements here.

### DIFF
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -39,10 +39,10 @@ final class ApplicationResolver
 
         if (file_exists($composerFile)) {
             $namespace = (string) key(json_decode((string) file_get_contents($composerFile), true)['autoload']['psr-4']);
-            /* @var class-string $class */
             $serviceProviders = array_values(array_filter(self::getProjectClasses($namespace), function (string $class) use (
                 $namespace
             ) {
+                /** @var class-string $class */
                 return substr($class, 0, strlen($namespace)) === $namespace && self::isServiceProvider($class);
             }));
 
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -39,7 +39,7 @@ final class ApplicationResolver
 
         if (file_exists($composerFile)) {
             $namespace = (string) key(json_decode((string) file_get_contents($composerFile), true)['autoload']['psr-4']);
-
+            /* @var class-string $class */
             $serviceProviders = array_values(array_filter(self::getProjectClasses($namespace), function (string $class) use (
                 $namespace
             ) {
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -66,10 +66,12 @@ final class ApplicationResolver
      * @param  string $class
      *
      * @return bool
+     * @throws \ReflectionException
      */
     private static function isServiceProvider($class): bool
     {
-        return in_array(\Illuminate\Support\ServiceProvider::class, class_parents($class), true);
+        return in_array(\Illuminate\Support\ServiceProvider::class, class_parents($class), true) &&
+               !((new \ReflectionClass($class))->isAbstract());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/ApplicationResolver.php
```diff
@@ -66,10 +66,12 @@ protected function getEnvironmentSetUp(\Illuminate\Foundation\Application $app):
      * @param  string $class
      *
```
canvural: Maybe we can add here `@phpstan-param class-string $class` 

There is an error in TravisCI about this.

tests/Application/app/Providers/AbstractProvider.php
```diff
@@ -0,0 +1,16 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: alex
+ * Date: 3/02/20
+ * Time: 5:00 PM.
+ */
```
canvural: Please remove this docblock.

tests/Application/app/Providers/ConcreteProvider.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: alex
+ * Date: 3/02/20
+ * Time: 5:01 PM.
+ */
```
canvural: Please remove this docblock.

src/ApplicationResolver.php
CyberiaResurrection: What's the difference between the two?

canvural: PHPStan supports a special string type. `class-string` its the fully qualifed class name. For example `User::class` And it uses this in the places that you expect not just any string but a class name. Like in `ReflectionClass` here.

canvural: It seems this is not enough. `$class` should be documented as `class-string` also where the `isServiceProvider` method is called.

tests/Application/app/Providers/AbstractProvider.php
CyberiaResurrection: Sorry for missing that, phpstorm auto-added it.  Gone now.

tests/Application/app/Providers/ConcreteProvider.php
CyberiaResurrection: Sorry for missing that, phpstorm auto-added it.  Gone now.

src/ApplicationResolver.php
CyberiaResurrection: That's up on line 46 of ApplicationResolver - I'm not sure how to annotate the use there as you're wanting to.

canvural: You can try to put the docblock on line 42 maybe `@var class-string $class`

CyberiaResurrection: Ok, that's up, but I still had larastan complain when running locally, on line 46 of ApplicationResolver:
  46     Parameter #1 $class of static method NunoMaduro\Larastan\ApplicationResolver::isServiceProvider() expects class-string, string given.  

`

canvural: Removing it from line 42 and putting it right above the `return substr($class, 0, strlen($namespace)) === $namespace && self::isServiceProvider($class);` line, works :+1:

canvural: It should be `/**` With 2 *. That way PHPStan can recognize it.

CyberiaResurrection: Thanks for your help and patience, @canvural .  I've sent your suggested change up after confirming analyse runs clean locally.

canvural:
If you can rebase with the current master, it is good to go :+1:

CyberiaResurrection:
@canvural , that's rebased as requested - thanks again for your help and patience.

canvural:
Thank you!

PR #442, opened 2020-02-03
refactor: remove license headers from files

canvural:


szepeviktor:
@nunomaduro New era of license-free source?

src/Properties/SchemaTable.php
```diff
@@ -3,12 +3,12 @@
 declare(strict_types=1);
 
 /**
- * This file is part of Larastan.
+ * This file uses parts of code from following projects.
```
nunomaduro: I would remove this too.

src/Properties/SchemaColumn.php
```diff
@@ -3,12 +3,12 @@
 declare(strict_types=1);
 
 /**
- * This file is part of Larastan.
+ * This file uses parts of code from following projects.
```
nunomaduro: I would remove this too.

src/Properties/SchemaAggregator.php
```diff
@@ -3,12 +3,12 @@
 declare(strict_types=1);
 
 /**
- * This file is part of Larastan.
+ * This file uses parts of code from following projects.
```
nunomaduro: I would remove this too.

nunomaduro: Maybe just add a note on the class phpdocs.

src/Properties/SchemaColumn.php
nunomaduro: Maybe just add a note on the class phpdocs.

src/Properties/SchemaTable.php
nunomaduro: Maybe just add a note on the class phpdocs.

canvural: How should that look like?

nunomaduro:
```php
/**
 * This code was inspired by ..
 */
class Foo {}
```

src/Properties/SchemaAggregator.php
szepeviktor: @nunomaduro There is such a thing as PHPDoc file comment.

```php
/**
 * @package Larastan
 * @license GPL
 */
```

and so on https://docs.phpdoc.org/latest/references/phpdoc/index.html

nunomaduro: Nah, I rather keep it simple with a link to the copy/pasted code.

canvural: Updated.

src/Properties/SchemaColumn.php
canvural: Updated.

src/Properties/SchemaTable.php
canvural: Updated.

troelsselch:
@canvural The PR I did for `AppExtension` (https://github.com/nunomaduro/larastan/pull/431) includes a copyright notice. Perhaps that should also be removed? https://github.com/nunomaduro/larastan/blob/master/src/ReturnTypes/Helpers/AppExtension.php

nunomaduro:
Yeah - should be removed.

PR #443, opened 2020-02-04
feat: return type support for get on relations

canvural:
With the generics added for collections, now we can infer the types of relationship collections better. For example, now we know `$user->accounts()->where('active', 1)->get();` returns a collection of accounts.

### DIFF
--- a/src/ReturnTypes/RelationCreateExtension.php
+++ b/src/ReturnTypes/RelationCreateExtension.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace NunoMaduro\Larastan\ReturnTypes;
 
 use Illuminate\Contracts\Container\BindingResolutionException;
+use Illuminate\Database\Eloquent\Collection;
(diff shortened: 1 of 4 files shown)


src/ReturnTypes/RelationCreateExtension.php
```diff
@@ -26,7 +28,11 @@ public function getClass(): string
 
     public function isMethodSupported(MethodReflection $methodReflection): bool
     {
-        return $methodReflection->getName() === 'create';
+        return $methodReflection->getName() === 'get' ||
+               $methodReflection->getName() === 'make' ||
+               $methodReflection->getName() === 'create' ||
+               $methodReflection->getName() === 'getEager' ||
+               $methodReflection->getName() === 'getResults';
```
szepeviktor: @canvural Could we use in_array() here?

canvural: Definitely!

PR #450, opened 2020-02-05
fix: model scopes returns builder if scope return type is void

### LINKED ISSUE #448: Result of void scope is used
- Larastan Version: 0.5.1
- `--level` used: 5

### Description

After upgrading I receive this error:

 | Line  | app/Models/Concerns/HasUuid.php (in context of class App\Models\Media)  |
 |------|------------------------------------------------------------------------|
 | 30     | Cannot call method exists() on void.                                    |
 | 30     | Result of static method App\Models\Media::byUuid() (void) is used.      |

The scope is the following:
```php
    public function scopeByUuid(Builder $query, $token): void
    {
        $query->whereIn($this->getUuidName(), Arr::wrap($token));
    }
```

and it's called like: `static::byUuid($uuid)->exists()`. `void` scopes will still return the already in use query builder instance. So we've chosen to flag all scopes that only add something to the query as `void` and use real return types only for scopes that aggregate some real values.

canvural:
Fixes #448 

If the return type of a local model scope is `void`, Larasan will return the builder.

### DIFF
--- a/src/Methods/BuilderHelper.php
+++ b/src/Methods/BuilderHelper.php
@@ -18,6 +18,7 @@ use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
 use PHPStan\Type\VerbosityLevel;
+use PHPStan\Type\VoidType;
 
 class BuilderHelper
 {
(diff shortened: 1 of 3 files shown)


Gummibeer:
Thanks for adding it! ❤️

PR #453, opened 2020-02-05
fix: fixes query builder methods within custom builders

### LINKED ISSUE #352: Invalid "self" type for custom Builder
- Larastan Version: 0.4.3
- `--level` used: max

### Description
Based on https://timacdonald.me/dedicated-eloquent-model-query-builders/ I have converted most of my code base away from Model scopes to custom query builders. This made using them in an IDE a lot asier.

Problem is: Simple functions like
```php
public function whereActive(): self
{
  return $this->whereNotNull('accepted_at');
}
```
work as expected. The problem starts as soon as another "scope" gets called inside this builder:

(just an example)

```php
class UserBuilder extends Builder {
  public function whereVerified(): self
  {
    return $this->whereNotNull('email_verified_at');
  }

  public function whereVip(): self
  {
    return $this->whereVerified()->where('points', '>', 50000);
  }
}
```

Since whereVerified() returns an instance of `self` this seems to overwrite the analysis and any consecute calls get the wrong scope. Even attaching PHPDocs like `@return \Illuminate\Database\Eloquent\Builder|self` don't work.

canvural:
Fixes #352 

Calling query builder methods inside the custom builder classes wasn't returning the generic builder. This PR fixes that.

### DIFF
--- a/src/Methods/EloquentBuilderForwardsCallsExtension.php
+++ b/src/Methods/EloquentBuilderForwardsCallsExtension.php
@@ -16,6 +16,7 @@ use PHPStan\Reflection\MethodsClassReflectionExtension;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Generic\GenericObjectType;
+use PHPStan\Type\Generic\TemplateMixedType;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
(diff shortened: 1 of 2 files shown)


tests/Features/ReturnTypes/CustomEloquentBuilderTest.php
```diff
@@ -96,4 +101,15 @@ public function type(string $type): CustomBuilder
     {
         return $this->where(['type' => $type]);
     }
+
+    /**
+     * @param string[] $categories
+     *
+     * @return CustomBuilder
+     * @phpstan-return CustomBuilder<TModelClass>
```
szepeviktor: I think other phpdoc parsing software understand `CustomBuilder<TModelClass>`

src/Methods/EloquentBuilderForwardsCallsExtension.php
```diff
@@ -88,18 +89,24 @@ public function getMethod(ClassReflection $classReflection, string $methodName):
...
         $modelType = $templateTypeMap->getType('TModelClass');
 
-        if ($this->getBuilderReflection()->hasNativeMethod($methodName) && ($modelType === null || ! $modelType instanceof ObjectType)) {
+        if ($this->getBuilderReflection()->hasNativeMethod($methodName) && (! $modelType instanceof ObjectType)) {
             $methodReflection = $this->getBuilderReflection()->getNativeMethod($methodName);
+            $builderClass = EloquentBuilder::class;
+
+            if ($modelType instanceof TemplateMixedType) {
+                /** @var string $builderClass */
+                $builderClass = $modelType->getScope()->getClassName();
```
szepeviktor: Could class name be a non-string?

tests/Features/ReturnTypes/CustomEloquentBuilderTest.php
canvural: PHPStorm was complaining about it for me.

src/Methods/EloquentBuilderForwardsCallsExtension.php
canvural: Yes. Technically it can be `null`. But in reality, it'll never be. So this comment is just to tell PHPStan that it's a string.

tests/Features/ReturnTypes/CustomEloquentBuilderTest.php
szepeviktor: I see. Generics! https://plugins.jetbrains.com/plugin/12754-php-generics

src/Methods/EloquentBuilderForwardsCallsExtension.php
szepeviktor: Thanks.

tests/Features/ReturnTypes/CustomEloquentBuilderTest.php
canvural: Oh, cool. But I don't want to install a plugin for that. PHPStorm will eventually support it.

PR #454, opened 2020-02-05
Return correct collection instance when calling all

### LINKED ISSUE #452: Collection doesn't return correct model instance
- Larastan Version: 0.5.1
- `--level` used: max

### Description
When getting the first value in a collection the return type is invalid. You get the error:
`testAllFirst() should return App\\User but returns Illuminate\\Database\\Eloquent\\Model|null.`

```php
/**
* @return User
*/
public function testAllFirst(): User
{
    return User::all()->first();
}
```

BertvanHoekelen:
Fixes #452 

Calling `User::all();` returned a `Eloquent/Collection`. This PR makes sure a `Support/Collection` is returned so that `User::all()->first()` returns a `User|null`.

### DIFF
--- a/src/ReturnTypes/ModelExtension.php
+++ b/src/ReturnTypes/ModelExtension.php
@@ -110,8 +110,7 @@ final class ModelExtension implements DynamicStaticMethodReturnTypeExtension, Br
             $returnType = new GenericObjectType(EloquentBuilder::class, [new ObjectType($scope->resolveName($methodCall->class))]);
         }
 
-        if ((count(in_array(Collection::class, $returnType->getReferencedClasses())) > 0)
-            && $methodCall->name->name === 'all'
+        if (in_array(Collection::class, $returnType->getReferencedClasses()) && $methodCall->name->name === 'all'
         ) {
             $returnType = new IntersectionType([new ObjectType(\Illuminate\Support\Collection::class), $returnType->getTypes()[1]]);
         }
--- a/src/ReturnTypes/ModelExtension.php
+++ b/src/ReturnTypes/ModelExtension.php
@@ -18,6 +18,7 @@ use PHPStan\Reflection\FunctionVariantWithPhpDocs;
 use PHPStan\Reflection\MethodReflection;
 use PHPStan\Type\DynamicStaticMethodReturnTypeExtension;
 use PHPStan\Type\Generic\GenericObjectType;
+use PHPStan\Type\IntersectionType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
 use ReflectionClass;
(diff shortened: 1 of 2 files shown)


szepeviktor:
Thank you for you PR!
https://www.php.net/manual/en/function.in-array.php
in_array return a boolean. Could you please correct the code?

BertvanHoekelen:
@szepeviktor Done!

szepeviktor:
Thanks.
Could you add a 3rd parameter `true` to get in_array work in a strict mode?

src/ReturnTypes/ModelExtension.php
```diff
@@ -109,6 +110,10 @@ public function getTypeFromStaticMethodCall(
             $returnType = new GenericObjectType(EloquentBuilder::class, [new ObjectType($scope->resolveName($methodCall->class))]);
         }
 
+        if (in_array(Collection::class, $returnType->getReferencedClasses()) && $methodCall->name->name === 'all') {
+            $returnType = new IntersectionType([new ObjectType(\Illuminate\Support\Collection::class), $returnType->getTypes()[1]]);
```
canvural: It should be the generic type here. `new GenericType(Collection::class, [new ObjectType()])`

tests/Features/Methods/ModelExtension.php
```diff
@@ -13,13 +13,18 @@
 class ModelExtension
 {
     /**
-     * @return \Illuminate\Database\Eloquent\Collection<User>
+     * @return \Illuminate\Support\Collection<User>
```
canvural: It was correct before. `all` returns `\Illuminate\Database\Eloquent\Collection` not ` \Illuminate\Support\Collection`

src/ReturnTypes/ModelExtension.php
```diff
@@ -109,6 +110,10 @@ public function getTypeFromStaticMethodCall(
             $returnType = new GenericObjectType(EloquentBuilder::class, [new ObjectType($scope->resolveName($methodCall->class))]);
         }
 
+        if (in_array(Collection::class, $returnType->getReferencedClasses()) && $methodCall->name->name === 'all') {
```
canvural: We should use `$methodReflection->getName()` instead of `$methodCall->name->name`

Also, can you check if we need to do this to methods other than `all` ?

BertvanHoekelen: I think `all` is the only one that should be fixed in the `ModelExtension.php`. However when adding the test below it gives the same error. So maybe I'm working in the wrong file?

```php
public static function findFirstFooBarThread(): Thread
{
    return self::query()->where('foo', 'bar')->get()->first();
}
```

canvural: That `get` call is on the builder not model. So let's make another issue/PR for that one.

BertvanHoekelen: Will do!

PR #471, opened 2020-02-20
fix: dont treat whereHas as dynamic where

### LINKED ISSUE #469: wrong return type when using model with()
- Larastan Version: 0.5.2
- `--level` used: 5

### Description
When using `::with()` in model statement phpstan throws error of wrong return type while using it later `->with()` instead works fine.
Error:
`Method App\Services\ProcessRecords\Transactions\ProcessTravelTransactionRecord::getOwnershipWithCardAndUser() should return App\Ownership but returns  
         Illuminate\Database\Eloquent\Builder<mixed>|Illuminate\Database\Eloquent\Model.  `

doesn't work:
```php
private function getOwnershipWithCardAndUser(array $ownershipsIds): Ownership
{
    return Ownership::with('card.users')->whereHas('card')->whereIn('number', $ownershipsIds)->firstOrFail();
}
```
does work:
```php
private function getOwnershipWithCardAndUser(array $ownershipsIds): Ownership
{
    return Ownership::whereHas('card')->with('card.users')->whereIn('number', $ownershipsIds)->firstOrFail();
}
```

mr-feek:
fixes #469


src/Methods/Pipes/BuilderDynamicWheres.php
```diff
@@ -1,41 +0,0 @@
...
-use Illuminate\Database\Query\Builder;
-use NunoMaduro\Larastan\Contracts\Methods\PassableContract;
-use NunoMaduro\Larastan\Contracts\Methods\Pipes\PipeContract;
-use NunoMaduro\Larastan\Methods\BuilderHelper;
-use PHPStan\Type\ObjectType;
-
-/**
- * @internal
- */
-final class BuilderDynamicWheres implements PipeContract
```
mr-feek: I deleted this because I don't see it referenced anywhere anymore

src/ReturnTypes/EloquentBuilderExtension.php
```diff
@@ -27,6 +27,13 @@ final class EloquentBuilderExtension implements DynamicMethodReturnTypeExtension
 {
     use Concerns\HasBroker;
 
+    private const NON_DYNAMIC_WHERES = [
```
mr-feek: note: I may be able to figure this out via `hasNativeMethod`,  I'll look into it...

src/ReturnTypes/EloquentBuilderExtension.php
```diff
@@ -35,7 +42,7 @@ public function getClass(): string
     public function isMethodSupported(MethodReflection $methodReflection): bool
     {
         // Don't handle dynamic wheres
-        if (Str::startsWith($methodReflection->getName(), 'where')) {
+        if (Str::startsWith($methodReflection->getName(), 'where') && ! in_array($methodReflection->getName(), self::NON_DYNAMIC_WHERES)) {
```
szepeviktor: Please ass true as 3rd parameter.

mr-feek: good call! done.

canvural:
Hi,

I don't think this is how it should be fixed. I don't see why this is necessary. `whereHas` should have the correct return type from `Methods/EloquentBuilderForwardsCallsExtension` anyway.

I'll do some tests later and let you know.

mr-feek:
Sure, I wrote a regression test so feel free to refactor. I’m still trying to fully understand how this repo is laid out. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

mr-feek:
From my testing, `EloquentBuilderForwardsCallsExtension::hasMethod` isn't invoked when `whereHas` is called on the model.

canvural:
Yeah that's right. It would be called if you call it on a query builder. Like `User::with() - >whereHas() `

I still didn't have a chance to take a look. Next week I'll check.

mr-feek:
@canvural let me know if you would like to point me in the right direction,  I'm happy to refactor

mr-feek:
From my understanding `EloquentBuilderForwardsCallsExtension` shouldn't be relevant because `whereHas` is a method in `QueriesRelationships`, which is a trait on the eloquent builder, not the underlying database query builder. Let me know if I'm misunderstanding you.

src/ReturnTypes/EloquentBuilderExtension.php
```diff
@@ -27,6 +27,13 @@ final class EloquentBuilderExtension implements DynamicMethodReturnTypeExtension
 {
     use Concerns\HasBroker;
 
+    private const NON_DYNAMIC_WHERES = [
+        'whereHas',
+        'whereDoesntHave',
+        'whereHasMorph',
+        'whereDoesntHaveMorph',
+    ];
+
```
canvural: We can remove this.

src/ReturnTypes/EloquentBuilderExtension.php
```diff
@@ -35,7 +42,7 @@ public function getClass(): string
     public function isMethodSupported(MethodReflection $methodReflection): bool
     {
         // Don't handle dynamic wheres
-        if (Str::startsWith($methodReflection->getName(), 'where')) {
+        if (Str::startsWith($methodReflection->getName(), 'where') && ! in_array($methodReflection->getName(), self::NON_DYNAMIC_WHERES, true)) {
             return false;
         }
```
canvural:
```suggestion
        $builderReflection = $this->getBroker()->getClass(EloquentBuilder::class);

        // Don't handle dynamic wheres
        if (Str::startsWith($methodReflection->getName(), 'where') &&
            ! $builderReflection->hasNativeMethod($methodReflection->getName())
        ) {
            return false;
        }
```

Instead of checking if the method name is in the array, we can check if builder has that method. This is more future-proof.

mr-feek:
@canvural refactored, good point! I originally did the hardcoding because I saw that done in other extensions.

canvural:
Thank you!

PR #477, opened 2020-02-25
refactor: consolidate duplicated constants

mr-feek:
Looks like these were copy pasted. Let's consolidate and just reference constants from one place.


canvural:
LGTM if you fix the style error.

mr-feek:
thanks @canvural , all done

Issue #475, opened 2020-02-20
[False Positive] Dynamic Where's Do Not Require A Parameter Of Type Array


mr-feek:
- Larastan Version: 0.5.2
- `--level` used: 5

### Description
Dynamic where's that specify the column in the method name should not require an array to be passed. Instead, it should be the `type` of the inferred property

### Laravel code where the issue was found

```php
return Model::whereCurrencyCode(string $currencyCode);
```

### Expected
No error.

### Actual
```
Parameter #1 $parameters of method Illuminate\Database\Query\Builder::whereCurrencyCode() expects array, string given.
```

### Workaround
In the meantime, this can be silenced via the folflowing in ignoreErrors
`- '#Parameter \#1 \$parameters of method Illuminate\\Database\\Query\\Builder\:\:where.*\(\) expects array, .* given.#'`

### Notes
dynamic where's are invoked from `Builder::__call`, which automatically casts the args to an array at runtime

mr-feek:
note: dynamic where's are invoked from `Builder::__call`, which automatically casts the args to an array at runtime

mr-feek:
I believe the bug is on this line: https://github.com/nunomaduro/larastan/blob/master/src/Methods/BuilderHelper.php#L54

When this method is invoked from an eloquent model, the parameter type should be string, rather than array

mr-feek:
@canvural can you suggest an implementation please?

I think deep down, `getVariants` on the  `EloquentBuilderMethodReflection` needs to be updated to also return a new `FunctionVariant` accepting  a string type  parameter 

🤔 💭

PR #482, opened 2020-02-28
fix: dynamic wheres can have a mixed parameter

mr-feek:
fixes https://github.com/nunomaduro/larastan/issues/475

I'm not sure if I've implemented this properly, but the test passes.

Basically the parameter should be a mixed type, instead of just `array` when `dynamicWhere` is invoked via the  `ForwardsCalls` trait on the model

### DIFF
--- a/src/Methods/BuilderHelper.php
+++ b/src/Methods/BuilderHelper.php
@@ -14,7 +14,9 @@ use PHPStan\Reflection\FunctionVariantWithPhpDocs;
 use PHPStan\Reflection\MethodReflection;
 use PHPStan\Reflection\MissingMethodFromReflectionException;
 use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Reflection\Php\DummyParameter;
 use PHPStan\ShouldNotHappenException;
+use PHPStan\Type\MixedType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
 use PHPStan\Type\VerbosityLevel;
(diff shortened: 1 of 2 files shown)


src/Methods/BuilderHelper.php
```diff
@@ -53,10 +55,15 @@ public function dynamicWhere(
...
+        $originalParameter = $originalDynamicWhereVariant->getParameters()[1];
+
+        $actualParameter = new DummyParameter($originalParameter->getName(), new MixedType(), $originalParameter->isOptional(), $originalParameter->passedByReference(), $originalParameter->isVariadic(), $originalParameter->getDefaultValue());
+
         return new EloquentBuilderMethodReflection(
             $methodName,
             $classReflection,
-            [$originalDynamicWhereVariant->getParameters()[1]],
+            [$actualParameter],
             $returnObject
```
canvural:
```suggestion
            $returnObject,
            true
```

Laravel also allows to do something like `User::whereIdAndEmail(1, 'foo@example.com')->first()`

So, dynamic wheres can accept multiple arguments. If we can add this `true` here we can mark it as variadic and that should do the trick.

mr-feek: Cool, wasn't aware that you could do dynamicWhere's with multiple  columns, fixed!

mr-feek:
@canvural are you able to pass an object/array to a dynamic where? Not sure if we can get more specific than `mixed`

canvural:
mixed is okay. We just need to mark it as variadic by passing true as last argument.

mr-feek:
done 👍

canvural:
Can you also add a test case for multiple arguments?

mr-feek:
@canvural I believe I did, lemme know if you had something else in mind 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

canvural:
Oh, sorry. Now I see it. I guess Github showed me the old changes.

Now I also see the tests are at `Models/Relation` Oh, sorry. Now I see it. I guess Github showed my the old changes.

Now I also see the tests are at `Models/Relations` I don't think this is the right place. Maybe we can put it in the `Methods/Builder` cause its related to query builder and dynamic wheres.

mr-feek:
@canvural done

PR #492, opened 2020-03-17
Removes annotation extension

canvural:
This PR removes the annotation extension we used.

This was a hack for to solve some issue. But it was wrong. We should not change the PHPStan's behaviour.


nunomaduro:
> This was a hack for to solve some issue.

What is the difference from before? 👍

canvural:
@nunomaduro Before, with the `AnnotationExtension`, first it was reading the annotations to find the method or the return type. PHPStan also does this, but it has some order. And `AnnotationExtension` is the last one in PHPStans order. So my `hack` was to bring it to the first place. To solve some issue.

But this is bad. We should not change the PHPStan's behaviour. Also I fixed the original issue with a stub. So everything is also fine now.

nunomaduro:
I see, fell free to merge once tests are passing. Or if they are failing for other reason, you can merge.

canvural:
Tests failing because of Lumen. And it's an annotation error in the Lumen source code.

I'll do a PR to Lumen repo for the issue. Also, I can't merge, because one check is failing. Can you merge this?

PR #493, opened 2020-03-17
Fixes handler method return type for Lumen

canvural:
The PR I made got rejected: https://github.com/laravel/lumen/pull/146 Even though the same docblock is changed in `laravel/laravel` [repo](https://github.com/laravel/laravel/blob/master/app/Exceptions/Handler.php#L47) with [this](https://github.com/laravel/laravel/pull/5187) PR  :man_shrugging: 

So I'm adding another change to the test script, just like the one [we have](https://github.com/nunomaduro/larastan/blob/master/tests/laravel-test.sh#L14) before.


szepeviktor:
Taylor O. may not use static analysis.

PR #496, opened 2020-03-18
ModelPropertyExt.::getDateClass returns Carbon

### LINKED ISSUE #495: Model's dates are forced to use Illuminate\Support\Carbon
- **Larastan Version: dev-master**
- `--level` used: 6

### Description
This ticket follows conversation opened in #494 <br>
After updating to dev-master I got errors related to Carbon in models.

When assigning Carob::now() into models attribute, this pops out `Property SunApp\Models\Term::$start (Illuminate\Support\Carbon|string|null) does not accept Carbon\Carbon.` like this in code below.

```php
use Carbon\Carbon;

$term->start = Carbon::now();
```

Issues thrown: `Cannot call method diffInDays() on string.` , `Cannot call method format() on string.` and `Cannot access property $year on string.` All of them are on lines in getTermRangeAttribute method.

```php
use Carbon\Carbon;

/**
 * @property Carbon $start
 * @property Carbon $end
 */
class Term extends BaseModel
{
    protected $casts = [
        'start'             => 'date',
        'end'               => 'date',
    ];

    /**
     * Eloquent method for getting date "name", ie start and end date formatted
     */
    public function getTermRangeAttribute(): string
    {
        if ($this->start->diffInDays($this->end) === 0) {
            $termRange = $this->start->format('d.m.Y');
        } else {
            $start = $this->start->format('d.m.');
            if ($this->start->year != $this->end->year) {
                $start .= $this->start->year;
            }
            $end = $this->end->format('d.m.Y');
            $termRange = "{$start} - {$end}";
        }
        if (!empty($this->label)) {
            $termRange .= " - {$this->label}";
        }

        return $termRange;
    }
}
```

szepeviktor:
Closes #495

### DIFF
--- a/src/Properties/ModelPropertyExtension.php
+++ b/src/Properties/ModelPropertyExtension.php
@@ -134,7 +134,10 @@ final class ModelPropertyExtension implements PropertiesClassReflectionExtension
         }
     }
 
-    private function getDateClass(): string
+    /**
+     * @return string|\Carbon\Carbon
+     */
+    private function getDateClass()
     {
         return $this->dateClass ?? class_exists(\Illuminate\Support\Facades\Date::class)
             ? '\\'.get_class(\Illuminate\Support\Facades\Date::now())


szepeviktor:
```
 ------ ----------------------------------------------------------------------- 

  Line   Properties/ModelPropertyExtension.php                                  

 ------ ----------------------------------------------------------------------- 

  140    Method                                                                 

         NunoMaduro\Larastan\Properties\ModelPropertyExtension::getDateClass()  

         never returns Carbon\Carbon so it can be removed from the return       

         typehint.                                                              

 ------ ----------------------------------------------------------------------- 
```

canvural:
Hi,

This is not correct. It shouldn't be an annotation but it should be in the code actually.

Maybe something ike this: 
```php
if ($this->dateClass !=== null) {
	return $this->dateClass;
}

$dateClass =  class_exists(\Illuminate\Support\Facades\Date::class)
	? '\\'.get_class(\Illuminate\Support\Facades\Date::now())
	: '\Illuminate\Support\Carbon';

$dateClass .= '\Carbon\Carbon';

$this->dateClass = $dateClass;

return $this->dateClass;
```

So the idea is just to add it to the output. You can clean up my code sample and use it.

szepeviktor:
Okay.
I'm much like a janitor.
I let developers do it!

PR #498, opened 2020-03-19
Fatal error when migrations folder does not exist

deleugpn:
With the recent update I'm getting a fatal error when the `migrations` folder doesn't exist:

```
Internal error: RecursiveDirectoryIterator::__construct(/var/task/database/migrations): failed to open dir: No such file or directory
```

This is an attempt to patch that up.

### DIFF
--- a/src/Properties/ModelPropertyExtension.php
+++ b/src/Properties/ModelPropertyExtension.php
@@ -100,6 +100,10 @@ final class ModelPropertyExtension implements PropertiesClassReflectionExtension
 
     private function initializeTables(): void
     {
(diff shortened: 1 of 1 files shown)


canvural:
Thank you.

In the future we should make this configurable. Users should be able to specify the path in their `phpstan.neon` file, if they use different path.

deleugpn:
Thanks @canvural. When making this configurable, please keep in mind that perhaps `null` should be a valid option. For me it wasn't about a different folder, but rather about not having laravel migrations in the project at all.

Issue #491, opened 2020-03-17
Properties not inferred correctly from database migrations


DSpeichert:
- Larastan Version: 0.5.2
- `--level` used: 5

### Description

```
 ------ ------------------------------------------------------- 
  Line   DatabaseNotificationPolicy.php                         
 ------ ------------------------------------------------------- 
  34     Access to an undefined property App\Models\User::$id.  
  34     Access to an undefined property App\Models\User::$id.  
 ------ ------------------------------------------------------- 
```

### Laravel code where the issue was found

```php
    /**
     * Determine whether the user can view the model.
     *
     * @param  \App\Models\User                               $user
     * @param  \Illuminate\Notifications\DatabaseNotification $notification
     * @return mixed
     */
    public function view(User $user, DatabaseNotification $notification)
    {
        return $notification->notifiable instanceof User
            and $user->id == $notification->notifiable->id;
    }
```

It is my understanding that with #435 merged, the properties found in database migrations should be automatically detected, as the tests in that PR suggests.

Am I missing something here? Is there a way to trace what properties larastan "knows" for a given model?

szepeviktor:
Thank you for your report.
How does your User model look like?

szepeviktor:
It is best to use `===` for comparison :)

DSpeichert:
I have multiple migrations that modify the table later (application grew over time).
The first one that creates it looks like this:
```
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::create('users', function (Blueprint $table) {
            $table->increments('id');
            $table->timestamps();
            $table->softDeletes();
            $table->string('email');
            $table->string('email_cc')->nullable();
            $table->string('email_previous')->nullable();
            $table->boolean('email_confirmed')->default(0);
            $table->string('email_code')->nullable();
            $table->string('password');
            $table->rememberToken();
            $table->string('old_password')->nullable();
            $table->string('first_name');
            $table->string('last_name');
            $table->string('company')->nullable();
            $table->string('address1')->nullable();
            $table->string('address2')->nullable();
            $table->string('post_code', 10)->nullable();
            $table->string('city', 50)->nullable();
            $table->string('country_code', 2)->nullable();
            $table->string('phone', 14)->nullable();
            $table->string('phone_previous')->nullable();
            $table->boolean('phone_confirmed')->default(0);
            $table->string('phone_code')->nullable();
            $table->string('vat_id', 20)->nullable();
[...]

PR #499, opened 2020-03-19
Always infer return type of id property on models

canvural:
There are some cases when there is migration but not a model file. For example pivot tables. So Larastan will complain about the property accesses.

This PR, adds logic to always assume a property named `id` exists on the model. I think this is a safe assumption to make. And it can solve couple of errors.


szepeviktor:
Related #491

tectiv3:
Assuming that `id` exists on a model is reasonable, but that it is always integer - probably not.
UUID as a primary key is pretty common, I think.

zlodes:
How to disable this?
We are using UUID and `string` type in `@property` was ignored...

PR #513, opened 2020-03-29
Fix incorrect boolean type cast

leo108:
Boolean properties were cast to `boolean` here:
https://github.com/nunomaduro/larastan/blob/master/src/Properties/ModelPropertyExtension.php#L233
So we need to compare `$column->readableType` with `boolean` instead of `bool`

### DIFF
--- a/src/Properties/ModelPropertyExtension.php
+++ b/src/Properties/ModelPropertyExtension.php
@@ -195,15 +195,16 @@ final class ModelPropertyExtension implements PropertiesClassReflectionExtension
                 $readableType = $writableType = $column->readableType.($column->nullable ? '|null' : '');
                 break;
 
+            case 'boolean':
             case 'bool':
                 switch ((string) config('database.default')) {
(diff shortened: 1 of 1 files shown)


canvural:
Hi, thanks for the PR!

Can you explain why this is necessary? Cause [here](https://github.com/nunomaduro/larastan/blob/master/src/Properties/ModelPropertyExtension.php#L117) the type resolver would convert both `bool` and `boolean` to the correct type. So I don't think there will be an issue.

leo108:
@canvural In old logic, if `$column->readableType` is `boolean`, the `getReadableAndWritableTypes` method will return `[mixed, mixed]` instead of `[bool, bool]`.

The core of this PR is this line: https://github.com/nunomaduro/larastan/pull/513/files#diff-8c4331b9e7c6c4544f1895befdc48f1dR198

canvural:
In that case, you are right :+1: Thank you!

PR #518, opened 2020-04-02
feat: generic relations

canvural:
This PR mainly does 2 things:

- Before, to get the related model of a relation we were executing the users' code and then getting the related model with `getRelatedModel` method. While analyzing some public Laravel apps with Larastan I saw this caused some issues. So this PR changes the behavior and tries to find the related model statically.
- Relations are now also generic. This allows us to drop some extensions, and use stubs to provide the same functionality. Also, it'll make easier to fix any problem that can come up. We just need to edit the stubs.

The existing code is not affected by any changes. So this is not a breaking change.


canvural:
That's a good idea! Sometimes I not sure if a particular thing that I'm doing is the right way or there is an easier way.

Unfortunately I don't know anyone. We had some contributors from time to time. But not sure if they are still interested or not.

So yeah we can try to ask on Twitter!

PR #522, opened 2020-04-04
fix: dont assume primary key is id

mr-feek:
> Eloquent will also assume that each table has a primary key column named id. You may define a protected $primaryKey property to override this convention:

I have lots of models with custom ids

### DIFF
--- a/tests/Features/Properties/ModelPropertyExtension.php
+++ b/tests/Features/Properties/ModelPropertyExtension.php
@@ -9,6 +9,7 @@ use App\Group;
 use App\Role;
 use App\User;
 use Carbon\Carbon as BaseCarbon;
+use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Carbon;
 
 class ModelPropertyExtension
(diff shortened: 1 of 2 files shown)


canvural:
Thank you for the PR!

I don't quite understand why is this necessary. Don't you also put the primary key with the custom name in the migrations? Doesn't that work?

mr-feek:
@canvural That's assuming that laravel migrations are present.

I have an enterprise application with thousands of database tables, and we manage migrations with https://github.com/skeema/skeema

Laravel migration files shouldn't be a dependency for knowing the  primary key of the model.

mr-feek:
@canvural also I believe if a migration defined a custom id field , then `id` would always be allowed as a property and would not be caught as an error.

tests/Features/Properties/ModelPropertyExtension.php
```diff
@@ -95,4 +96,16 @@ public function testWriteIdPropertyWhenMigrationsCouldntBeRead(): bool
 
         return $group->save();
     }
+
+    /** @test */
+    public function it_knows_name_of_primary_key_column(): void
+    {
+        $model = new CustomPrimaryKeyColumn();
+        $primaryKey = $model->custom_primary_key_name;
```
canvural, line 104: Can you return this here and check the return type is correctly recognized as an integer?

PR #525, opened 2020-04-05
feature: relationship count properties

### LINKED ISSUE #466: Relationship count should be recognized as property
From the [docs](https://laravel.com/docs/6.x/eloquent-relationships#counting-related-models):

> If you want to count the number of results from a relationship without actually loading them you may use the withCount method, which will place a {relation}_count column on your resulting models.

We should add support for this. It should be simple check. If a property named `{relation}_count` is accessed on the model, and `{relation}` method exists in the model, then we can say this property also exists. And has an `integer` type.

mr-feek:
closes #466

### DIFF
--- a/src/Properties/ModelRelationsExtension.php
+++ b/src/Properties/ModelRelationsExtension.php
@@ -16,6 +16,7 @@ use PHPStan\Reflection\Dummy\DummyPropertyReflection;
 use PHPStan\Reflection\PropertiesClassReflectionExtension;
 use PHPStan\Reflection\PropertyReflection;
 use PHPStan\Type\Generic\GenericObjectType;
+use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\NeverType;
 use PHPStan\Type\NullType;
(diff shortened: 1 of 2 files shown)


src/Properties/ModelRelationsExtension.php
```diff
@@ -44,11 +45,19 @@ public function hasProperty(ClassReflection $classReflection, string $propertyNa
...
+            $propertyName = Str::beforeLast($propertyName, '_count');
+        }
+
         return $classReflection->hasNativeMethod($propertyName);
     }
 
     public function getProperty(ClassReflection $classReflection, string $propertyName): PropertyReflection
     {
+        if (Str::endsWith($propertyName, '_count')) {
+            return new ModelProperty($classReflection, IntegerRangeType::fromInterval(0, null), IntegerRangeType::fromInterval(0, null), false);
```
canvural, line 58: This property should not be writable I guess. So you can give `NeverType` as writable type.

tests/Features/Models/Relations.php
```diff
@@ -141,6 +141,14 @@ public function testRelationWithTrait(Account $account): BelongsTo
     {
         return $account->ownerRelation();
     }
+
+    /** @test */
+    public function it_has_relationship_count_properties(): void
+    {
+        User::withCount('accounts')->get()->each(function (User $user) {
+            echo $user->accounts_count;
```
canvural, line 149: Instead of echoing it here, can you return it and check the return type is an integer?

PR #528, opened 2020-04-10
Allow composer v2

GrahamCampbell:
[Composer v2](https://github.com/composer/composer/blob/master/UPGRADE-2.0.md) will be released this month.

### DIFF
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace NunoMaduro\Larastan;
 
+use Composer\Autoload\ClassLoader;
 use Composer\Autoload\ClassMapGenerator;
 use Illuminate\Contracts\Foundation\Application;
 use function in_array;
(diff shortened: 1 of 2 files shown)


src/ApplicationResolver.php
```diff
@@ -98,7 +99,7 @@ class_exists($class, true);
      */
     private static function getProjectSearchDirs(string $namespace): array
     {
-        $reflection = new \ReflectionClass(\Composer\Autoload\ClassLoader::class);
+        $reflection = new \ReflectionClass(ClassLoader::class);
```
szepeviktor, line 102:
```suggestion
        $reflection = new ReflectionClass(ClassLoader::class);
```

Could you add `use ReflectionClass;` ?

GrahamCampbell: Why? Not related to this PR.

szepeviktor: If you're at it please ...

nunomaduro:
@GrahamCampbell Would be nice to address proper testing on this. If doesn't take too much time of yours, can you adjust https://github.com/nunomaduro/larastan/blob/master/.travis.yml also to make sure we test composer 2 with all versions of Laravel in PHP?

GrahamCampbell:
Current Laravel doesn't support composer 2 yet, but this will be fixed in Tuesday's releases.

GrahamCampbell:
For 6 and upwards.

PR #530, opened 2020-04-10
Test with Composer 2.0

szepeviktor:
Fullfills https://github.com/nunomaduro/larastan/pull/528#issuecomment-611978645

Should fail as per https://github.com/nunomaduro/larastan/pull/528#issuecomment-611981562


nunomaduro:
Perfect! But I think with that pull request we are always teting against composer `^2.0` once is out, right?

szepeviktor:
Yes, big matrix!

kép

szepeviktor:
After thinking a bit... after Composer 2.0 will be released **and** supported we should change `COMPOSER_VERSION` to `^1.0`

nunomaduro:
I would say we should have something like:
```
    - LARAVEL_VERSION=^6.0 COMPOSER_VERSION=^1.0
     - LARAVEL_VERSION=^6.0 COMPOSER_VERSION=^2.0
     - LARAVEL_VERSION=^7.0 COMPOSER_VERSION=^1.0
     - LARAVEL_VERSION=^7.0 COMPOSER_VERSION=^2.0
```

szepeviktor:
Done.

GrahamCampbell:
Why not just only test on composer 2? Composer 1 will not even work soon, once packagist drops the endpoints it was using,

nunomaduro:
@GrahamCampbell Can you give me the source of that `Composer 1 will not even work soon`?

GrahamCampbell:
https://github.com/composer/composer/issues/8785#issuecomment-616209238

canvural:
Thanks @szepeviktor I'll try to improve the things mentioned above in the other PR.

Issue #531, opened 2020-04-14
Class self was not found while trying to analyse it - autoloading is probably not configured properly


mfn:
Using `master` I encountered the following bug ( similar but not quite as https://github.com/nunomaduro/larastan/issues/516 ):
```
 ------ --------------------------------------------------------------------------------------------------------
  Line   Some/Klass.php
 ------ --------------------------------------------------------------------------------------------------------
         Class self was not found while trying to analyse it - autoloading is probably not configured properly.
  56     Class self was not found while trying to analyse it - autoloading is probably not configured properly.
  56     Class self was not found while trying to analyse it - autoloading is probably not configured properly.
  56     Class self was not found while trying to analyse it - autoloading is probably not configured properly.
 ------ --------------------------------------------------------------------------------------------------------
```
Not kidding, the first line does truly not report a (line) number. Since it doesn't "crash", adding `--debug` didn't reveal anything.

The reported line looks like this:
```php
            Model
                ::whereIn(
                    'some_column',
                    $anotherModel
                    ->children() # <- a `HasMany` relation defined as `return $this->hasMany(self::class, 'parent_id');`
                    ->pluck('id')
                );
```
In my case its triggered because I call `pluck` directly on the relation; if I insert `->get()` the reported error is gone:
```php
            Model
                ::whereIn(
                    'some_column',
                    $anotherModel
                    ->children()
                    ->get() # <- line added
                    ->pluck('id')
                );
```
I think the repeated reported in due to the span across so many lines, it changes when I join the lines etc.

_Originally posted by @mfn in https://github.com/nunomaduro/larastan/issues/516#issuecomment-613123559_

szepeviktor:
# 😲

mfn:
Btw I _think_ this is the error which seems to prevent me a baseline, never seen this 😄 but I assume so because of the missing line number:
```
> phpstan analyse --configuration=phpstan.neon app --generate-baseline
 978/978 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%


 [WARNING] Baseline generated with 115 errors.
           Some errors could not be put into baseline. Re-run PHPStan and fix
           them.
```
:-)

mfn:
Ah, can't be manually ignored either:
> `Error message "Class self was not found while trying to analyse it - autoloading is probably not configured properly." cannot be ignored, use excludes_analyse instead.`

[...]

PR #533, opened 2020-04-16
fix: better return type inference when reading the body of relation methods

canvural:
Related to #531 

This PR provides better support for return types of relations methods when reading their body.

cc @ondrejmirtes


src/Types/RelationParserHelper.php
```diff
@@ -47,14 +69,27 @@ public function findRelatedModelInRelationMethod(string $fileName, string $metho
...
+
+        if ($argType instanceof ConstantStringType) {
+            return $argType->getValue();
         }
 
-        if ($argValue instanceof Node\Scalar\String_) {
-            return $argValue->value;
+        if ($argType instanceof GenericClassStringType) {
+            /** @var ObjectType $modelType */
+            $modelType = $argType->getGenericType();
```
ondrejmirtes, line 90: This doesn't always have to be `ObjectType`. `@template T` without bound is TemplateMixedType and it doesn't have `getClassName()` on it. So you should add another `if`.

src/Types/RelationParserHelper.php
```diff
@@ -7,24 +7,46 @@
...
-    public function findRelatedModelInRelationMethod(string $fileName, string $methodName): ?string
-    {
+    public function findRelatedModelInRelationMethod(
+        MethodReflection $methodReflection
+    ): ?string {
+        $fileName = $methodReflection
+            ->getDeclaringClass()
+            ->getNativeReflection()
+            ->getMethod($methodReflection->getName())
+            ->getFileName();
```
canvural, line 40: If the method is defined in a trait `$methodReflection->getDeclaringClass()->getFileName()` wouldn't return the trait filename. So this is the only way I found to get the trait filename if it's defined in a trait. @ondrejmirtes Is there a better way?

ondrejmirtes: I don't think this will always return the trait 🤔

There's no easy way in PHP reflection to find out if a method comes from a trait or not. PHPStan itself does some complex logic to find out about it (https://github.com/phpstan/phpstan-src/blob/master/src/Reflection/Php/PhpClassReflectionExtension.php#L640-L674) to get the right PHPDocs but even this sometimes fails.

There's a nice article about it: http://mouf-php.com/blog/php_reflection_api_traits

What I'd recommend to you is to have guards around this code - if you don't find the method definition in the filename that reflection reports (so it means it's somewhere else, like in a trait), just return `null` or `mixed` or whatever you need to return to not analyse that piece of code..

canvural: Oh, ok. Good to know :+1: 

Yeah, there are already some guards. If `getFileName` returns false, or if we can't find the method we are looking for in the file, it'll not analyze it further.

canvural:
Is there a point of caching the result of analyzing the method body, in the extension? Or the performance gain would be negligible?

ondrejmirtes:
@canvural I don't cache it in phpstan-doctrine and it's fine. You can do a simple in-memory cache by an array in private property if you're concerned. That way, the AST will not be traversed multiple times looking for the same method during the same PHPStan run.

Issue #436, opened 2020-02-01
Support for custom Eloquent Collection


canvural:
Users can override the Eloquent's `newCollection` method to return a [custom Eloquent collection](https://laravel.com/docs/6.x/eloquent-collections#custom-collections).

We should add support for this. 

It'll affect the collection returned by methods like `find`, `all` etc. As well as collections returned by relationships.

Gummibeer:
Because it's flagged as `good first issue` - could you point out where the custom collection stuff should hook in/get placed?

canvural:
Hi,

Thanks for taking an interest.

For starters, everywhere we use `Collection::class` in the code, needs to be changed with the custom collection class name. Figuring out if model uses custom collection and getting the custom collection class name can be extracted to a helper class/method. Like in the `BuilderHelper` with `determineBuilderType`

Gummibeer:
I think that we even doesn't have to check if the method is overridden because by default the `newCollection` method returns the default collection. So it could be simpler to just do `get_class($model->newCollection())` all the time!?

I'm not sure when I will find time, but if no one has already started with it I could try it. If I start I will write here - would be great if another dev starting it would also write so we won't do it twice. :)

canvural:
Yes, that can also work. :+1:

cweiske:
When calling a method on that custom collection class that has been returned by `newCollection()`, I still got an error:

> Call to an undefined method Illuminate\Database\Eloquent\Collection<App\Account>::collectionMethod()

The solution was to define a "property-read" for the HasMany-property:
```
@property-read \App\AccountCollection $accounts
```

This was with phpstan 0.12.57, larastan v0.6.10

----

Full code:

AccountCollection.php:
```
<?php
namespace App;
use Illuminate\Database\Eloquent\Collection;
/**
 * @template TModel
 * @extends Collection<TModel>
 */
class AccountCollection extends Collection
{
    function collectionMethod(): void
    {
    }
}
```

Account.php:
```
<?php
namespace App;
class Account extends \Illuminate\Database\Eloquent\Model
{
    /**
     * @param array<int, Account> $models
     *
     * @return AccountCollection<Account>
     */
    public function newCollection(array $models = []): AccountCollection
    {
        return new AccountCollection($models);
    }
}
```

User.php:
```
<?php
namespace App;
use Illuminate\Database\Eloquent\Relations\HasMany;
/**
 * @property-read \App\AccountCollection $accounts <-- This is still required
 */
[...]

PR #537, opened 2020-04-18
Support custom eloquent collections

timacdonald:
Fixes: https://github.com/nunomaduro/larastan/issues/436

This PR brings support for [custom eloquent collections](https://laravel.com/docs/7.x/eloquent-collections#custom-collections).

## BuilderHelper::determineCollectionType

Now that this PR is done, I'm not entirely sure `determineCollectionType` should be on the builder helper. It was suggested this could be its home, however maybe it should be extracted to a Collection helper as it isn't really just about the builder? Maybe it doesn't matter.

[See method](https://github.com/nunomaduro/larastan/pull/537/files#diff-7feea80f9bc562bad5bf39fba90c9582R198-R213)

## Fixed find...() null return type

If any `find...()` methods receive an iterable, they will never return `null`. They always proxy to `findMany` which in all cases returns a collection.

See [here](https://github.com/nunomaduro/larastan/pull/537/files#diff-dc0c9bf2638b255d8cf20a1067ed6782L85-L87) and [here](https://github.com/nunomaduro/larastan/pull/537/files#diff-ffcea4aec97c2745f3004fe1464993bdL76-L78)

## Fixed fromQuery return type

`Builder::fromQuery` was included in some existing type checks, but as that method always returns a collection I removed it.

[See removal](https://github.com/nunomaduro/larastan/pull/537/files#diff-7feea80f9bc562bad5bf39fba90c9582L31)

## Doesn't handle factory helper

Due to the dynamic nature of the factory helper method and the builder, I was unable to get the correct return type for factory calls. Not sure if this needs to be included in this PR or if we can follow up with another PR to add this functionality. Here is some examples of what we'd have to handle...

```php
// should return custom class collection...
factory(Account::class, 1)->create();
factory(Account::class, 'name', 1)->create();
factory(Account::class)->times(1)->create();
factory(Account::class)->times(null)->times(1)->create();


// should return a single instance...
factory(Account::class)->create();
factory(Account::class, 'name')->create();
factory(Account::class)->times(null)->create();
factory(Account::class)->times(1)->times(null)->create();
```

[See factory helper](https://github.com/laravel/framework/blob/4e9f584689a185b2d10a543bf20fbca5e4e2989c/src/Illuminate/Foundation/helpers.php#L481-L497)

## Doesn't handle MorphTo::getResultsByType

This method accepts a string that could be either a class or a morph map key. Without booting the framework and retrieving the class from the Morph Map there is not way (that I'm aware) to determine the class, and therefore the collection type.

It is a protected method, and I'd guess this method isn't really used outside the framework anyway, however I've got no data to back that up.

```php

Relation::morphMap([
    'videos' => 'App\Video',
]);

$morphTo->getResultsByType('videos');
```

## Doesn't handle instance proxying by design

All methods that are accessible via static calls on the model are also available via instance calls. Looks like Larastan doesn't currently support that for other methods, so I didn't add support for this feature, but I could if needed.

```php
$instance->find($id);
$instance->hydrate($records);
```

[See method](https://github.com/laravel/framework/blob/efdd2781b9cf7924bef73c047a25caba907903de/src/Illuminate/Database/Eloquent/Relations/MorphTo.php#L107-L131)

### DIFF
--- a/src/Methods/BuilderHelper.php
+++ b/src/Methods/BuilderHelper.php
@@ -28,7 +28,7 @@ class BuilderHelper
     public const MODEL_RETRIEVAL_METHODS = ['first', 'find', 'findMany', 'findOrFail', 'firstOrFail'];
 
     /** @var string[] */
-    public const MODEL_CREATION_METHODS = ['make', 'create', 'forceCreate', 'findOrNew', 'firstOrNew', 'updateOrCreate', 'fromQuery', 'firstOrCreate'];
+    public const MODEL_CREATION_METHODS = ['make', 'create', 'forceCreate', 'findOrNew', 'firstOrNew', 'updateOrCreate', 'firstOrCreate'];
 
     /**
      * @var Broker
@@ -192,5 +192,18 @@ class BuilderHelper
         }
 
         return $this->dynamicWhere($methodName, $customReturnType);
+    }
+
+    public function determineCollectionClassName(string $modelClassName): string
+    {
+        $newCollectionMethod = $this->broker->getClass($modelClassName)->getNativeMethod('newCollection');
+
+        $returnType = ParametersAcceptorSelector::selectSingle($newCollectionMethod->getVariants())->getReturnType();
+
+        if ($returnType instanceof ObjectType) {
+            return $returnType->getClassName();
+        }
+
+        return $returnType->describe(VerbosityLevel::value());
     }
 }
--- a/src/ReturnTypes/EloquentBuilderExtension.php
+++ b/src/ReturnTypes/EloquentBuilderExtension.php
@@ -90,9 +90,12 @@ final class EloquentBuilderExtension implements DynamicMethodReturnTypeExtension
             );
         }
 
-        // 'get' method return type
-        if ($methodReflection->getName() === 'get') {
-            return new GenericObjectType(Collection::class, [$modelType]);
+        if ($modelType instanceof ObjectType && in_array(Collection::class, $returnType->getReferencedClasses(), true)) {
+            $builderHelper = new BuilderHelper($this->getBroker());
+
+            $collectionClassName = $builderHelper->determineCollectionClassName($modelType->getClassName());
+
+            return new GenericObjectType($collectionClassName, [$modelType]);
         }
 
         return $returnType;
--- a/src/ReturnTypes/ModelExtension.php
+++ b/src/ReturnTypes/ModelExtension.php
@@ -63,6 +63,10 @@ final class ModelExtension implements DynamicStaticMethodReturnTypeExtension, Br
             return false;
         }
 
+        if (in_array($name, ['get', 'hydrate', 'fromQuery'], true)) {
+            return true;
+        }
+
         return $methodReflection->getDeclaringClass()->hasNativeMethod($name);
     }
 
@@ -115,8 +119,12 @@ final class ModelExtension implements DynamicStaticMethodReturnTypeExtension, Br
             );
         }
 
-        if ($methodReflection->getName() === 'all' && in_array(Collection::class, $returnType->getReferencedClasses(), true) && $methodCall->class instanceof \PhpParser\Node\Name) {
-            $returnType = new GenericObjectType(Collection::class, [new ObjectType($scope->resolveName($methodCall->class))]);
+        if ($methodCall->class instanceof \PhpParser\Node\Name && in_array(Collection::class, $returnType->getReferencedClasses(), true)) {
+            $builderHelper = new BuilderHelper($this->getBroker());
+
+            $collectionClassName = $builderHelper->determineCollectionClassName($methodCall->class->toString());
+
+            return new GenericObjectType($collectionClassName, [new ObjectType($methodCall->class->toString())]);
         }
 
         return $returnType;
--- a/tests/Application/app/Account.php
+++ b/tests/Application/app/Account.php
@@ -16,4 +16,19 @@ class Account extends Model
     {
         return $query->where('active', 1);
     }
+
+    public function transactions(): HasMany
+    {
+        return $this->hasMany(Transaction::class);
+    }
+
+    /**
+     * @param array<int, Account> $models
+     *
+     * @return AccountCollection<Account>
+     */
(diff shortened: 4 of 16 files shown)


szepeviktor:
Please un-draft thei PR when you are ready!

timacdonald:
Will do!

src/Methods/BuilderHelper.php
```diff
@@ -30,6 +30,9 @@ class BuilderHelper
     /** @var string[] */
     public const MODEL_CREATION_METHODS = ['make', 'create', 'forceCreate', 'findOrNew', 'firstOrNew', 'updateOrCreate', 'fromQuery', 'firstOrCreate'];
 
+    /** @var string[] */
+    public const COLLECTION_RETRIEVAL_METHODS = ['hydrate', 'findMany', 'get'];
```
canvural, line 34: I think this is not needed. We can solve it in a more general way. In other comments, I will try to describe how might we do it.

src/ReturnTypes/EloquentBuilderExtension.php
```diff
@@ -98,9 +98,13 @@ public function getTypeFromMethodCall(
...
-        // 'get' method return type
-        if ($methodReflection->getName() === 'get') {
-            return new GenericObjectType(Collection::class, [$modelType]);
+        // collection return methods...
+        if (in_array($methodReflection->getName(), BuilderHelper::COLLECTION_RETRIEVAL_METHODS, true)) {
+            $class = $modelType->getReferencedClasses()[0];
+
+            $collection = (new $class)->newCollection([]);
+
+            return new GenericObjectType(get_class($collection), [$modelType]);
```
canvural, line 107: We can do this in a more general way. Similar to how custom Eloquent builder is implemented. We can extract the logic of getting the Collection class to a method. Like [here.](https://github.com/nunomaduro/larastan/blob/master/src/Methods/BuilderHelper.php#L133-L148) It can be also in the `BuilderHelper` class I guess.

Then we can have another `if` here and that would check if `Eloquent\Collection` is referenced in return type. With something like `in_array(Collection::class, $returnType->getReferencedClasses(), true)` 
Then we can just do `$returnType = new GenericObjectType($collectionClass, [$modelType])`

src/ReturnTypes/ModelExtension.php
```diff
@@ -116,7 +116,11 @@ public function getTypeFromStaticMethodCall(
         }
 
         if ($methodReflection->getName() === 'all' && in_array(Collection::class, $returnType->getReferencedClasses(), true) && $methodCall->class instanceof \PhpParser\Node\Name) {
-            $returnType = new GenericObjectType(Collection::class, [new ObjectType($scope->resolveName($methodCall->class))]);
+            $class = $methodCall->class->toCodeString();
+
+            $collection = (new $class)->newCollection([]);
+
+            $returnType = new GenericObjectType(get_class($collection), [new ObjectType($scope->resolveName($methodCall->class))]);
```
canvural, line 123: Here also we can clean it up using the method that will be in `BuilderHelper`

tests/Application/app/AccountCollection.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace App;
+
+use Illuminate\Database\Eloquent\Collection;
+
+class AccountCollection extends Collection
```
canvural, line 7: The Collection class is marked as generic in the [stubs](https://github.com/nunomaduro/larastan/blob/master/stubs/EloquentCollection.stub). So any child class should also mark it as generic. You can do it like this
```php
/**
 * @template TModel
 * @extends Collection<TModel>
 */
class AccountCollection extends Collection
{}
```

tests/Features/ReturnTypes/CustomEloquentCollectionTest.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+namespace Tests\Features\ReturnTypes;
+
+use App\Account;
+use App\AccountCollection;
+
+class CustomEloquentCollectionTest
+{
+    public function testQueryBuilderHydrateReturnsCustomBuilder(): AccountCollection
```
canvural, line 10: You can add a `return` docblock here that will look something like `@phpstan-return AccountCollection<Account>` So that way we can assert it returns correct Collection class with the correct model.

(This applies to other tests where AccountCollection is returned)

tests/Application/app/Account.php
```diff
@@ -14,4 +14,9 @@ public function scopeActive(Builder $query): Builder
     {
         return $query->where('active', 1);
     }
+
```
canvural, line 17: Here there should be a docblock describing the return type

```php
/**
 * @param array<int, mixed> $models
 *
 * @return AccountCollection<Account>
 */
```

timacdonald:
Thanks so much for all this feedback @canvural - much appreciated getting pointed in the right direction. Making some more progress on this today.

szepeviktor:
👍 👍

szepeviktor:
@timacdonald You are trying **very** hard.
Thank you.

timacdonald:
Haha, that I am. This is a whole new world and I needed a few goes at it to get my head around it all and in the right frame of mind. Now I at least know how it all (kinda) comes together. Hoping to have it wrapped up next weekend 😀

timacdonald:
soOoOoooOoOOoo close now.

Got everything except for the `FactoryBuilder` return type. It is different to the others as I've got to detect if a method has *previously* been called.

Also, looks like CI is failing on `^6.0`, so there must be an API difference on one of the methods.

Should have both of these addressed in the next few days.

szepeviktor:
@timacdonald Success! 🍏

timacdonald:
Thanks @canvural and @szepeviktor! Glad to have been able to help out on the project. While my head is still in the headspace, I'd love to get the factory helper stuff done.

Any pointers on where to start with supporting `factory(User::class, 5)->create()`?

All my current work has just relied on what was passed into the final method, however this would require looking back and seeing what was called *before* the `create` method.

mr-feek:
@timacdonald you can solve model factories very simply via templates. Here's an implementation i did over at psalm https://github.com/mr-feek/psalm-plugin-laravel/pull/1/files

Issue #326, opened 2019-10-10
proposal: rules specific to laravel


canvural:
I was thinking maybe we can implement some Laravel specific PHPStan rules.

What common mistakes do you see Laravel developers do, that can be caught by static analysis?

CyberiaResurrection:
I'm not sure I've seen it recently, but the following idiom has had me scratching my head:

`
$foo = $someBuilder->first();
$foo = $foo ? $foo : null;
`
Where the second line is completely implied by the semantics of the ->first() call off the query builder.

BrandonSurowiec:
Using a Model reference in a migration could raise a warning. If the model gets changed down the line, the migration could break: https://twitter.com/calebporzio/status/1184497422187274240

crissi:
reusing names for routes

PR #538, opened 2020-04-18
Add rule for detecting expensive calls on a Collection

Daanra:
# Description

This pull request serves as a proof of concept for a new rule I'd like to have added to Larastan in light of https://github.com/nunomaduro/larastan/issues/326. 

This rule would detect method calls such as `count()` on a collection that could have been called immediately on a query builder instance. Using the query builder to determine the result uses less memory and can be much faster.

#### Example 1
To determine the number of users, one could query all users into a Collection and then call `count()` on that Collection:
```php
$num_users = User::all()->count();
```
But it is much more efficient to just user the query builder to determine the count:
```php
$num_users = User::count();
```
#### Example 2
To determine if a certain `$user` has a role with a particular name, we could query all names of the roles that belong to that user and then loop over the collection to find if it contains that particular name:
```php
$user->roles()->pluck('name')->contains('a role name');
```
But again, we can also do this in a single query:
```php
$user->roles()->where('name', 'a role name')->exists();
```

#### More examples
There are plenty more scenario's where only a query could be used to more efficiently determine something. Below are some more examples. On the left side is the incorrect / inefficient form, on the right side is the more efficient form (only using a query).

<details>

```php

User::all()->first(); // User::first()
User::all()->find([1, 2]); // User::whereIn('id', [1, 2])->get()
User::all()->pluck('id'); // User::pluck('id')
User::whereActive()->get()->only([1, 2]); // User::whereActive()->whereIn('id' [1, 2])->get()
$user->roles()->whereSomething()->get()->isEmpty(); // $user->roles()->whereSomething()->doesntExist()
User::query()->get()->modelKeys(); // User::query()->pluck('id');
User::all()->max('age'); // User::max('age');
User::all()->take(3); // User::take(3)->get();
Post::pluck('num_likes')->sum(); // Post::sum('num_likes');
User::get()->last(); // <Flip the applied ordering> User::orderBy('id', 'desc')->first()
User::all()->slice(4, 2); // User::take(2)->skip(4)->get();
User::all()->where('name', 'Taylor'); // User::where('name', 'Taylor')->get()
User::all()->contains('id', '>', 5); // User::where('id', '>', 5)->exists()
DB::table('users')->where('age', '>', 90)->get()->isNotEmpty(); // DB::table('users')->where('age', '>', 90)->exists()
```
</details>

## Implementation
The implementation I provide checks all method calls on any object that is a subtype of `Illuminate\Support\Collection`. It then checks whether that collection was obtained by calling some method on a query builder instance. Depending on what method is called on the collection and what arguments are supplied to that method, the rule might produce an error. For example, if you call sum with a closure:
```php
User::all()->sum(function (User $user): int { return $user->age; });
```
the rule does not produce an error. But if you use the shorthand notation using a string:
```php
User::all()->sum('age');
```
It will throw an error. The rule actually checks whether the argument exists as a property on your model. The following, will not report an error:
```php
User::with(['roles'])->get()->pluck('roles');
```
### Overriding methods on Custom Collections
Since this implementation supports custom Collection classes, it is possible that a user has overridden some of the native collections methods with a nonintuitive implementation. The `max()` method might do something that is not possible with a regular query and Larastan will report a false positive. We could disable the rule for methods that are called on a custom collection and are also overridden, but I do not think it's worth it as this should be, as far as I know, an extremely rare scenario.

### Proper testing
As far as I could see, there isn't a proper way to test rules yet in this repository, is there? The current tests are not that useful as they can only check whether no error message is reported, but we should really be checking the inverse.

Here some tests that should produce an error message:

<details>

```php

<?php

declare(strict_types=1);

namespace Tests\Features\Rules;

use App\User;

class NoUnnecessaryCollectionCallRuleTest
{
    public function testCallCountWrong(): int
    {
        return User::where('id', '>', 5)->get()->count();
    }

    public function testCallGetPluckWrong(): Collection
    {
        return User::query()->get()->pluck('id');
    }

    public function testCallCountWrongly(): int
    {
        return User::all()->count();
    }

    public function testCallFirstWrongly(): ?User
    {
        return User::all()->first();
    }

    public function testCallRelationTakeWrongly(): \Illuminate\Database\Eloquent\Collection
    {
        return User::firstOrFail()->accounts()->get()->take(2);
    }

    public function testDbQueryBuilder(): int
    {
        return DB::table('users')->get()->count();
    }

    public function testVarWrong(): bool
    {
        $query = User::query()->limit(3)->where('status', 'active');

        return $query->get()->isEmpty();
    }

    public function testVarWrongFirst(): ?User
    {
        return User::where('id', 1)->get()->first();
    }

    public function testContainsWrong(): bool
    {
        return User::all()->contains(User::firstOrFail());
    }

    public function testPluckCountWrong(): int
    {
        return User::query()->pluck('id')->count();
    }

    public function testCallWhereWrong(): Collection
    {
        return User::all()->where('id', '<', 4);
    }

    public function testCallDiffWrong(): Collection
    {
        return User::all()->diff([1, 2, 3]);
    }

    /**
     * @return int[]
     */
    public function testCallModelKeysWrong(): array
    {
        return User::all()->modelKeys();
    }

    public function testContainsStrictWrong(): bool
    {
        return User::query()->get()->containsStrict('id', 1);
    }

    public function testMaxWrong(): int
    {
        return User::query()->get()->max('id');
    }
}

```
</details>

#### Disclaimer
I have no previous experience with Larastan, PHPStan or PHP-parser. My implementation might be really shitty. This pull request serves more as a proof of concept. You are free to throw this implementation away. As long as something like this gets added to Larastan, I'd be really happy. If you have any suggests for improvements, I'd be happy to implement them.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -190,6 +190,11 @@ services:
             - phpstan.phpDoc.typeNodeResolverExtension
 
     -
+        class: NunoMaduro\Larastan\Rules\NoUnnecessaryCollectionCallRule
+        tags:
+            - phpstan.rules.rule
+
+    -
         class: NunoMaduro\Larastan\Types\GenericEloquentBuilderTypeNodeResolverExtension
         tags:
             - phpstan.phpDoc.typeNodeResolverExtension
--- a/extension.neon
+++ b/extension.neon
@@ -33,6 +33,18 @@ parameters:
     bootstrap: %rootDir%/../../nunomaduro/larastan/bootstrap.php
     reportUnmatchedIgnoredErrors: false
     checkGenericClassInNonGenericObjectType: false
+    noUnnecessaryCollectionCall: true
+    noUnnecessaryCollectionCallOnly: []
+    noUnnecessaryCollectionCallExcept: []
+
+parametersSchema:
+    noUnnecessaryCollectionCall: bool()
+    noUnnecessaryCollectionCallOnly: listOf(string())
+    noUnnecessaryCollectionCallExcept: listOf(string())
+
+conditionalTags:
+    NunoMaduro\Larastan\Rules\NoUnnecessaryCollectionCallRule:
+    		phpstan.rules.rule: %noUnnecessaryCollectionCall%
 
 services:
     -
@@ -196,8 +208,9 @@ services:
 
     -
         class: NunoMaduro\Larastan\Rules\NoUnnecessaryCollectionCallRule
-        tags:
-            - phpstan.rules.rule
+        arguments:
+            onlyMethods: %noUnnecessaryCollectionCallOnly%
+            excludeMethods: %noUnnecessaryCollectionCallExcept%
 
     -
         class: NunoMaduro\Larastan\Types\GenericEloquentBuilderTypeNodeResolverExtension
--- a/src/Rules/NoUnnecessaryCollectionCallRule.php
+++ b/src/Rules/NoUnnecessaryCollectionCallRule.php
@@ -241,6 +241,7 @@ class NoUnnecessaryCollectionCallRule implements Rule
 
             return $this->propertyExtension->hasProperty($modelReflection, $firstArg->value);
         }
+
         return false;
     }
 
@@ -261,6 +262,7 @@ class NoUnnecessaryCollectionCallRule implements Rule
             $class = $call->class;
             if ($class instanceof Node\Name) {
                 $modelClassName = $class->toCodeString();
+
                 return is_subclass_of($modelClassName, Model::class);
             }
         }
@@ -309,6 +311,7 @@ class NoUnnecessaryCollectionCallRule implements Rule
     protected function isNotCalledOnCollection(Node\Expr $expr, Scope $scope): bool
     {
         $calledOnType = $scope->getType($expr);
+
         return ! (new ObjectType(Collection::class))->isSuperTypeOf($calledOnType)->yes();
     }
 
@@ -319,6 +322,6 @@ class NoUnnecessaryCollectionCallRule implements Rule
      */
     protected function formatError(string $method_name): string
     {
-        return "Called '{$method_name}' on collection, but could have been retrieved as a query.";
+        return "[NoUnnecessaryCollectionCallRule] Called '{$method_name}' on collection, but could have been retrieved as a query.";
     }
 }
--- a/src/Rules/NoUnnecessaryCollectionCallRule.php
+++ b/src/Rules/NoUnnecessaryCollectionCallRule.php
@@ -189,27 +189,6 @@ class NoUnnecessaryCollectionCallRule implements Rule
     }
 
     /**
-     * Returns whether the method call was called with a string as its only argument.
-     * @param MethodCall $node
-     * @return bool
-     */
-    protected function hasStringAsOnlyArgument(MethodCall $node): bool
-    {
-        /** @var \PhpParser\Node\Arg[] $args */
-        $args = $node->args;
-
-        if (count($args) !== 1) {
-            return false;
-        }
-
-        if (!($args[0]->value instanceof Node\Scalar\String_)) {
-            return false;
-        }
-
-        return true;
-    }
-
-    /**
      * Returns whether the method call is a call on a builder instance.
      * @param Node\Expr $call
      * @param Scope $scope
--- a/src/Rules/NoUnnecessaryCollectionCallRule.php
+++ b/src/Rules/NoUnnecessaryCollectionCallRule.php
@@ -326,6 +326,6 @@ class NoUnnecessaryCollectionCallRule implements Rule
      */
     protected function formatError(string $method_name): string
     {
-        return "[NoUnnecessaryCollectionCallRule] Called '{$method_name}' on collection, but could have been retrieved as a query.";
+        return "Called '{$method_name}' on Laravel collection, but could have been retrieved as a query.";
     }
 }
--- a/src/Rules/NoUnnecessaryCollectionCallRule.php
+++ b/src/Rules/NoUnnecessaryCollectionCallRule.php
@@ -189,7 +189,7 @@ class NoUnnecessaryCollectionCallRule implements Rule
             if ($this->firstArgIsDatabaseColumn($node, $scope)) {
                 return [$this->formatError($name->toString())];
             }
-        } elseif (in_array($name->toLowerString(),  ['contains', 'containsstrict'], true)) {
+        } elseif (in_array($name->toLowerString(), ['contains', 'containsstrict'], true)) {
             // 'contains' can also be called with Model instances or keys as its first argument
             /** @var \PhpParser\Node\Arg[] $args */
             $args = $node->args;
--- a/src/Rules/NoUnnecessaryCollectionCallRule.php
+++ b/src/Rules/NoUnnecessaryCollectionCallRule.php
@@ -115,17 +115,18 @@ class NoUnnecessaryCollectionCallRule implements Rule
         );
 
         if (! empty($onlyMethods)) {
-            $this->shouldHandle = $onlyMethods;
+            $this->shouldHandle = array_map(function (string $methodName): string {
+                return Str::lower($methodName);
+            }, $onlyMethods);
         } else {
             $this->shouldHandle = $allMethods;
         }
 
         if (! empty($excludeMethods)) {
-            $this->shouldHandle = array_diff($this->shouldHandle, $excludeMethods);
+            $this->shouldHandle = array_diff($this->shouldHandle, array_map(function (string $methodName): string {
+                return Str::lower($methodName);
+            }, $excludeMethods));
         }
-        $this->shouldHandle = array_map(function (string $methodName): string {
-            return Str::lower($methodName);
-        }, $this->shouldHandle);
     }
 
     /**
--- a/src/Rules/NoUnnecessaryCollectionCallRule.php
+++ b/src/Rules/NoUnnecessaryCollectionCallRule.php
@@ -1,0 +1,262 @@
+<?php
+
+declare(strict_types=1);
+
+namespace NunoMaduro\Larastan\Rules;
+
+use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
+use Illuminate\Database\Eloquent\Relations\Relation;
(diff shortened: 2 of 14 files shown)


mfn:
I just ran this on a private project, only found one entry:
```
 ------ ------------------------------------------------------------------
  Line   …/….php
 ------ ------------------------------------------------------------------
  243    Called 'first' on collection, but could have been retrieved as a
         query.
 ------ ------------------------------------------------------------------
```
Code:
```php
        return User
            ::select('users.*')
            ->join('…')
            ->where(…)
            ->orderBy('users.id', 'asc')
            ->get()
            ->first();
```
Kazinga! Indeed this is a technical bug!

My single-hit-verdict: this has potential!

canvural:
Hi,

Thanks for the idea! This sounds really interesting! I'm on board with this.

Right now, I don't know much time to look at the implementation, but I can say some stuff about testing this.

Yes, right now we just have the high-level integration tests. But rules should not be tested in the same way. Check out the [`PHPStan\Testing\RuleTestCase`](https://github.com/phpstan/phpstan-src/blob/master/src/Testing/RuleTestCase.php) and how are the rules tested in [PHPStan](https://github.com/phpstan/phpstan-src/tree/master/tests/PHPStan/Rules). This way we can also have some feature/integration tests for levels, and also can check the failures.

extension.neon
```diff
@@ -189,6 +189,11 @@ services:
         tags:
             - phpstan.phpDoc.typeNodeResolverExtension
 
+    -
+        class: NunoMaduro\Larastan\Rules\NoUnnecessaryCollectionCallRule
```
mr-feek, line 193: I definitely think this should be opt-in if it is going to be included. This rule is quite opinionated

szepeviktor: Here at Larastan everything is one big opt-in!

mr-feek: This would be the first rule introduced to this codebase. Not everything needs to be done at the database layer, especially the likes of `firstWhere` if an index won't  be hit, unless I'm misunderstanding something.

Don't get me wrong, I'm sure there's lots of benefit and this is on the right track for sure. I do think there needs to be some customization or opt-in nature to it.

mfn: I feel the same, could be too opinionated.

Opt-in? Opt-out? Enable by default, easy to disable?

Not too acquainted on the technical level with phpstan/larastan what the options are here (instead of simply not including and having to add it manually: which could hinder users of larastan even know about it)

szepeviktor: Usage of PHPStan goes basically like this.

1. you run phpstan
1. fix everything
1. Not possible to fix? Add it to `ignoreErrors:`

szepeviktor: Although we can add this particular rule on a higher Level

mr-feek: That makes sense for actual issues, but this is quite opinionated and will likely have a lot of false positives depending on use cases

Daanra: @mr-feek Good point.

I added three new parameters to the config that allows you to customize the behavior of the rule:
```
noUnnecessaryCollectionCall: bool
noUnnecessaryCollectionCallOnly: string[]
noUnnecessaryCollectionCallExcept: string[]
```

This way a developer can easily enable or disable the rule. In addition, certain collection method calls can be ignored all together. E.g. if you only want to check `count` and `first`, you could set:
```
noUnnecessaryCollectionCallOnly: ['count', 'first']
```
in the configuration file. Similarly you can ignore specific method calls.

I'm not sure what sensible defaults would be though.

Enable the rule by default, but only check for `['count', 'first']` seems like it would already catch a fair amount of actual errors without being too opinionated. I'm also fine with disabling the rule altogether by default.

mr-feek: awesome! I'd be totally onboard with enabling it by default.

szepeviktor:
Very _little_ remaining :) https://github.styleci.io/analyses/ADe979

mfn:
Always remember: don't fiddle manually around with styleci, just download and apply the from "the button"❗️
image

This is such a time safer and keeps the annoyance out :)

szepeviktor:
Rain of arrows!

# 🌧️

BertvanHoekelen:
❤️ this PR. 

I think having `[NoUnnecessaryCollectionCallRule]` in the error output can come in handy as this rule's behaviour can be customised in the settings.

Daanra:
The reason why I prepended `[NoUnnecessaryCollectionCallRule]` to the error output is that it allows a developer to easily google and find the correct documentation concerning this rule. 

When a developer only sees `Called 'first' on collection, but could have been retrieved as a query.`, it might not be apparent what the actual issue is. They might have to dig through issues on GitHub to find the meaning behind this rule and how to configure it.

I think it might be a good idea to display the name of the rule in the error output for all of the rules that we add. I think that if we were to add more rules, it is quite likely that these rules will be a bit opinionated and for that reason, it is also likely that they will be configurable. Prepending the rule name just makes the life of a developer a bit easier (and we will probably have fewer developers opening issues here).

I do agree with you though that having the inconsistency with PHPStan's way of formatting rule errors is not that great.

mfn:
> PHPStan's way of formatting rule errors

Does an error formatter have a access to this information (phpstans options `--error-format`)? I'm just wondering if a custom formatter could solve this on a general level.

But OTOH, the vast majority of violations I had to deal with, the name of the rule didn't matter so this one, being configurable, truly stands out I guess.

canvural:
@mfn I don't think PHPStan error formatters provide that information.

I also think displaying the rule name is a good idea. But there are a lot of PHPStan rules out there and as far as I can see none of them have it in the error message like that. Maybe it's PHPStan's job to display it with the errors. 

What about changing error message to `Called 'first' on Laravel collection, but could have been retrieved as a query` or `Called 'first' on Eloquent collection, but could have been retrieved as a query` ? It'll be clear this error is related to Laravel, thus Larastan.

Daanra:
@canvural I removed the prefix from the error and changed the error message to your first suggestion. I'm guessing that should be clear enough.

src/Rules/NoUnnecessaryCollectionCallRule.php
```diff
@@ -0,0 +1,331 @@
...
+        }
+
+        /** @var \PHPStan\Type\ObjectType $iterableType */
+        if ((new ObjectType(Model::class))->isSuperTypeOf($iterableType)->yes()) {
+            $modelReflection = $this->reflectionProvider->getClass($iterableType->getClassName());
+
+            /** @var \PhpParser\Node\Scalar\String_ $firstArg */
+            $firstArg = $args[0]->value;
+
+            return $this->propertyExtension->hasProperty($modelReflection, $firstArg->value);
```
canvural, line 244: `ModelPropertyExtension` only works if we can read the database migrations. Also this will ignore if the property is defined by annotations on the class.

So I think it's better to use `return $modelReflection->hasProperty($firstArg->value);` here. And we can also remove the `ModelPropertyExtension` injection in the constructor.

canvural:
> @canvural I removed the prefix from the error and changed the error message to your first suggestion. I'm guessing that should be clear enough.

It was just a suggestion. But if no one objected, I guess we can go with that :smile: Thanks!

src/Rules/NoUnnecessaryCollectionCallRule.php
Daanra: I chose to use the ModelPropertyExtension, because as far as I understood this will tell us whether the property exists as a column in our database. 

Using `return $modelReflection->hasProperty($firstArg->value);` will result in false positives because the `hasProperty` on the `modelReflection` will return `true` for, for example, relations. 

E.g. the following should be valid:
```
User::with(['accounts'])->get()->pluck('accounts');
```

but if we use the modelReflection it will be marked as invalid.

> Also this will ignore if the property is defined by annotations on the class.

That's what we want, right? If there is a PHPDoc property on a Model, we have no idea where that property comes from. We can not assume that the property exists as a column in our database because that will lead to falsely reporting errors again.

canvural: Hmm. Yeah, that's a good point! So it's good like this.

canvural:
Thank you again for this!

Daanra:
Thanks for taking the time and merging it! It's much appreciated. 😄

src/Rules/NoUnnecessaryCollectionCallRule.php
mr-feek: I totally understand your reasoning, although I personally don't think we should rely on database migration files for this. Especially since in laravel 8, migrations are being optimized via dump commands and we won't be able to consistently rely on them moving forward. 

Although  I am biased as I use a more scalable solution [Skeema](https://github.com/skeema/skeema) for managing migrations

mr-feek: It's my understanding that since this PR is merged, this package explicitly depends on database migrations being present, or at least when this rule is enabled.

canvural: If you don't have migrations, the `ModelPropertyExtension` would return false for `hasProperty`, therefore the rule will not give you an error. It gives an error when we are really sure the given string is a database column.

mr-feek: wohoo, i inverted that check in my head when going through the logic

deleugpn:
I just got this on my CI complaining about a case where there's a `pluck` after a `get`. Truly amazing addition to Larastan. Thanks Daanra and the whole team of maintainers.
Larastan is coming up really nice even though Laravel isn't optimized for static analyses.

nunomaduro:
@Daanra Do you have twitter? I would like to make a tweet about this, but can't find your twitter handle. :/

Daanra:
@nunomaduro Unfortunately, I do not have Twitter :(

barryvdh:
I got this error on a project a few times, but it seems there are different between `->get()->count()` when `groupBy()` is used.

```php
 return $messages->groupBy('users.id')->get()->count();
```

Is not the same as
```php
 return $messages->groupBy('users.id')->count();
```

But it could be replaced with

```php
 return $messages->groupBy('users.id')->getQuery()->getCountForPagination();
```

Or in this case with

```php
 return $messages->distinct('users.id')->count();
```

The message suggests it's safe to change this, but maybe it would be possible to check groupBy?

PR #539, opened 2020-04-20
Null safe mixin acquisition

oruborus:
Ensure functionality when config is missing.

### DIFF
--- a/src/Methods/Pipes/Mixins.php
+++ b/src/Methods/Pipes/Mixins.php
@@ -62,7 +62,7 @@ final class Mixins implements PipeContract
             $classReflection->getParentClassesNames(),
             $this->resolve('config') === null ?
                 [] :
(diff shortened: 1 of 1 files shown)


canvural:
Hi, thanks for the PR! I'm just curious what is your use case for this? When would there be no config? If you are developing a package that uses Larastan you should also require [testbench]((https://github.com/orchestral/testbench)

oruborus:
> I'm just curious what is your use case for this?
I encountered this problem while using rector. Under this circumstance no config seems to be loaded.

canvural:
Interesting.

We are using config also [here.](https://github.com/nunomaduro/larastan/blob/master/src/ReturnTypes/GuardDynamicStaticMethodReturnTypeExtension.php#L41) So maybe we can add a check for null also here, and if it's null we can return the default return type.

PR #540, opened 2020-04-20
Return the default return type on missing config

oruborus:
As mentionend in [here](https://github.com/nunomaduro/larastan/pull/539#issuecomment-616539668).
I created another PR as this is not the same concern.


canvural:
Thanks!

PR #543, opened 2020-04-23
Add retry helper stub

Daanra:
Adds a stub for the `retry` helper with more accurate PHPDoc annotations.

This ensures that the following things will be seen as invalid by Larastan:

```php
public function do(): int
{
   return retry(2, function (): string { // int and string mismatch
      return 'hello';
   });
}
```

```php
retry(2, function ($param): int { // Retry does not accept a parameter in its closure
      return 5;
});
```


canvural:
Thank you!

szepeviktor:
maybe a file with procedural code should not have uppercases

PR #545, opened 2020-04-26
Add value helper extension

Daanra:
Adds an extension for the value helper function such that the return type is correctly inferred.

The following will now be seen as invalid:

```php
public function returnInt(): int
{
    return value('string'); // string and int mismatch
}

public function returnClosureInt(): int
{
   return value(function (): string {
       return 'string'; // string and int mismatch
  });
}
```


canvural:
Thank you!

PR #546, opened 2020-04-27
Fix NoUnnecessaryCollectionCallRule when query result is mixed

Daanra:
Ensures the rule is only applied when we are certain that we have a collection.

Fixes false positives when the result of a query is inferred as `mixed` instead of `Builder`.


szepeviktor:
Thank you @Daanra

nunomaduro:
@szepeviktor @Daanra Can you pull request or commit directly an new entry under the `unreleased` to the changelog?

canvural:
I updated the changelog and released [0.5.7](https://github.com/nunomaduro/larastan/releases/tag/v0.5.7) :+1:

PR #547, opened 2020-04-27
Add better collection stubs

Daanra:
Adds better stubs for Collection methods such that Larastan understands them better.

Larastan will now understand that e.g. the following things are not valid:
```php
User::all()->each(function (int $user, User $key): void { // Expects User as first parameter and int as second

});

/**
  * @param EloquentCollection<User>
  * @return Collection<User>    // Collection<User> and Collection<string> mismatch
  */
public function getNames($items)
{
   return $items->map(function (User $user): string {
       return $user->name;
  });
}
```

### DIFF
--- a/stubs/Collection.stub
+++ b/stubs/Collection.stub
@@ -55,4 +55,23 @@ class Collection implements \ArrayAccess, Enumerable
      * @return TValue|null
      */
     public function shift() {}
+
+    /**
+     * @param callable(TValue, TKey): (void|bool) $callable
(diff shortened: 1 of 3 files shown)


nunomaduro:
@Daanra Can you pull request an entry just bellow `Unreleased`  on the changelog?

canvural:
Thank you!

I updated the changelog and released [0.5.7](https://github.com/nunomaduro/larastan/releases/tag/v0.5.7) :+1:

Issue #129, opened 2018-08-22
Check routes


szepeviktor:
Would it be possible to check route classes and methods?
See https://github.com/slevomat/coding-standard/issues/412

nunomaduro:
@szepeviktor Sure! Thanks for the proposal. Great idea!

szepeviktor:
Maybe my code cloud be used
https://github.com/szepeviktor/debian-server-tools/blob/master/webserver/laravel/Commands/RouteCheckCommand.php#L55-L105

nunomaduro:
No plans at this time to implement this myself. Fell free to pull request this feature.

PR #551, opened 2020-04-28
No invalid route action rule

Daanra:
Closes https://github.com/nunomaduro/larastan/issues/129
Related to https://github.com/nunomaduro/larastan/issues/326

# Introduction

If you do not use [invokable controllers](https://laravel.com/docs/7.x/controllers#single-action-controllers), then you could have registered routes that point to a non-existing method, because the existence of a controller method action is only checked if you use invokable controllers.

This PR adds a new rule that checks if the methods used by your routes actually exist. If they do not exist, this rule will produce an error.

### Example

```php
Route::get('/users', [UserController::index, 'typo']);
```
will result in the following error:
```
Detected non-existing method 'typo' on class '\App\Http\Controllers\UserController' during route registration.
```

The following three methods of route registration are supported:
- Array with class and method as elements: 
```php
Route::get('/', [UserController::class, 'index'])
```
- Array with 'uses' parameter and string: 
```php
Route::get('/', ['uses' => '\App\Http\Controller\UserController@index'])
```

- String: 
```php
Route::get('/', '\App\Http\Controller\UserController@index')
```

### Considerations

Since this rule does not work with [route namespaces ](https://laravel.com/docs/7.x/routing#route-group-namespaces), I decided to disable it by default. 

Magic methods on controllers have to be annotated using the `@method` PHPDoc annotations on the controller (or one of its parents).

Currently, only route registration by using the imported Route facade is supported. I want to add support for a non-imported Route facade as well, but I'm not sure what the best way to do that is.

Simple expressions like:
```php
Route::get('/', UserController::class . '@index')
```
will not work. I'm not sure if there's an easy way to evaluate them.

Maybe we can add another parameter that allows a developer to specify a single global route namespace as is being done in the [default RouteServiceProvider](https://github.com/laravel/laravel/blob/master/app/Providers/RouteServiceProvider.php#L17).

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -51,3 +51,37 @@ on the `contains` method, one could set the following value:
 noUnnecessaryCollectionCallExcept: ['contains']
 ```
 
+## NoInvalidRouteAction
+
+Checks if the class and method that a route uses exist when registering a route using 
+the Route facade.
+
+#### Examples
+
+```php
+Route::get('/hello', [UserController::class, 'nonExistingMethod']);
+
+Route::match(['put', 'patch'], '/endpoint', 'App\Http\Controllers\UserTypoController@index');
+
+Route::post('/post', [
+    'uses' => 'App\Http\Controllers\UserController@typo',
+]);
+```
+
+results in the following errors:
+```
+Detected non-existing method 'nonExistingMethod' on class 'App\\Http\\Controllers\\UserController during route registration.
+Detected non-existing class 'App\Http\Controllers\UserTypoController' during route registration.
+Detected non-existing method 'typo' on class 'App\\Http\\Controllers\\UserController during route registration.
+```
+
+This rule assumes that no [namespace is set on your route groups](https://laravel.com/docs/7.x/routing#route-group-namespaces).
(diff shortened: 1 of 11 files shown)


szepeviktor:
@Daanra stealing my brain ❤️ https://github.com/szepeviktor/debian-server-tools/blob/master/webserver/laravel/Commands/RouteCheckCommand.php#L70-L95

Daanra:
@szepeviktor Nice! Your tool seems like it is still a bit more powerful, as that would also work if you define your route action as an expression (e.g. `Controller::class . '@index'`), right?

szepeviktor:
@Daanra Actually I don't know what I'm doing!

It should work for `class@method` pairs.

szepeviktor:
@Daanra I was hoping I could drop my wanna-be solution mentioned above and use Larastan to detect route problems...

Daanra:
@szepeviktor I thought that this feature was not wanted so I closed it.

I could reopen and polish it a bit if it's something that might be merged. 

The biggest problem with this approach was that it didn't work for route namespaces, but those are no longer present by default in Laravel 8 installations.

szepeviktor:
I recognize that statically analyzing routes is much harder than my run-time solution.

PR #553, opened 2020-05-01
find* method calls chained after a relationship call

8633brown:
Chaining a call to find methods after a relationship call fails. 
i.e.
```$user->roles()->findOrFail(1)```

This is more or less identical to the [BuilderModelFindExtension](https://github.com/nunomaduro/larastan/blob/12af69f657b88182fa5608e137b7c47fba386b7e/src/ReturnTypes/BuilderModelFindExtension.php) other than instead of calls being made on the builder class calls are made on a relation class.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -144,6 +144,11 @@ services:
             - phpstan.broker.dynamicMethodReturnTypeExtension
 
     -
+        class: NunoMaduro\Larastan\ReturnTypes\RelationModelFindExtension
+        tags:
+            - phpstan.broker.dynamicMethodReturnTypeExtension
+
+    -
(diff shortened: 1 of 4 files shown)


canvural:
Hi, thank you for the PR!

Actually there was another PR (#537) that did the same thing. And that PR also had the support for custom collections. I just merged that one, so this is redundant now.

But thank you anyway for your time and PR! You can try the master and if you have any issues you can open an issue or PR.

8633brown:
Awesome! Works perfectly. Apologies for the extra PR, i did look for similar PR's, i don't know how i missed that one.

Issue #550, opened 2020-04-28
Eloquent Collection map method does not always has model as first parameter of its callback


canvural:
- Larastan Version: 0.5.7
- `--level` used: max

### Description

Given the following code:
```php
/**
 * @param EloquentCollection<User> $collection
 */
public function testMapPartition(EloquentCollection $collection): SupportCollection
{
    return $collection->partition(function (User $user) {
        return $user->accounts()->count() > 1;
    })->map(function (EloquentCollection $users) {
        return $users->count();
    });
}
```

Larastan gives this error:
```
Parameter #1 $callback of method  lluminate\Database\Eloquent\Collection<App\User>::map() expects callable(App\User, int): mixed, Closure(Illuminate\Database\Eloquent\Collection): mixed given.
```

Did not test but this might also apply to other methods than `partition`

Daanra:
I think we might have to remove the ` @template TModel of Model` annotation
 from the `EloquentCollection.stub` file altogether.

Collection methods like `partition` and `groupBy` return a `EloquentCollection<EloquentCollection<TModelClass>>`, so that would be considered invalid.

We might want to treat `EloquentCollection`s the same way we treat the regular SupportCollection class and allow the EloquentCollection to enumerate over any type of value. E.g.:
```php
/**
 * @template TValue
 * @extends \Illuminate\Support\Collection<int, TValue>
 */
class Collection extends \Illuminate\Support\Collection
{
}
```

canvural:
@Daanra Yeah, we might have to do that. Logically `Eloquent Collection` should hold just model instances, that's why I added the `TModel of Model`

PR #556, opened 2020-05-02
Improve collection stubs

Daanra:
This MR ensures that an EloquentCollection can also contain non-model values.
Keys are still hardcoded as `int`s this should be changed in the future.

The more I looked into properly type-hinting the collection methods using stubs the more I realised what a can of worms I had opened. The return-types of the collection methods can be really counter-intuitive. Take for example:
```php
EloquentCollection::times(5); // Returns an EloquentCollection

EloquentCollection::times(5, function ($i) {
   return $i;
}); // Returns a SupportCollection

User::all()->mapToGroups(function (User $user) {
   return [$user->email => $user->name];
}) // Returns SupportCollection<string, EloquentCollection<string>>
```

Another problem with the EloquentCollection methods is that a lot of them use `map()` internally, and `map()` may return either a SupportCollection or an EloquentCollection. However, Laravel's PHPDoc states the return type as `static` which is technically incorrect and may result in false positives.

Further work remains to be done to completely support all the collection magic without any false positives.

Closes  https://github.com/nunomaduro/larastan/issues/550

### DIFF
--- a/stubs/EloquentCollection.stub
+++ b/stubs/EloquentCollection.stub
@@ -3,7 +3,7 @@
 namespace Illuminate\Database\Eloquent;
 
 /**
- * @template TModel of Model
+ * @template TModel
  * @extends \Illuminate\Support\Collection<int, TModel>
  */
 class Collection extends \Illuminate\Support\Collection
@@ -21,4 +21,36 @@ class Collection extends \Illuminate\Support\Collection
      * @return static<TReturn>|\Illuminate\Support\Collection<int, TReturn>
      */
     public function map($callable) {}
+
+    /**
+     * @param callable(TValue, TKey): array<mixed> $callback
+     * @return \Illuminate\Support\Collection<mixed, static<mixed>>
+     */
+    public function mapToGroups(callable $callback) {}
+
+    /**
+     * @param callable(TValue, TKey): mixed $callback
+     * @return \Illuminate\Support\Collection<mixed, mixed>
+     */
+    public function flatMap(callable $callback) {}
+
+    /**
+     * @param callable $callback
+     * @return static<mixed>|\Illuminate\Support\Collection<int, mixed>
+     */
+    public function mapSpread(callable $callback) {}
+
+    /**
+     * @param int $number
+     * @param null|callable(int, int): mixed $callback
+     * @return mixed
+     */
+    public static function times($number, callable $callback = null) {}
+
+    /**
+     * @param string|array<mixed> $value
+     * @param string|null $key
+     * @return \Illuminate\Support\Collection<int, mixed>
+     */
+    public function pluck($value, $key = null) {}
 }
--- a/stubs/EloquentCollection.stub
+++ b/stubs/EloquentCollection.stub
@@ -23,13 +23,13 @@ class Collection extends \Illuminate\Support\Collection
     public function map($callable) {}
 
     /**
-     * @param callable(TValue, TKey): array<mixed> $callback
+     * @param callable(TModel, int): array<mixed> $callback
      * @return \Illuminate\Support\Collection<mixed, static<mixed>>
      */
     public function mapToGroups(callable $callback) {}
 
     /**
-     * @param callable(TValue, TKey): mixed $callback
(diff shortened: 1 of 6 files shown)


stubs/EloquentCollection.stub
```diff
@@ -3,7 +3,7 @@
 namespace Illuminate\Database\Eloquent;
 
 /**
- * @template TModel of Model
+ * @template TModel
```
spawnia, line 6: In which case would this not be true? I get that an EloquentCollection can be downgraded to a SupportCollection by transforming it's elements to something that is not a model. However, in that case it would no longer be an EloquentCollection.

Daanra: Although it does not make a lot of sense, this is true quite often. There are two examples in the description of this MR and there is one more in the linked issue.

I really think it's something that should be fixed on Laravel's side as it's quite non-intuitive atm. However, this would be quite a breaking change so I don't see that happening any soon.

spawnia: There are ways that make it possible to put non-models in an EloquentCollection, agreed. That does not make it correct to do so. The goal of static analysis tools is not to validate nonsensical or plain wrong code, i would argue that it is correct for larastan to fail in such cases.

Daanra: But these examples aren't nonsensical or wrong, would you not agree? This is a perfectly normal piece of code:
```php
return User::all()->partition(function (User $user) {
        return $user->accounts()->count() > 1;
})->map(function (EloquentCollection $users) {
        return $users->count();
});
```
If we enforce that an EloquentCollection can only hold instances of a `Model` then Larastan will always complain about this and similar pieces of code. I do not see how we can tackle this problem in a different way without completely dropping support for `partition`, `mapToGroups` and a few other methods.

canvural: Maybe we could do something like `TModel of Model|EloquentCollection` but it seems this is not supported by PHPStan. https://phpstan.org/r/be88b35a-0f5e-4da4-ae40-93bac1e949b1

spawnia: That piece of code looks fine and quite idiomatic indeed, i 100 % agree that it should be possible to write it this way.

The EloquentCollection already overrides a few functions to return a SupportCollection instead, e.g.:

```php
    /**
     * Flip the items in the collection.
     *
     * @return \Illuminate\Support\Collection
     */
    public function flip()
    {
        return $this->toBase()->flip();
    }
```

Maybe we can try and PR additional methods onto there that also change the elements within the collection?

Until then, i am not too sure what the best way of going forward is. Your PR allows more valid programs, but also fails to detect some incorrect ones.

spawnia: Code such as the one in your example could be fixed like this, until Laravel actually fixes the EloquentCollection:

```php
return User::all()->partition(function (User $user) {
        return $user->accounts()->count() > 1;
})->toBase()->map(function (EloquentCollection $users) {
        return $users->count();
});
```

Daanra: @canvural That would still not be sufficient, as the EloquentCollection can literally hold anything. This can happen when using `mapToGroups` for example.

@spawnia Yeah it'd be perfect if it could be fixed on Laravel's side, it would then become a lot easier on our side. Though they might consider this breaking changes which means we'd still have to wait quite a while. 

I'm fine with not introducing most of these stubs and just waiting / hoping for Laravel to change their return types. However, I do think the current false positives should be fixed asap. Because apart from the false positive example in the issue, there are also false positives when you don't specify the generic type of your Collection which is quite annoying.

spawnia: See https://github.com/laravel/framework/pull/32677 😉

canvural: @spawnia Yes, I agree. Ideally, this should be handled in Laravel. But currently removing this piece of code prevents a lot of false positives. So I'm gonna merge this right now. But definelty we should discuss/improve on this in a new issue or PR!

nunomaduro:
@Daanra Just added you to the list of maintainers of this package.

PR #557, opened 2020-05-03
Ignore `Undefined variable: $factory` in factories

spawnia:
This comes up when adding `- database/factories` to the list of analysed paths.

## Before

```
$ vendor/bin/phpstan analyse
Note: Using configuration file /home/spawnia/Projects/lighthouse-example/phpstan.neon.
 14/14 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ------------------------------------ 
  Line   database/factories/UserFactory.php  
 ------ ------------------------------------ 
  20     Undefined variable: $factory        
 ------ ------------------------------------

[ERROR] Found 1 error
```

## After

```
$ vendor/bin/phpstan analyse
Note: Using configuration file /home/spawnia/Projects/lighthouse-example/phpstan.neon.
 14/14 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

[OK] No errors
```

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -31,8 +31,8 @@ parameters:
         - '#Call to an undefined method Illuminate\\Support\\HigherOrder#'
         - '#Method App\\Exceptions\\Handler::render\(\) should return Illuminate\\Http\\Response but returns Symfony\\Component\\HttpFoundation\\Response#'
         - '#Property App\\Http\\Middleware\\TrustProxies::\$headers \(string\) does not accept default value of type int#'
-        - path: %rootDir%/../../../database/factories/*
-          message: '#Undefined variable: \$factory#'
+#        - path: %rootDir%/../../../database/factories/*
+#          message: '#Undefined variable: \$factory#'
     bootstrap: %rootDir%/../../nunomaduro/larastan/bootstrap.php
     reportUnmatchedIgnoredErrors: false
     checkGenericClassInNonGenericObjectType: false
--- a/extension.neon
+++ b/extension.neon
@@ -31,8 +31,8 @@ parameters:
         - '#Call to an undefined method Illuminate\\Support\\HigherOrder#'
         - '#Method App\\Exceptions\\Handler::render\(\) should return Illuminate\\Http\\Response but returns Symfony\\Component\\HttpFoundation\\Response#'
         - '#Property App\\Http\\Middleware\\TrustProxies::\$headers \(string\) does not accept default value of type int#'
-#        - path: %rootDir%/../../../database/factories/*
-#          message: '#Undefined variable: \$factory#'
+        - path: %rootDir%/../../../database/factories/*
+          message: '#Undefined variable: \$factory#'
     bootstrap: %rootDir%/../../nunomaduro/larastan/bootstrap.php
     reportUnmatchedIgnoredErrors: false
     checkGenericClassInNonGenericObjectType: false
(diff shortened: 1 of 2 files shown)


spawnia:
I tried to trigger the error in the tests, but i don't fully grok the test setup. How can we run analysis on the factory file i added?

Daanra:
Maybe instead of ignoring the error, we can update Larastan to actually recognize the variable and its type. I'm pretty sure it can be done by overriding `hasVariableType`  method in the `Scope` class but I feel like there must be a cleaner way of doing it.

nunomaduro:
@spawnia Thanks for trying to help. I think @Daanra is correct, we should try a lit bit add support to the factory variable.

canvural:
Hi, thanks for the PR!

Currently, PHPStan does not have support for this. You can check these issues for the discussions: https://github.com/phpstan/phpstan/issues/104, https://github.com/phpstan/phpstan/issues/351

Also you can work around this by doing
```php
/** @var \Illuminate\Database\Eloquent\Factory $factory */
$factory = $factory->define(...)
...
```

And users can add this to their PHPStan config files, if they want to ignore it. So I'd say no for this PR.

spawnia:
@canvural the workaround you describe does improve autocompletion, but it does not get rid of the error.

> And users can add this to their PHPStan config files, if they want to ignore it.

I think the whole point of this package is to include configuration that is common to all Laravel projects?

canvural:
If the user don't analyze the `database` path and have `reportUnmatchedIgnoredErrors` set to true, this would cause a warning.

> I think the whole point of this package is to include configuration that is common to all Laravel projects?
Yes, but to some extent. I think we should not decide on the user's behalf to what to ignore or not.

spawnia:
> If the user don't analyze the database path and have reportUnmatchedIgnoredErrors set to true, this would cause a warning.

Good point, i just tried that. Turns out Larastan already ignores a few errors which are not triggered in one of my projects:

```
     Ignored error pattern #Result of function abort \(void\) is used# was not matched in reported errors.
     Ignored error pattern #Call to an undefined method Illuminate\\Support\\HigherOrder# was not matched in reported errors.
     Ignored error pattern #Method App\\Exceptions\\Handler::render\(\) should return Illuminate\\Http\\Response but returns
     Symfony\\Component\\HttpFoundation\\Response# was not matched in reported errors.
     Ignored error pattern #Property App\\Http\\Middleware\\TrustProxies::\$headers \(string\) does not accept default value of type int# was
     not matched in reported errors.
```

> I think we should not decide on the user's behalf to what to ignore or not.

Agreed. The interaction with `reportUnmatchedIgnoredErrors` is unfortunate indeed. I think we should be consistent and either:
1. Ignore common errors and accept that we break `reportUnmatchedIgnoredErrors`
2. Leave it completely up to the user to ignore errors

I think 2. is a more clean solution. We can still help users by listing those common ignore statements in the docs. In an ideal world, that would not be needed, but until Laravel cleans up their act/PHPStan can decipher all of its magic, it is a necessary evil.

canvural:
Yeah, I agree 2. is a better solution. I was already thinking to remove those, but never got around to do it :sweat_smile: 

So can we close this one and in another PR we can remove those ignored errors?

spawnia:
@canvural working on it.

PR #564, opened 2020-05-06
Document errors to ignore instead of forcing it upon users

spawnia:
- [ ] ~Added or updated tests~
- [x] Updated CHANGELOG.md

Relates to the discussion in https://github.com/nunomaduro/larastan/pull/557

**Changes**

- Document common errors for users to ignore

**Breaking changes**

- Do not overwrite PHPStan's default for `reportUnmatchedIgnoredErrors`
- Stop ignoring errors

### DIFF
--- a/docs/errors-to-ignore.md
+++ b/docs/errors-to-ignore.md
@@ -1,0 +1,31 @@
+## Errors To Ignore
+
+Some parts of Laravel are currently too magical for Larastan/PHPStan to understand.
+
+If you hit those errors in your project, you can add them to your `phpstan.neon` as needed.
+For example:
+
+```neon
+ignoreErrors:
+- '#Result of function abort \(void\) is used#'
+```
+
+> Tip: If you set `reportUnmatchedIgnoredErrors: true`, you will be notified once Larastan
+> properly recognizes what's going on, and you no longer have to ignore the error.
+
+### Higher Order Messages
+
+This comes up when using [higher order messages](https://laravel.com/docs/collections#higher-order-messages). 
+
+```neon
+- '#Call to an undefined method Illuminate\\Support\\HigherOrder#'
+```
+
+### Factories
+
+This comes up when you add `database/factories` to your analysed paths.
+
+```neon
+- path: database/factories/*
+  message: '#Undefined variable: \$factory#'
+```
--- a/docs/errors-to-ignore.md
+++ b/docs/errors-to-ignore.md
@@ -3,15 +3,7 @@
 Some parts of Laravel are currently too magical for Larastan/PHPStan to understand.
 
 If you hit those errors in your project, you can add them to your `phpstan.neon` as needed.
-For example:
-
-```neon
-ignoreErrors:
-- '#Result of function abort \(void\) is used#'
-```
-
-> Tip: If you set `reportUnmatchedIgnoredErrors: true`, you will be notified once Larastan
-> properly recognizes what's going on, and you no longer have to ignore the error.
+[Learn more about ignoring errors in PHPStan.](https://phpstan.org/user-guide/ignoring-errors)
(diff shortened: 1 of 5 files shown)


extension.neon
```diff
@@ -26,13 +26,7 @@ parameters:
         - Illuminate\Http\Request
     excludes_analyse:
         - *.blade.php
-    ignoreErrors:
-        - '#Result of function abort \(void\) is used#'
```
spawnia, line 30: Is this a common thing people do? It looks like the result of `abort()` actually **is** `void`.

extension.neon
```diff
@@ -26,13 +26,7 @@ parameters:
...
-        - '#Call to an undefined method Illuminate\\Support\\HigherOrder#'
-        - '#Method App\\Exceptions\\Handler::render\(\) should return Illuminate\\Http\\Response but returns Symfony\\Component\\HttpFoundation\\Response#'
```
spawnia, line 32: That one can be fixed in existing projects and is no longer required in new projects, Laravel cleaned up their act.

extension.neon
```diff
@@ -26,13 +26,7 @@ parameters:
...
-        - '#Property App\\Http\\Middleware\\TrustProxies::\$headers \(string\) does not accept default value of type int#'
```
spawnia, line 33: Also fixable and fixed in the latest Laravel version.

extension.neon
canvural: Yes. I've seen people using `return abort(401)` before. See [here](https://grep.app/search?q=return%20abort%28&filter[lang][0]=PHP) for examples.

spawnia: It kind of makes the code more clear, i can see why someone would like to do it. Still, i would let that up to personal preference and not mention it in the docs, as it is technically wrong.

canvural:
Should we increase the patch version (0.6) once this is merged?

spawnia:
> Should we increase the minor version (0.6) once this is merged?

Makes sense, there is some breakage that requires a bunch of work. Should i write up a quick upgrade guide?

extension.neon
szepeviktor: Let's tell `return abort`-ers to stop doing it.

canvural:
> Should i write up a quick upgrade guide?

Yes, that would be nice. We can maybe redirect them to the actual [docs](https://github.com/spawnia/larastan/blob/reportUnmatchedIgnoredErrors/docs/errors-to-ignore.md) in the upgrade guide.

extension.neon
```diff
@@ -26,13 +26,7 @@ parameters:
...
     bootstrap: %rootDir%/../../nunomaduro/larastan/bootstrap.php
-    reportUnmatchedIgnoredErrors: false
```
mfn, line 35: Ah, this is great!

I was using phpstan before switching to larastan a few months ago in a private project and, although I knew about this behaviour, I totally forgot about it. When I manually set this flag to true (override) I was prompted with a lot of outdated rules from my codebase! This is so hideous 😄

docs/errors-to-ignore.md
```diff
@@ -0,0 +1,31 @@
...
+
+If you hit those errors in your project, you can add them to your `phpstan.neon` as needed.
+For example:
+
+```neon
+ignoreErrors:
+- '#Result of function abort \(void\) is used#'
+```
+
+> Tip: If you set `reportUnmatchedIgnoredErrors: true`, you will be notified once Larastan
```
canvural, line 13: This defaults to `true` in PHPStan. So maybe this tip is not needed.

spawnia: Replaced with a link to PHPStan's docs.

canvural:
Thank you!

Issue #555, opened 2020-05-02
Take advantage of `@mixin` in Laravel's code itself


ondrejmirtes:
Hi,
PHPStan's dev-master now supports `@mixin` annotation (https://github.com/phpstan/phpstan-src/commit/b02ee1444440ed995c17171c456feec2c56e0cb2). I've been told that Laravel's codebase uses this annotation heavily, so maybe Larastan could take advantage of it. Maybe we could delete some of Larastan's custom code if the only thing it does is describing Laravel magic that's also described by the `@mixin` annotation.

What I'd like to see happen:

1) Require `phpstan/phpstan: ^0.12.20` (soon to be released, in a matter of days) so that people that would update Larastan but not PHPStan are not left without this functionality.
2) Go through extensions that describe Laravel's code that uses `@mixin` and try to delete them/reduce them.

(If you don't find any code in Larastan that can be deleted thanks to `@mixin` support, then `^0.12.20` doesn't have to be required.)

Thank you!

ondrejmirtes:
Also, one more idea - it's also possible that Larastan's describes some code from Laravel that would justify adding `@mixin` annotation to Laravel itself. In that case, you can just create a [stub file](https://github.com/nunomaduro/larastan/issues/555) for that class in Larastan with `@mixin` and it would work.

nunomaduro:
If I remember correctly, we are already using those `mixings` somewhere on the larastan codebase. We need to check.

ondrejmirtes:
It might be possible to simplify that code, or maybe somehow re-use this extension: https://github.com/phpstan/phpstan-src/blob/master/src/Reflection/Mixin/MixinMethodsClassReflectionExtension.php

canvural:
Actually we were doing this manually. Here: https://github.com/nunomaduro/larastan/blob/master/src/Methods/Pipes/Mixins.php Looking for `@mixin` and `@see` in the class docs, and searching the missing method in those classes.

So, I guess we can remove that. I'll look into it :+1:

canvural:
`laravel-ide-helper` package adds `@mixin \Eloquent` to the models docblock when you generate ide helper files.

`\Eloquent` is a class alias to `Illuminate\Database\Eloquent\Model` So basically it is adding a mixin for the class' actual parent. This is to have some auto-completion in the IDE.

[...]

PR #565, opened 2020-05-06
Forward calls on Model when subclass is not known

spawnia:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Avoid false-positive when calling static builder methods such as `::find()` on Model classes where the concrete subclass is not yet known.

**Breaking changes**

No, allows more valid programs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Document common errors for users to ignore ([#564](https://github.com/nunomaduro/larastan/pull/564))
 
+### Fixed
+- Avoid false-positive when calling static builder methods such as `::find()` on Model classes where
+  the concrete subclass is not yet known ([#565](https://github.com/nunomaduro/larastan/pull/565))
+
 ### Changed
 - Do not overwrite PHPStan's default for `reportUnmatchedIgnoredErrors` ([#564](https://github.com/nunomaduro/larastan/pull/564))
 
--- a/src/Methods/ModelForwardsCallsExtension.php
+++ b/src/Methods/ModelForwardsCallsExtension.php
@@ -30,7 +30,7 @@ final class ModelForwardsCallsExtension implements MethodsClassReflectionExtensi
 
     public function hasMethod(ClassReflection $classReflection, string $methodName): bool
     {
-        if (! $classReflection->isSubclassOf(Model::class)) {
+        if ($classReflection->getName() !== Model::class && ! $classReflection->isSubclassOf(Model::class)) {
             return false;
         }
 
--- a/tests/Features/Methods/ModelExtension.php
+++ b/tests/Features/Methods/ModelExtension.php
@@ -75,6 +75,19 @@ class ModelExtension
         return User::find(1);
     }
 
+    public function testFindOnGenericModel(Model $model): ?Model
+    {
+        return $model::find(1);
+    }
+
+    /**
+     * @param  class-string<Model>  $modelClass
+     */
+    public function testFindOnModelClassString(string $modelClass): ?Model
+    {
+        return $modelClass::find(1);
+    }
+
     /**
      * @return Collection<\App\User>|null
      */


canvural:
I think it's failing because the type hint is `Model` If you give a specific model like `User` or `Account` it should work.

What is the use case for this?

spawnia:
> I think it's failing because the type hint is `Model` If you give a specific model like `User` or `Account` it should work.

Right, i saw that in the test case above. It should also work with the generic `Model` type, though.

> What is the use case for this?

I hit this in Lighthouse, where we interact with user-defined models in a generic way. See https://github.com/nuwave/lighthouse/blob/f32243e71930e515a4b8661520c67d4f1ba04416/src/Schema/Directives/DeleteDirective.php#L60

canvural:
[This](https://github.com/nunomaduro/larastan/blob/master/src/Methods/ModelForwardsCallsExtension.php) extension is responsible for finding the missing methods on models. But [this](https://github.com/nunomaduro/larastan/blob/master/src/Methods/ModelForwardsCallsExtension.php#L33-L35) check prevents it running for `Model` (Model is not subclass of Model)

So you can try to change that check to `if ($classReflection->getName !== Model::class && ! $classReflection->isSubclassOf(Model::class))` and then you can see what is the output and try to improve on that.

canvural:
Seems like it works with my suggested change. But you have to remove the `@mixin` [here.](https://github.com/nunomaduro/larastan/blob/master/tests/Application/app/User.php#L19) I added that to simulate some cases where users would run `laravel-ide-helper` and it adds this mixin. You can see the discussion at #555 

So, I guess `@mixin \Eloquent` needs to be ignored somehow.

spawnia:
@canvural your proposed fix does indeed, work - thank you!

I found that this does allow some invalid programs:

```
\Illuminate\Database\Eloquent\Model::find(1);

/** @var class-string<\Illuminate\Database\Eloquent\Model> */
$model = 'Illuminate\Database\Eloquent\Model';
$model::find(1);
```

I dug around PHPStan's `ClassReflection` to check if the source of the class can be known. For example, here it can be known for sure that the passed in `Model` is an instance and calling `::find()` works:

```
    public function testFindOnGenericModel(Model $model): ?Model
    {
        return $model::find(1);
    }
```

The following example is ambigious:

```
    /**
     * @param  class-string<Model>  $modelClass
     */
    public function testFindOnModelClassString(string $modelClass): ?Model
    {
        return $modelClass::find(1);
    }
```

Is there a way to specify a class-string must be a subclass of the given type?

spawnia:
I guess my branch was not up to date, the mixin does seem to be causing problems.

spawnia:
@canvural opened an issue with PHPStan that might help to solve this use case. https://github.com/phpstan/phpstan/issues/3256

Pragmatically, i think it should be extremely rare that developers actually try to call something like `Model::find()`. If they do, the error is very immediate and clear:

```
PHP Error:  Cannot instantiate abstract class Illuminate/Database/Eloquent/Model in /var/www/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php on line 1634
```

canvural:
Hi,

PHPStan 0.12.24 is released. So I updated master branch, and [added](https://github.com/nunomaduro/larastan/commit/300ec5ac7c2227e47eea4e403a8cd8f44d180360) `Eloquent` to `mixinExcludeClasses` config option.

So if you rebase this branch on master the tests should pass. And if you can update the changelog we can merge this :+1:

spawnia:
Nicely done @canvural. It's unfortunate the types generated by `laravel-ide-helper` are so inaccurate.

mfn:
> It's unfortunate the types generated by `laravel-ide-helper`

We can improve it, what needs to be done ? 😄

spawnia:
@mfn just to be clear, i was not bashing `laravel-ide-helper`. I think it does its part quite well actually - provide useful type hints that play nicely with IDE's.

Many of the nuances of Laravel's magic can't really be expressed correctly through a nominal type system. Other parts can be expressed only through non-standard, PHPStan-specific annotations, for which proper IDE support does not exist.

PR #567, opened 2020-05-06
Add abort to list of early terminating calls

Daanra:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Adds abort to the list of early terminating function calls. See https://phpstan.org/writing-php-code/solving-undefined-variables.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -24,6 +24,8 @@ parameters:
     scopeClass: NunoMaduro\Larastan\Analyser\Scope
     universalObjectCratesClasses:
         - Illuminate\Http\Request
+    earlyTerminatingFunctionCalls:
+        - abort
     excludes_analyse:
         - *.blade.php
     ignoreErrors:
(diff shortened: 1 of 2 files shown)


szepeviktor:
Thank you very much @Daanra!

This is _almost_ perfect.

Could you add abort variations?
https://github.com/laravel/framework/blob/7.x/src/Illuminate/Foundation/helpers.php

Daanra:
> Could you add abort variations?
> https://github.com/laravel/framework/blob/7.x/src/Illuminate/Foundation/helpers.php

Do you mean `abort_if` and `abort_unless`?

If so, these are not guaranteed to always terminate. Because they only terminate based on some condition we should not add them to the list of `earlyTerminatingFunctionCalls`, right? What you can do, is [write a type specifying extension](https://phpstan.org/developing-extensions/type-specifying-extensions) for both these methods but we already have them [here](https://github.com/nunomaduro/larastan/blob/master/src/Types/AbortIfFunctionTypeSpecifyingExtension.php).

szepeviktor:
Yes.

PR #579, opened 2020-05-19
Fix false positive NoUnnecessaryCollectionCallRule

### LINKED ISSUE #578: 'avg' on collection vs query - false positive
- Larastan Version: 0.5.8
- `--level` used: 5

### Description
I started seeing this when using this library:
```
 ------ ------------------------------------------------------------------------------- 
  Line   some/file.php                                                              
 ------ ------------------------------------------------------------------------------- 
  123    Called 'avg' on Laravel collection, but could have been retrieved as a query.  
 ------ ------------------------------------------------------------------------------- 
```
Now I had a few cases in my project where this was legit and did help, but in one specific case I think it is a false positive.

```php
// snip
$avg_of_fancy_pants_relationship = $model->my_relationship()
  ->withCount(['another_relationship' => $fancyQueryGoesHere])
  ->pluck('another_relationship_count')
  ->avg();
// snip
```

Because I am taking an average of a "withCount" I dont think that can be dont with Eloquent. At least, not this way:
```php
$avg_of_fancy_pants_relationship = $model->my_relationship()
  ->withCount(['another_relationship' => $fancyQueryGoesHere])
  ->avg('another_relationship_count');
// Illuminate/Database/QueryException with message 'SQLSTATE[42S22]: Column not found: 1054 Unknown column 'another_relationship_count' in 'field list'...
```

Is this an actual false positive for the static analyzer? Is there maybe a better way to handle this?

Daanra:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Resolves #578 
**Changes**

Ensures that if you call an aggregate function on a collection without any arguments (e.g. `$collection->sum()` ) does not result in a false positive.

### DIFF
--- a/src/Rules/NoUnnecessaryCollectionCallRule.php
+++ b/src/Rules/NoUnnecessaryCollectionCallRule.php
@@ -184,8 +184,14 @@ class NoUnnecessaryCollectionCallRule implements Rule
             return [$this->formatError($name->toString())];
         } elseif ($this->isRiskyParamMethod($name)) {
             if (count($node->args) === 0) {
-                // Calling e.g. DB::table()->pluck('age')-sum()
-                return [$this->formatError($name->toString())];
+                // Calling e.g. DB::table()->pluck($columnName)-sum()
+                // We have to check whether $columnName is actually a database column
+                // and not an alias for some computed attribute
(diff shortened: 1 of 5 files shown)


canvural:
Thank you!

PR #595, opened 2020-06-08
Added missing branch alias

GrahamCampbell:
This allows developers to ask for `^0.5.9` even before it is tagged, and composer will know to download the wip code that will become the 0.5.9 release.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,6 +50,9 @@
         "preferred-install": "dist"
     },
     "extra": {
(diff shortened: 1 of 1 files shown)


canvural:
Next release will be 0.6 actually. There were some changes that can be breaking changes.

GrahamCampbell:
Another reason a branch alias is important. :P

PR #599, opened 2020-06-10
feat: use latest version of composer in travis

canvural:
Testing latest snaphot version of `composer` in travis build


canvural:
Compared to the latest `master` build, individual build times seems to be faster upto 4 minutes now :+1:

PR #575, opened 2020-05-11
Add support for tap helper

Daanra:
- [x] Added or updated tests
- [ ] Documented user facing changes
- [ ] Updated CHANGELOG.md

**Changes**

This MR adds support for the global tap helper function.

Larastan will now detect the following errors:

```php
/** @var \App\User $user */

tap($user, function (User $user): void {
    $user->update(['name' => 'New Name']);
})->typo();

tap($user)->anotherTypo();
```

Output:
```
Call to an undefined method App\User::typo().
Call to an undefined method Illuminate\Support\HigherOrderTapProxy<App\User>::anotherTypo().
```

PR #601, opened 2020-06-11
Fix tap $this bug

Daanra:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**
Updates the HigherOrderTapProxyExtension such that statements using `$this`  are now properly understood as well. See https://github.com/nunomaduro/larastan/pull/575#issuecomment-642527659

### DIFF
--- a/src/Methods/HigherOrderTapProxyExtension.php
+++ b/src/Methods/HigherOrderTapProxyExtension.php
@@ -11,6 +11,7 @@ use PHPStan\Reflection\Dummy\DummyMethodReflection;
 use PHPStan\Reflection\MethodReflection;
 use PHPStan\Reflection\MethodsClassReflectionExtension;
 use PHPStan\Type\ObjectType;
+use PHPStan\Type\ThisType;
 
 final class HigherOrderTapProxyExtension implements MethodsClassReflectionExtension
 {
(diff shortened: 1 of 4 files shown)


src/Methods/HigherOrderTapProxyExtension.php
```diff
@@ -35,9 +36,13 @@ public function getMethod(
         ClassReflection $classReflection,
         string $methodName
     ): MethodReflection {
-        /** @var ObjectType $templateType */
+        /** @var ObjectType|ThisType $templateType */
         $templateType = $classReflection->getActiveTemplateTypeMap()->getType('TClass');
 
+        if ($templateType instanceof ThisType) {
+            $templateType = $templateType->getStaticObjectType();
```
Daanra, line 43: @canvural Honestly, I'm not really sure whether this is the correct way to do it. This works, but it doesn't really feel like the *right* way to do it. I'm still too unfamiliar with the PHPStan codebase. Maybe there's a better way?

nunomaduro: Maybe @ondrejmirtes can help here.

ondrejmirtes: I don't have the context - can you show me what you're trying to achieve on an analysed code sample?

Daanra: I realised I could move this logic to tap's DynamicFunctionReturnTypeExtension instead, which makes a lot more sense, and makes it a bit simpler.

So I suppose this can be resolved.

canvural:
Thank you!

PR #602, opened 2020-06-13
Load default auth model from config

0xb4lint:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

This PR adds support for dynamic auth model load from config.

Replaces

```php
$userModel = $config->get('auth.providers.users.model');
``` 

with

```php
if ($guard = $config->get('auth.defaults.guard')) {
    if ($provider = $config->get('auth.guards.'.$guard.'.provider')) {
        if ($authModel = $config->get('auth.providers.'.$provider.'.model')) {
            return $authModel;
        }
    }
}
```

_(In the future maybe it can be extended with dynamic guard handling.)_

### DIFF
--- a/src/Methods/Pipes/Auths.php
+++ b/src/Methods/Pipes/Auths.php
@@ -43,9 +43,9 @@ final class Auths implements PipeContract
         if (in_array($classReflectionName, $this->classes, true)) {
             $config = $this->resolve('config');
 
-            $authModel = $this->getAuthModel($config);
+            $authModel = $this->getDefaultAuthModel($config);
 
-            if ($authModel) {
+            if ($authModel !== null) {
                 $found = $passable->sendToPipeline($authModel);
             }
         } elseif ($classReflectionName === \Illuminate\Contracts\Auth\Factory::class || $classReflectionName === \Illuminate\Auth\AuthManager::class) {
@@ -59,19 +59,14 @@ final class Auths implements PipeContract
         }
     }
 
-    /**
-     * Returns the default auth model from config.
-     *
-     * @return string
-     */
-    private function getAuthModel(ConfigRepository $config)
+    private function getDefaultAuthModel(ConfigRepository $config): ?string
     {
         if (
             ! ($guard = $config->get('auth.defaults.guard')) ||
             ! ($provider = $config->get('auth.guards.'.$guard.'.provider')) ||
             ! ($authModel = $config->get('auth.providers.'.$provider.'.model'))
         ) {
-            return '';
+            return null;
         }
 
         return $authModel;
--- a/src/Methods/Pipes/Auths.php
+++ b/src/Methods/Pipes/Auths.php
@@ -62,7 +62,7 @@ final class Auths implements PipeContract
     /**
      * Returns the default auth model from config.
      *
-     * @return string|void
+     * @return string
      */
     private function getAuthModel(ConfigRepository $config)
     {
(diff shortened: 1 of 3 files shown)


szepeviktor:
Thank you for your contribution! ❤️ 

Could you replace return+else-es?

0xb4lint:
Hey @szepeviktor!

I'd like to fix the style but there is a stylelint-PHPstan conflict:

stylelint wants this form:
```php
/**
 * Returns the default auth model from config.
 *
 * @return string|null
 */
private function getAuthModel(ConfigRepository $config)
{
    if (
        ($guard = $config->get('auth.defaults.guard')) &&
        ($provider = $config->get('auth.guards.'.$guard.'.provider')) &&
        ($authModel = $config->get('auth.providers.'.$provider.'.model'))
    ) {
        return $authModel;
    } else {
        return;
    }
}
```

But PHPStan will gives me the following errors:
```
 ------ ---------------------------------------------------------------------------------------------------------------------------------------- 
  Line   ReturnTypes/AuthExtension.php                                                                                                           
 ------ ---------------------------------------------------------------------------------------------------------------------------------------- 
  65     Method NunoMaduro\Larastan\ReturnTypes\AuthExtension::getAuthModel() never returns null so it can be removed from the return typehint.  
  74     Method NunoMaduro\Larastan\ReturnTypes\AuthExtension::getAuthModel() should return string|null but empty return statement found.        
 ------ ---------------------------------------------------------------------------------------------------------------------------------------- 
```

If I change `@return string|null` to `@return string|void` and remove the `else { return; }` part, like this:
```php
/**
 * Returns the default auth model from config.
 *
 * @return string|void
 */
private function getAuthModel(ConfigRepository $config)
{
    if (
        ($guard = $config->get('auth.defaults.guard')) &&
        ($provider = $config->get('auth.guards.'.$guard.'.provider')) &&
        ($authModel = $config->get('auth.providers.'.$provider.'.model'))
    ) {
        return $authModel;
    }
}
```

Then I think a PHPStan bug (or an unhandled case) occurs:
```
 ------ --------------------------------------------------------------------------------------------------------- 
  Line   ReturnTypes/AuthExtension.php                                                                            
 ------ --------------------------------------------------------------------------------------------------------- 
  54     Parameter #1 $className of class PHPStan\Type\ObjectType constructor expects string, string|void given.  
 ------ --------------------------------------------------------------------------------------------------------- 
```

There is a condition to handle `false`-ish values:
```php
if ($authModel = $this->getAuthModel($config)) {
    return TypeCombinator::addNull(new ObjectType($authModel));
}
```

So I've commited a much uglier solution and stylelint still wants to remove `return null;`

szepeviktor:
It is preferable to put errors in conditionals, normal execution unindented.

```php

        if (! ($guard = $config->get('auth.defaults.guard'))
            || ! ($provider = $config->get('auth.guards.'.$guard.'.provider'))
            || ! ($authModel = $config->get('auth.providers.'.$provider.'.model'))
        ) {
            return;
        }

        return $authModel;
```

0xb4lint:
Thank you!
I've changed the code, now styleci passes, but if I run PHPStan on the project the bug is still present:
```
 ------ --------------------------------------------------------------------------------------------------------- 
  Line   ReturnTypes/AuthExtension.php                                                                            
 ------ --------------------------------------------------------------------------------------------------------- 
  54     Parameter #1 $className of class PHPStan\Type\ObjectType constructor expects string, string|void given.  
 ------ --------------------------------------------------------------------------------------------------------- 
```

szepeviktor:
Yes, we must pass a string only.
Could you do that in all cases?

0xb4lint:
Thank you @szepeviktor!
It looks like everything works.

szepeviktor:
Thank you.
Actually we write return typehints: `?string`

0xb4lint:
I see, all right.
Shall I merge your commit into this PR?

szepeviktor:
As you wish, it is example code.

PR #529, opened 2020-04-10
Import classes and get rid of temp vars

szepeviktor:

PR #614, opened 2020-07-21
feat: introduce code style

canvural:
From the comment here: https://github.com/nunomaduro/larastan/pull/529#issuecomment-647034392

I just copied the config file from Pest. But if anyone has any suggestion we can change some configuration.

### DIFF
--- a/src/Properties/ModelProperty.php
+++ b/src/Properties/ModelProperty.php
@@ -26,9 +26,9 @@ class ModelProperty implements PropertyReflection
     public function __construct(ClassReflection $declaringClass, Type $readableType, Type $writableType, bool $writeable = true)
     {
         $this->declaringClass = $declaringClass;
-        $this->readableType = $readableType;
-        $this->writableType = $writableType;
-        $this->writeable = $writeable;
+        $this->readableType   = $readableType;
+        $this->writableType   = $writableType;
+        $this->writeable      = $writeable;
     }
 
     public function getDeclaringClass(): ClassReflection
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ composer.lock
 /vendor/
 coverage.xml
 .phpunit.result.cache
+.php_cs.cache
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,28 @@
+<?php
+
+$finder = PhpCsFixer\Finder::create()
+    ->in(__DIR__ . DIRECTORY_SEPARATOR . 'tests')
+    ->in(__DIR__ . DIRECTORY_SEPARATOR . 'src')
+    ->append(['.php_cs']);
+
+$rules = [
+    '@Symfony'               => true,
+    'phpdoc_no_empty_return' => false,
+    'array_syntax'           => ['syntax' => 'short'],
+    'yoda_style'             => false,
+    'binary_operator_spaces' => [
+        'operators' => [
+            '=>' => 'align',
+            '='  => 'align',
+        ],
+    ],
+    'concat_space'            => ['spacing' => 'one'],
+    'not_operator_with_space' => false,
+];
+
+$rules['increment_style'] = ['style' => 'post'];
+
+return PhpCsFixer\Config::create()
+    ->setUsingCache(true)
+    ->setRules($rules)
+    ->setFinder($finder);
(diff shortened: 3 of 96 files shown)


szepeviktor:
😢 Does not catch the original reason behind my PR

https://github.com/nunomaduro/larastan/blob/8c060dfa01edeade33bde1b31e6f0b4c31bdd178/src/ApplicationResolver.php#L51

canvural:
> Does not catch the original reason behind my PR
> 
> https://github.com/nunomaduro/larastan/blob/8c060dfa01edeade33bde1b31e6f0b4c31bdd178/src/ApplicationResolver.php#L51

Even with `global_namespace_import` turned on, it didn't work :slightly_frowning_face: 

Actually I'm thinking [Slevomat Coding Standard](https://github.com/slevomat/coding-standard) and PHP CodeSniffer would be better. What you guys think?

szepeviktor:
> Even with `global_namespace_import` turned on, it didn't work 🙁

It could be a bug 🐞 PhpCsFixer may ignore typehints.

src/Properties/ModelProperty.php
```diff
@@ -26,9 +26,9 @@ class ModelProperty implements PropertyReflection
     public function __construct(ClassReflection $declaringClass, Type $readableType, Type $writableType, bool $writeable = true)
     {
         $this->declaringClass = $declaringClass;
-        $this->readableType = $readableType;
-        $this->writableType = $writableType;
-        $this->writeable = $writeable;
+        $this->readableType   = $readableType;
+        $this->writableType   = $writableType;
+        $this->writeable      = $writeable;
```
mfn, line 31: Add a new line with a variable which is longer and the whole block gets reformatted again?

(seen this in other places too in the PR)

👎 on that, IMHO unnecessary for any diff and should be disabled.

nunomaduro: Personally I am not fan of this rule either.

canvural:
Does anyone have any idea why `php-cs-fixer` finds errors when CI runs with `COMPOSER_VERSION=^2.0` whereas when `COMPOSER_VERSION=^1.0` used everything is fine?

nunomaduro:
cc @GrahamCampbell

GrahamCampbell:
Why not just properly configure StyleCI? The fact that this PR exists is evidence to the time devs waste on code style.

szepeviktor:
We'd better listen to php-cs-fixer

```
   1) src/Properties/ContainerProperty.php (fully_qualified_strict_types)

      ---------- begin diff ----------

--- Original

+++ New

@@ @@

 

-    public function getReadableType(): \PHPStan\Type\Type

+    public function getReadableType(): Type

     {

         return $this->getType();

     }

 

-    public function getWritableType(): \PHPStan\Type\Type

+    public function getWritableType(): Type

@@ @@

 

-    public function isDeprecated(): \PHPStan\TrinaryLogic

+    public function isDeprecated(): TrinaryLogic

@@ @@

 

-    public function isInternal(): \PHPStan\TrinaryLogic

+    public function isInternal(): TrinaryLogic

     {

         return TrinaryLogic::createNo();

     }

 }

 

      ----------- end diff -----------
```

mfn:
I for one prefer using tools I can also use locally

GrahamCampbell:
StyleCI can be used locally too: github.com/StyleCI/CLI.

GrahamCampbell:
But, the philosophy is that you are only wasting your own time fixing CS locally.

canvural:
@nunomaduro Can you setup the StyleCI so it can create the style fix PRs automatically and also merges it automatically? According to the docs this possible, but needs to be enabled from the repository settings page.

canvural:
ping @nunomaduro

nunomaduro:
Screenshot 2020-07-28 at 14 50 43

Can you choose herE?

canvural:
I guess the 3rd option. `Automatically send and merge fix pull requests`

nunomaduro:
done

canvural:
Ok, so lets let StyleCI handle everything. Closing this.

GrahamCampbell:
NB laravel/framework uses the 4th option, rather than the 3rd option, so there's only one fix commit, and not a merge commit too.

PR #617, opened 2020-07-24
fix: dont override return type when eloquent builder returns a query builder instance

### LINKED ISSUE #613: getQuery() results in infering wrong type of result
- Larastan Version: 0.6.1 / ^0.6-dev
- `--level` used: 6

### Description
I've code like this:
```php
<?php
$result = SomeModel
    ::getQuery()
    ->select('some_model.created')
    ->where('some_model.some_column', '=', true)
    ->orderBy('some_model.created', 'desc')
    ->first();

return $result ? new CarbonImmutable($result->created) : null;
```
The model (ide-helper generated):
```php
/**
 * …
 * @property \Carbon\CarbonImmutable $created
 * @mixin \Eloquent
 */
class SomeModel extends Model
```
The reason ide-helper generates is that is that `created` is part of `SomeModel::$dates`.

The reported error:
```
 ------ ----------------------------------------------------------------
  Line   App/Foo.php
 ------ ----------------------------------------------------------------
  415    Parameter #1 $time of class Carbon\CarbonImmutable constructor
         expects string|null, Carbon\CarbonImmutable given.
 ------ ----------------------------------------------------------------
```

Larastan thinks that the result after `::getQuery()` is indeed a model => but in  fact it isn't. Those results are not hydrated and are anonymous `stdClass` which raw properties, not casted etc.

So it thinks `$result` is `SomeModel` and thus `->created` is a `CarbonImmutable` => but it isn't. It's a raw string (or could be null or literally anything).

(the code is deliberately written to not hydrate a model when not necessary as only a specific part is needed for a short duration)

canvural:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Closes #613

**Changes**

In the `BuilderHelper` we were overriding the return type whenever a query builder or eloquent builder instance was referenced in a return type. This PR adds a check to not override it if return type is just the `QueryBuilder`

**Breaking changes**

N/A

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -22,6 +22,7 @@ parameters:
         - stubs/MorphMany.stub
         - stubs/Helpers.stub
         - stubs/HigherOrderTapProxy.stub
+        - stubs/QueryBuilder.stub
     scopeClass: NunoMaduro\Larastan\Analyser\Scope
     universalObjectCratesClasses:
         - Illuminate\Http\Request
--- a/src/Methods/BuilderHelper.php
+++ b/src/Methods/BuilderHelper.php
@@ -178,8 +178,14 @@ class BuilderHelper
                 $returnType = $customReturnType;
             }
 
-            $isBuilderReferenced = (count(array_intersect([EloquentBuilder::class, QueryBuilder::class], $returnType->getReferencedClasses())) > 0)
+            $isBuilderReferenced = (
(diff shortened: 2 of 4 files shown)


canvural:
I had to add a stub file for `QueryBuilder` to support Laravel 6.x This was needed only to support Laravel 6.x, so if we want to stop support for Laravel 6.x we can remove it.

PR #618, opened 2020-07-24
Run PHPStan on Larastan only once 

szepeviktor:
...and Larastan on Laravel and Lumen sample apps only once.

From 3 minutes/job -> to 1 minute/job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,16 @@ env:
 
 jobs:
   fast_finish: true
+  include:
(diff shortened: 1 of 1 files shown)


mfn:
Maybe also time to consider GHA :)

szepeviktor:
> Maybe also time to consider GHA

Because of the hype? :)

canvural:
Thanks! LGTM!

PR #633, opened 2020-08-06
fix: return type of `associate`, `dissociate` and `getChild` methods of `BelongsTo` relations

### LINKED ISSUE #631: False negative on BelongsTo::associate()
- Larastan Version: 0.6.2
- `--level` used: 5

### Description

PHPStan reports the following error: `Method App\Post::setAuthorAttribute() should return App\User but returns App\Post`.

The `associate` method of the `BelongsTo` relation does not return the related model, but the child model: https://github.com/laravel/framework/blob/7.x/src/Illuminate/Database/Eloquent/Relations/BelongsTo.php#L216

But Larastan assumes the opposite: https://github.com/nunomaduro/larastan/blob/master/stubs/BelongsTo.stub#L11

```php

class Post extends Model
{
    public function author(): BelongsTo
    {
        return $this->belongsTo(User::class, 'author_id');
    }

    public function setAuthorAttribute(User $user): self
    {
        return $this->author()->associate($user);
    }
}
```

I have no idea how Larastan or PHPstan works, otherwise I would send a PR for this issue.

canvural:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

This PR fixes #631

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Return type of `associate`, `dissociate` and `getChild` methods of `BelongsTo` relations ([#633](https://github.com/nunomaduro/larastan/pull/633))
+
 ## [0.6.2] - 2020-07-30
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


canvural:
Hmm. The commit from StyleCI has `skip ci` So Travis did not run for this PR. We probably do not want that :sweat_smile:

PR #651, opened 2020-09-03
fix: correct parameter type for  `where`, `orWhere` and `addArrayOfWheres` in stubs

Daanra:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Updates the QueryBuilder stubs such that the following expression does not result in an error:
```php
return User::where([
        ['active', true],
        ['id', '>=', 5],
        ['id', '<=', 10]
])->get();
```


canvural:
Thank you!

PR #654, opened 2020-09-07
feat: add new view-string type

Daanra:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

This merge request introduces a new custom type to Larastan: `blade-view`.

A `blade-view` is any string that passes the `view()->exists($string)` test. I have updated the stubs to use this new type as well, such that one can no longer run into errors when trying to render a view that does not exist.

I'm not quite happy with the `blade-view` name. A CSS file will also pass the `view()->exists()` test, so I think the 'blade-' part is a bit misleading. However, simply `view` seems a bit ambiguous. I'm not sure what the best name for it is.

EDIT: Renamed `blade-view`  to `view-string`.


tests/Features/Types/BladeViewType.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Tests\Features\Types;
+
+class BladeViewType
+{
+    public function testBladeView(): void
+    {
+        $this->doSomethingWithAView('home');
+        $this->doSomethingWithAView('emails.orders.shipped');
+        $this->doSomethingWithAView('users.index');
```
Daanra, line 13: I would really like to have tests that can assert that an error is thrown when the string does not point to an existing view. However, due to how the tests are currently set up, this is quite difficult to do. I'm not sure if I should change the testing set up a bit or just leave this as it is.

canvural (approved):
Hi,

Thank you! This sounds interesting. Although it adds more dynamic analysis to the codebase, which I don't quite like, it's ok I guess.

About the name of the new type. What about `view-string`? It's clear that it represents a view in a string form. Just a suggestion.

tests/Features/Types/BladeViewType.php
canvural: That is a bigger case than this PR. It needs a huge refactoring I guess. We can discuss it in a separate issue or PR if you want to start refactoring.

canvural:
Unrelated question: Why didn't StyleCI send a PR to fix the issues? :thinking:

canvural:
Thank you! LGTM!

Error in Travis seems unrelated to this. I'll fix that later.

gazben:
@Daanra 
One thing is not clear to me from the documentation. How to handle the case where I modify the view name?

Example:
```
    /**
     * @param string $view
     * @return \Illuminate\View\View
     */
    public function renderView(string $view): View
    {
        /**
         * @phpstan-param view-string $bladeView
         * @var string $bladeView
         */
        $bladeView = 'pages.' . $view;
        return view($bladeView);
    }
```

Output:
```
 ------ -----------------------------------------------------------------------------
  Line   Http\Controllers\DashboardController.php
 ------ -----------------------------------------------------------------------------
  28     Parameter #1 $view of function view expects view-string|null, string given.
 ------ -----------------------------------------------------------------------------

 [ERROR] Found 1 error

```

Did I miss something?

Daanra:
@gazben 

Your code would work on lower levels of PHPStan, but on the maximum level it will throw an error because it's not 100% sure that it's a `view-string`. To fix this, you can type hint the `$bladeView` variable like you did. However, you used `@phpstan-param` instead of `@phpstan-var`, which is the reason why PHPStan is ignoring it, I believe.

This should work:
```php
public function renderView(string $view): View
{
   /** @phpstan-var view-string $bladeView */
   $bladeView = 'pages.' . $view;
   return view($bladeView);
}
```

gazben:
It works. Thanks

jordyvandomselaar:
This feature saved me a lot of time today. Thanks!

bobbybouwmann:
@Daanra  I think this is an awesome feature. However, the error message is very cryptic. It took me half an hour to figure out why I get this message in the first place. 
```
 ------ -----------------------------------------------------------------------------
  Line   app/Http/Livewire/Tables/TurbinesTable.php
 ------ -----------------------------------------------------------------------------
  104    Parameter #1 $view of function view expects view-string|null, string given.
 ------ -----------------------------------------------------------------------------
```
It turns out that I renamed by the view and that's why I get this error, so that is good. However, the error doesn't describe that the view is not found. Why don't we show a message like `View not found` or something?

Daanra:
@bobbybouwmann I agree with you that the error message is not clear at all. The error message is generated by PHPStan's own type checking rules and it's not trivial to alter this message. It's an unfortunate downside to using custom Types in such a scenario. An alternative would be to define our own rules to perform the type checking (similar to https://github.com/nunomaduro/larastan/pull/673), this would give us complete control over the error message, but requires quite a bit of implementation work.

szepeviktor:
I agree with you too!

szepeviktor:
Maybe adding it to the no-one-read-docs will not help :)

bobbybouwmann:
@Daanra Maybe renaming the type would already help. Something like `missing-view-string`. I believe this is the only place where this type check is used.

szepeviktor:
> Maybe renaming the type would already help. Something like `missing-view-string`. I

That is a really bad idea.

PHPStan is not the most user-friendly product, a bit esoteric.

szepeviktor:
> not the most user-friendly product,

But @phpstan is my favorite software!

PR #657, opened 2020-09-11
[POC] feat: Add model-property type

Daanra:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

## Introduction

Laravel has a lot of methods that have string parameters that represent a property of a Model class. For example, the `where()` method on a query builder could receive a model property string as its first parameter:
```php
// Here, we expect 'username' to be a property on the User model.
User::where('username', 'test')->get();
``` 

Similarly, when creating a new User:
```php
User::create([
   'name' => 'Test',
   'email' => 'test@tester.com',
]);
```
we're expected both `name` and `email` to be a property on the User class.

This pull request attempts to provide a proof of concept for easily verifying that these strings actually exists as a property on a model by introducing a new type: `model-property`.

## How does it work?

The `model-property` type is a generic type whose argument can be a subclass of `\Illuminate\Database\Eloquent\Model`. For example, the type `model-property<User>` only accepts strings that are properties on the User class. 

The property uses the `ModelPropertyExtension` under the hood, so this type is very dependent on Larastan being able to analyze the migrations. I've currently disabled the type by default. It will just be seen as a string. To enable it, you have to set `checkModelProperties: true` in your configuration file.

## Caveats

In the case of Builders methods, the current implementation can give false positives. E.g. the following examples are perfectly valid but are seen as invalid:
```php
User::selectRaw('id as newid')->pluck('newid')
User::get('users.id')
```

I could remove the typehint from all Builder `where` methods and only put them on `update()`, `create()` and, for example, on model properties such as `$fillable` and `$hidden`.

The errors are not clear at all. If a property does not exist on your model the following error will be printed:
```
Parameter #1 $property of method TestClass::test() expects model-property<App\\User>, string given.
```
which most users probably won't understand unless they read through the documentation.

The current implementation is not entirely finished yet, but before I put more time in it, I'd like to know whether this is something you'd consider adding to Larastan. If you like the idea, I could brush it up and add some more extensive testing. Any suggestions would also be appreciated.

If this is something you'd see the potential in, then I think a similar `model-relation` property would also be of great value. This could verify the strings passed to `whereHas()`, `with()`, `load()` and many more.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,9 +61,11 @@
     },
     "scripts": {
         "test:types": "phpstan analyse --ansi",
+        "test:laravel-types": "phpstan analyse --ansi --level=max --configuration ./extension.neon ./tests/Application/app",
         "test:unit": "phpunit --colors=always",
         "test": [
             "@test:types",
+            "@test:laravel-types",
             "@test:unit"
         ]
     }
--- a/src/Methods/BuilderHelper.php
+++ b/src/Methods/BuilderHelper.php
@@ -163,7 +163,8 @@ class BuilderHelper
         // This can be a custom EloquentBuilder or the normal one
         $builderName = $this->determineBuilderType($modelName);
 
-        $builderReflection = $this->broker->getClass($builderName);
+        /** @var \PHPStan\Reflection\ClassReflection $builderReflection */
+        $builderReflection = (new GenericObjectType($builderName, [new ObjectType($modelName)]))->getClassReflection();
 
         $methodReflection = $this->searchOnEloquentBuilder($builderReflection, $methodName, $modelName);
 
--- a/src/Types/GenericModelPropertyType.php
+++ b/src/Types/GenericModelPropertyType.php
@@ -1,0 +1,215 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace NunoMaduro\Larastan\Types;
+
+use NunoMaduro\Larastan\Properties\ModelPropertyExtension;
+use PHPStan\Broker\Broker;
+use PHPStan\TrinaryLogic;
+use PHPStan\Type\ClassStringType;
+use PHPStan\Type\CompoundType;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\Generic\TemplateType;
+use PHPStan\Type\Generic\TemplateTypeMap;
+use PHPStan\Type\Generic\TemplateTypeVariance;
+use PHPStan\Type\IntersectionType;
+use PHPStan\Type\MixedType;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\ObjectWithoutClassType;
+use PHPStan\Type\StaticType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+use PHPStan\Type\UnionType;
+use PHPStan\Type\VerbosityLevel;
+
+class GenericModelPropertyType extends ModelPropertyType
+{
+    /** @var Type */
+    private $type;
+
+    /** @var ModelPropertyExtension */
+    protected $propertyExtension;
+
+    public function __construct(Type $type, ModelPropertyExtension $propertyExtension)
+    {
+        $this->type = $type;
+        $this->propertyExtension = $propertyExtension;
+    }
+
+    public function getReferencedClasses(): array
+    {
+        return $this->getGenericType()->getReferencedClasses();
+    }
+
+    public function getGenericType(): Type
+    {
+        return $this->type;
+    }
+
+    public function describe(VerbosityLevel $level): string
+    {
+        return sprintf('%s<%s>', parent::describe($level), $this->getGenericType()->describe($level));
+    }
+
+    public function accepts(Type $type, bool $strictTypes): TrinaryLogic
+    {
+        if ($type instanceof CompoundType) {
+            return $type->isAcceptedBy($this, $strictTypes);
+        }
+
+        if ($type instanceof ConstantStringType) {
+            $genericType = $this->getGenericType();
+
+            if ($genericType instanceof UnionType || $genericType instanceof IntersectionType) {
+                $types = $genericType->getTypes();
+                /** @var Type $lastType */
+                $lastType = array_pop($types);
+                $newGenericType1 =  new self($lastType, $this->propertyExtension);
+                if ($genericType instanceof UnionType) {
+                    $newGenericType2 = (new self(count($types) > 1 ? new UnionType($types) : $types[0], $this->propertyExtension));
+
+                    return TrinaryLogic::createNo()->or(
+                        $newGenericType1->accepts($type, $strictTypes),
+                        $newGenericType2->accepts($type, $strictTypes)
+                );
+                }
+                $newGenericType2 = (new self(count($types) > 1 ? new IntersectionType($types) : $types[0], $this->propertyExtension));
+
+                if ($newGenericType1->accepts($type, $strictTypes)->no() || $newGenericType2->accepts($type, $strictTypes)->no()) {
+                    return TrinaryLogic::createNo();
+                }
+
+                return TrinaryLogic::createYes();
+            }
+
+            if ($genericType instanceof MixedType) {
+                return TrinaryLogic::createYes();
+            }
+
+            if (! $genericType instanceof ObjectType) {
+                return TrinaryLogic::createNo();
+            }
+
+            $classReflection = $genericType->getClassReflection();
+
+
+            if ($classReflection === null) {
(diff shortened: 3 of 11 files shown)


composer.json
```diff
@@ -61,9 +61,11 @@
     },
     "scripts": {
         "test:types": "phpstan analyse --ansi",
+        "test:laravel-types": "phpstan analyse --ansi --level=max --configuration ./extension.neon ./tests/Application/app",
```
Daanra, line 64: Kind of unrelated. Realized we did not run Larastan on the test application. Figured it would be a good addition to the current tests.

src/Methods/BuilderHelper.php
```diff
@@ -163,7 +163,8 @@ public function getMethodReflectionFromBuilder(
         // This can be a custom EloquentBuilder or the normal one
         $builderName = $this->determineBuilderType($modelName);
 
-        $builderReflection = $this->broker->getClass($builderName);
+        /** @var \PHPStan\Reflection\ClassReflection $builderReflection */
+        $builderReflection = (new GenericObjectType($builderName, [new ObjectType($modelName)]))->getClassReflection();
```
Daanra, line 167: This change was necessary so that the GenericType is already resolved when calling a builder method statically on a model class.

E.g.
```php
User::create([])
```
Before, its type would be inferred as `Builder<TModelClass of Model>` rather than `Builder<User>`

src/Types/GenericModelPropertyType.php
```diff
@@ -0,0 +1,215 @@
...
+
+class GenericModelPropertyType extends ModelPropertyType
+{
+    /** @var Type */
+    private $type;
+
+    /** @var ModelPropertyExtension */
+    protected $propertyExtension;
+
+    public function __construct(Type $type, ModelPropertyExtension $propertyExtension)
```
Daanra, line 35: The `ModelPropertyExtension` should really not be in the constructor. But I'm not sure how I can fetch an instance of the extension from a static context?

mfn:
> E.g. the following examples are perfectly valid but are seen as invalid:
> 
> ```
> User::get('users.id')
> ```

Not sure I follow, as you also used `->where()` at the beginning of the description of the PR.

Is this invalid because it's using `::get()` or because it's using `<table prefix>.<column>`?

thanks!

Daanra:
> I because it's using `<table prefix>.<column>`




It's because the prefix is infront of it, indeed. Given some simple User class, `model-property<User>` would accept only the following values: `id`, `name`, `email`, `password`. 

For that reason it might not be worth checking the builder methods because prefixing a column name will result in a false positive.

mfn:
Ah that's too bad.

In most private projects, the table prefix is a required code style.

Primary reason is that conditions coupled with joins and ambiguous column names require this and I've had one too many bug in the past where the same in select would break things later.

Daanra:
> Ah that's too bad.

I could also create a `db-column` type, or however you want to call it, which would allow for prepending the table name. We could then use this type for all `where()` like methods, and we could use the current `model-property` type for methods such like `update()` and `create()`.

mfn:
Sounds good to me!

I sure have at least one big project to run this against with lots of variations, so I think I can help testing this.

canvural:
Hey,

I was also working on this same thing 😅 But my approach is different. I didn't think model property as a type. It can be useful I think.

I can't do detailed review now. I'm on holiday until this weekend and next week.

But this is definitely something we want. So feel free to continue working on this. And when I'm back we can go through together if you want. Fixing false positives, edge cases etc.

Thank you!

src/Types/GenericModelPropertyType.php
```diff
@@ -0,0 +1,215 @@
...
+            $objectType = $type->type;
+        } elseif ($type instanceof ClassStringType) {
+            $objectType = new ObjectWithoutClassType();
+        } elseif ($type instanceof StringType) {
+            return TrinaryLogic::createMaybe();
+        } else {
+            return TrinaryLogic::createNo();
+        }
+
+        return $this->getGenericType()->accepts($objectType, $strictTypes);
```
szepeviktor, line 115: This is unreachable code.

canvural:
Hi again, 

I did some quick review. The idea is mostly fine. But as you mentioned there is one caveat:

> The errors are not clear at all. If a property does not exist on your model the following error will be printed:
 `Parameter #1 $property of method TestClass::test() expects model-property<App\\User>, string given.`
 which most users probably won't understand unless they read through the documentation.

This is because of the new type we are using. And PHPStan checks the types of the arguments passed into the methods. And when we say this model doesn't have that property, it generates the error message above.

As far as I know, there is no real way to improve that message. One hack is to have more information in the `describe` [method.](https://github.com/nunomaduro/larastan/pull/657/files#diff-6679bc1cee728af751f7f524402eebd1R51) But at best it'd be something like `Parameter #1 $property of method TestClass::test() expects a property of App\User, string given.` which is still confusing.

Another way to go about this is to write custom rules. Then we can write our own error messages. This is also how `phpstan-doctrine` extension does it. And I have a wip feature branch doing this.

I think, to sum up:

- New type allows us to just type hint it anywhere we want, and PHPStan will automatically check it. But the error message is confusing. 

- Custom rules allows us to configure better error mesages, but requires us to write more rules for every usage of model property in Laravel code. Plus it's just bound to Laravel core. User can't typehint their code to just allow a certain model's properties. But on the other hand custom rules might give us more flexibility in some cases.

I'm open to disccusion about the issues. What you think?

Daanra:
@canvural 

I'd like to add that using a Type also has the benefit of it being easier to check values that are stored in a variable. E.g.
```php
$key = 'name';
$arr = [$key => 'Daan'];
User::create($arr);
```

In addition, Rules require a lot of logic to register them, in contrast to the declarative approach of using a Type in PHPDoc. 

Maybe we can find another way to customize the error message that is printed. 
- Perhaps we could [write a custom error formatter](https://phpstan.org/developing-extensions/error-formatters) to transform the error in something that's easier to understand.
- In the end, the error message is generated by a PHPStan rule. Maybe we could override the rule itself to change the error message.
- Or maybe we can change something in PHPStan's codebase that would make it easier to override the text of a type mismatch error or just an error from a rule in general.

canvural:
All of those solutions sound a lot like cheating or abusing the PHPStan 😀

I believe it's the [CallMethodsRule](https://github.com/phpstan/phpstan-src/blob/master/src/Rules/Methods/CallMethodsRule.php) in PHPStan that gives this error. Maybe we can do something to satisy this rule. But then write our own rule that is similar to that one, but outputs the correct error message.

But then writing a single rule instead of a new type feels more simpler. Less code compared to the new type approach.

Daanra:
> All of those solutions sound a lot like cheating or abusing the PHPStan

True, it's not neat.

> But then writing a single rule instead of a new type feels more simpler. 

This will probably not just be a single rule when you want to apply this to several methods (e.g. `where`, `whereColumn`, `create` )  both in a static and non-static context right?

I could see the logic becoming really complex in the Rule because for each method you'd have to precisely define what is allowed. E.g.
```php
whereColumn($first, $operator = null, $second = null, $boolean = 'and')
```

Accepts something like:
```php
/**
 * @param model-property<TModel>|array<model-property<TModel>|int, model-property<TModel>|array{model-property<TModel>, string, ?model-property<TModel>}> $first
 * @param string|null $operator
 * @param model-property<TModel>|null $second
 * @param string|null $boolean
 */
```

Which would require quite a bit of code in a Rule, right?

But then again, when using a Type the error message would be even worse:
```
Parameter #1 $first of method Builder::whereColumn() expects array<int|model-property<App\\User>, array(model-property<App\\User>, string, model-property<App\\User>|null)|model-property<App\\User>>|model-property<App\\User>, 'test' given.
```

I do see the benefit that a Type allows you to easily specify in a declarative manner what a method should accept as its arguments. However, a Type alone is not enough to completely validate the correctness of the arguments. For example, you will still need a Rule if you want to specify that no other arguments should be passed to the `Builder::where` method if its first argument is of type `array`. 


I'm not really sure what the best way is to go. If you feel like the Rule is the better way to go, then I understand and will close this PR.

canvural:
I think we can make it work with a hybrid approach. Having a new type that will allow us to use it wherever needed, and the actual validation of model property will happen in a rule. I will try to change my POC from before to do that. Do you mind if I take some code from this PR? :)

Daanra:
@canvural 

Sounds good.

> Do you mind if I take some code from this PR?

Sure, no problem. 😄

canvural:
@Daanra Do we need `ModelPropertyType` ? I think we just need the generic type with the model. Just a `model-property` would not make sense.

Daanra:
@canvural 

Correct. I treated a regular `ModelPropertyType` just as a string. But it can be removed because without a generic type it doesn't add any benefit to be honest.

PR #673, opened 2020-10-01
Property check

canvural:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

This PR adds support for checking arguments passed to methods that expect model properties. For example

```php
// `foo` does not exist in `users` table. So an error will occur
User::where('foo', 'bar')->get();
```

## How it works?
This PR uses the types introduced in #657 by @Daanra But unlike that PR there is no check done in type level. Instead there are 2 new rules (currently, more to come) that check severy method call and inspects if any of the arguments have the new `model-property` or generic `model-property<Model>` type. If so the parameter will be checked against the model property.

### Using in custom methods
It's not limited to Laravel core methods. Any custom method can typehint that it expects properties of a model with `model-property<Model>` typehint.

```php
class HelloWorld
{
	/**
	 * @phpstan-param model-property<\App\User> $property
	 */
    public function hello(string $property)
    {
		// here we can be sure `$property` is actually a property of User model
    }
}

// Will produce error. `foo` is not a \App\User property
(new HelloWorld)->hello('foo');
```

### Stubs
New types added to Laravel core methods by the help of stubs. Currently not all methods have the new type in docblocks. But it's fairly easy to add them to stubs so I'm thinking preparing a guide of how-to, and ask people to contribute. Would be a cool and easy way to contribute to OSS and Hacktoberfest.

## Next steps

- We need to test it on more projects. I already did some testing on some open-source Laravel projects + my work projects and fixed couple of bugs. But I'm sure there are more edge cases :smile: 
- I'll add property checking to dynamic wheres. 
```php
// `whereFoo` will look at the `users` column for `foo` column. We can report if this does not exists
User::whereFoo('bar');
```

- New rule for variable/property assigning. If a variable has a type for `model-property<Model>` we should check the assigned value.
```php
/** @var model-property<\App\User>` */
$property = 'foo';

// Should error if `foo` is not a property of `\App\User`

// Same here as `someProperty`
$someObject->someProperty = 'foo';
```

### DIFF
--- a/src/Rules/ModelProperties/ModelPropertiesRuleHelper.php
+++ b/src/Rules/ModelProperties/ModelPropertiesRuleHelper.php
@@ -1,0 +1,228 @@
+<?php
+
+declare(strict_types=1);
+
+namespace NunoMaduro\Larastan\Rules\ModelProperties;
+
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use NunoMaduro\Larastan\Properties\ModelAccessorExtension;
+use NunoMaduro\Larastan\Properties\ModelPropertyExtension;
+use NunoMaduro\Larastan\Types\ModelProperty\GenericModelPropertyType;
+use NunoMaduro\Larastan\Types\ModelProperty\ModelPropertyType;
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\Annotations\AnnotationsPropertiesClassReflectionExtension;
+use PHPStan\Reflection\ClassReflection;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Reflection\ParameterReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\Constant\ConstantArrayType;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\IntegerType;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeUtils;
+use PHPStan\Type\UnionType;
+
+class ModelPropertiesRuleHelper
+{
+    /** @var ModelPropertyExtension */
+    private $modelPropertyExtension;
+
+    /** @var AnnotationsPropertiesClassReflectionExtension */
+    private $annotationExtension;
+
+    /** @var ModelAccessorExtension */
+    private $modelAccessorExtension;
+
+    public function __construct(ModelPropertyExtension $modelPropertyExtension, ModelAccessorExtension $modelAccessorExtension, AnnotationsPropertiesClassReflectionExtension $annotationExtension)
+    {
+        $this->modelPropertyExtension = $modelPropertyExtension;
+        $this->modelAccessorExtension = $modelAccessorExtension;
+        $this->annotationExtension = $annotationExtension;
+    }
+
+    /**
+     * @param MethodReflection $methodReflection
+     * @param Scope            $scope
+     * @param Node\Arg[]       $args
+     * @param ClassReflection|null  $modelReflection
+     *
+     * @return string[]
+     */
+    public function check(MethodReflection $methodReflection, Scope $scope, array $args, ?ClassReflection $modelReflection = null): array
+    {
+        $modelPropertyParameter = $this->hasModelPropertyParameter($methodReflection, $scope, $args, $modelReflection);
+
+        if (count($modelPropertyParameter) !== 2) {
+            return [];
+        }
+
+        /** @var int $parameterIndex */
+        /** @var ObjectType $modelType */
+        [$parameterIndex, $modelType] = $modelPropertyParameter;
+
+        $modelReflection = $modelType->getClassReflection();
+
+        if ($modelReflection === null) {
+            return [];
+        }
+
+        if ($modelReflection->isAbstract()) {
+            return [];
+        }
+
+        if ($modelReflection->getName() === Model::class || ! $modelReflection->isSubclassOf(Model::class)) {
+            return [];
+        }
+
+        if (! array_key_exists($parameterIndex, $args)) {
+            return [];
+        }
+
+        $argValue = $args[$parameterIndex]->value;
+
+        if (! $argValue instanceof Node\Expr) {
+            return [];
+        }
+
+        $argType = $scope->getType($argValue);
+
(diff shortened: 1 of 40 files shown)


canvural:
@Daanra Can you da a review?

Daanra:
@canvural Yes. I'll try to make some time either tomorrow or the day after.

Daanra:
I'm getting quite a few false positives when testing this on some of my projects.

The main problems comes from columns that are present on pivot tables. E.g.:
```php
$user->groups()->where('group_created_at', '>=', now()->subWeek())->get();
```

I think, for now, the easiest would be to just return no errors in the rule when the method is called on an instance of `BelongsToMany`. And perhaps in a later iteration we could improve this.

tests/Rules/Data/model-property-builder.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+\App\User::query()->firstWhere('foo', 'bar');
```
Daanra, line 3: I think it would be nice to test a  column with a table prefix as well

canvural:
> I'm getting quite a few false positives when testing this on some of my projects.
> 
> The main problems comes from columns that are present on pivot tables. E.g.:
> 
> ```
> $user->groups()->where('group_joined_at', '>=', now()->subWeek())->get();
> ```
> 
> I think, for now, the easiest would be to just return no errors in the rule when the method is called on an instance of `BelongsToMany`. And perhaps in a later iteration we could improve this.

Isn't that code trying to find the `group_joined_at` at the Group model? Don't you need `wherePivot` to search the pivot table?

tests/Rules/Data/model-property-builder.php
canvural: If there is a `.` in the column name we currently skip it. So I didn't add test for it.

Daanra:
>Isn't that code trying to find the group_joined_at at the Group model? Don't you need wherePivot to search the pivot table?

Under the hood a join is performed and the where clause will match any joined table. `wherePivot`  seems to just call `where` under the hood and automatically prefixes the column name with the name of the pivot table to prevent ambiguity. So I suppose in all scenario's it's better to use `wherePivot` if you're using a pivot column, but `where` is also valid if the column name is unique with respect to the joined tables.

I'm fine with throwing an error when using `->where($pivotColumn)`, but maybe we can at least improve the error message to suggest that if the column exists on a pivot table that `->wherePivot` should be used instead?

canvural:
> I'm fine with throwing an error when using `->where($pivotColumn)`, but maybe we can at least improve the error message to suggest that if the column exists on a pivot table that `->wherePivot` should be used instead?

That'd be a good idea! I'll try to implement that.

canvural:
@Daanra How should we display the error for that case? There may be two reasons why column does not exists. First it can be that column really doesn't exist in related model. Second case is when it doesn't exist in pivot table. But we really don't know that.

So should we display 2 errors for that line? One for missing property. Second, a warning about that column name is ambiguous and tell user to prefix it with table name or use wherePivot.

Or just warn about the ambiguity of the colum name?

Daanra:
> So should we display 2 errors for that line? One for missing property. Second, a warning about that column name is ambiguous and tell user to prefix it with table name or use wherePivot.

I'm personally not a fan of displaying two errors. I think in this scenario a single error message with a small hint appended is best. Something along the lines of:
```
Property 'foo' does not exist in App\\User model. If 'foo' exists as a column on the pivot table, then use 'wherePivot' instead.
```

This hint should only be appended when the query is called on a subclass of `BelongsToMany`

src/Rules/ModelProperties/ModelPropertiesRuleHelper.php
```diff
@@ -0,0 +1,215 @@
...
+            }
+
+            foreach ($valueTypes as $valueType) {
+                // It could be something like `DB::raw`
+                // We only want to analyze strings
+                if (! $valueType instanceof ConstantStringType) {
+                    continue;
+                }
+
+                // TODO: maybe check table names and columns here. And for JSON access maybe just the column name
```
Daanra, line 113: I think a simple:
```php
$columnName = Str::after($valueType->getValue(), '.');
```

would be nice here (and below). So that we can also check the column name if it contains a table prefix.

canvural: I don't understand. We are already checking in the next line if the value contains a dot. There is no point of getting the substring after the dat and checking again.

canvural: Ah, I misunderstood I guess. You are suggesting a fix to this TODO. 

Yeah, we can do that. But then we need to be able to check the column in the given table, not the model we know. So it has to ask to `MigrationHelper` if that table exists and if that column is in that table.

Daanra: Ah for a second I thought ignoring the table would be enough here. But that's not the case when tables have been joined together.

Daanra:
LGTM!

I suppose we still need some docs before releasing this though.

canvural:
Yeah, we need docs. I'm working on it :+1: 

Plus we need to make sure every Laravel core function has the new type `model-property` in the stubs. Thinking about making a new issue for that and get some help from Hacktoberfest power :smile: 

And there is still stuff to do I wrote in the `Next Steps` section at the PR description.

mfn:
I just read https://github.com/nunomaduro/larastan/releases/tag/v0.6.5 and upgraded:
```
$ ./composer.phar show nunomaduro/larastan|grep versions
versions : * v0.6.5
```

This example does not report anything my codebase using 0.6.5:
```php
        Profile::where('column_does_not_exist', 'true')->get();
```
What are the requirements to make this work? How are the properties inferred?

Thanks!

Daanra:
@mfn Did you add:

```
parameters:
    checkModelProperties: true
```

to your config?

mfn:
It seems to ignore columns entirely which use the explicit table prefix, i.e. they contain a dot: `->where('tablename.column_does_not_exist', …`

Is this expected / a limitation or should I open an issue?

canvural:
Hi @mfn ,

I'll try to update the docs so it'll mention earlier that it is disabled by default :+1: 

> Is this expected / a limitation or should I open an issue?

Yes, this is expected. If the value contains a dot we simply ignore the check. Because there is no good way to determine which model it should be checked on.

This can be improved maybe. By comparing the model's table and the table name given in the argument. But it'll not work in every case.

mfn:
Damn 😄 

In our projects it's a requirement to us the "full form", because it happens to easily that joins shadow same-named columns and the only way to avoid this was the brute-force approach to make a rule to always use the table name too.

PR #679, opened 2020-10-08
feat: add support for value and orWhere methods on Eloquent Builder

robbplo:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Adds support for value and orWhere methods on the Eloquent query builder

### DIFF
--- a/tests/Rules/Data/model-property-builder.php
+++ b/tests/Rules/Data/model-property-builder.php
@@ -24,6 +24,9 @@ function bar(\Illuminate\Database\Eloquent\Builder $builder): \Illuminate\Databa
 \App\User::query()->where(getKey(), '=', 'foo');
 \App\User::query()->orWhere('foo', '=', 'foo');
 \App\User::query()->orWhere('foo', 'foo');
+\App\User::query()->orWhere([
+    'foo' => 'foo',
+]);
 \App\User::query()->value('foo');
 
 function getKey(): string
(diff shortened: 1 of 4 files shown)


tests/Rules/Data/model-property-builder.php
```diff
@@ -22,6 +22,8 @@ function bar(\Illuminate\Database\Eloquent\Builder $builder): \Illuminate\Databa
 
 \App\User::query()->whereColumn('foo', '=', 'foo');
 \App\User::query()->where(getKey(), '=', 'foo');
+\App\User::query()->orWhere('foo', '=', 'foo');
```
szepeviktor, line 25: Could you add this?

```suggestion
\App\User::query()->orWhere('foo', '=', 'foo');
\App\User::query()->orWhere('foo', 'foo');
```

szepeviktor: Thank you.

tests/Rules/Data/model-property-builder.php
```diff
@@ -22,6 +22,9 @@ function bar(\Illuminate\Database\Eloquent\Builder $builder): \Illuminate\Databa
...
+\App\User::query()->orWhere('foo', 'foo');
```
canvural, line 26: `orWhere` also accepts arrays. With the array key being the model property. So can you add a test case like
```php
\App\User::query()->orWhere([
	'foo' => 'bar'
]);
```

canvural:
All looks good! Thank you!

PR #686, opened 2020-10-15
feat: property check on dynamic wheres

canvural:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

This PR adds support for model property check to dynamic wheres. For example
```php
User::whereIdAndEmaiil(...)->get();
```

would produce an error. Because while `id` exists as a property on `User` class, `emaiil` does not.

This does not produce the same error message as our property check rule. It will just complain about that method not exists in the class. I guess this is ok. Developers will see the missing method error and can figure out it's coming from a missing property.

The one thing I wasn't certain about is, should this check be also under the `checkModelProperties` flag? I think it should. Or maybe a new flag like `checkPropertiesonDynamicWheres`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- Support fot checking model properties on dynamic wheres ([#686](https://github.com/nunomaduro/larastan/pull/686))
+
 ## [0.6.5] - 2020-10-15
 
 This release introduces a new rule that can check the arguments of methods that expects a model property name, and can warn you if the passed argument is not actually a property of the model. You can read the details about the rule [here](https://github.com/nunomaduro/larastan/blob/master/docs/rules.md#modelpropertyrule).
--- a/extension.neon
+++ b/extension.neon
@@ -332,3 +332,7 @@ services:
     -
         class: NunoMaduro\Larastan\Rules\ModelProperties\ModelPropertiesRuleHelper
 
+    -
+        class: NunoMaduro\Larastan\Methods\BuilderHelper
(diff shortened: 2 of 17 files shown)


mfn:
But it does not check `->where<Column>()`, or?

I don't use combinations but (sometimes) the above and made a deliberate error with 0.6.6 but didn't seem to get caught with `checkModelProperties: true`.

thanks

canvural:
@mfn It does check those also. Can you create a new issue and share some code snippets?

mfn:
> It does check those also. Can you create a new issue and share some code snippets?

I didn't create a new issue yet, because I'm still not confident if this PR and what I suggested it truly supposed to work or not. Once this is cleared, I'm happy to 😄 

But before that, please let me once more verify: it this supposed to check the (non-static!) way of invoking those methods (i.e. `->where…()` or just the static ones `::where…()`?

This works / no error, because there is a column `profiles.type`:
```php
Profile::whereType('test');
```
This generates an error now in *stan (correctly 👍 ) because there's no `profiles.yolo`:
```php
Profile::whereYolo('test');
```
produces
```
  57     Call to an undefined static method App\Profile::whereYolo().
```

But this does not generate an error (although there's no `profiles.yolo` column):
```php
Profile::whereType('test')->whereYolo('test');
```
I'm testing this right now with phpstan 0.12.58 and larastan 0.6.10 on level 6.

Can you confirm that the last example also should throw an error from *stan?

Thank you!

canvural:
Yes `Profile::whereType('test')->whereYolo('test');` should also throw error. And I guess I know why it's not working :smile: So, please create an issue, and I'll try to fix it. :+1:

mfn:
Thank you => https://github.com/nunomaduro/larastan/issues/726

Issue #690, opened 2020-10-19
withTrashed() calls on BelongsTo are no longer allowed since version 0.6.6


jdrieghe:
- Larastan Version: 0.6.6
- `--level` used: 2

### Description
Phpstan/Larastan complains about `->withTrashed()` calls on BelongsTo relationship definitions since version 0.6.6.
This was fine in version 0.6.5 and lower.

### Laravel code where the issue was found

```php
    public function category(): BelongsTo
    {
        return $this->belongsTo(Category::class)->withTrashed();
    }

    public function supplier(): BelongsTo
    {
        return $this->belongsTo(Supplier::class)->withTrashed();
    }
```

```
 ------ --------------------------------------------------------------------------------------------------------------------------- 
  Line   Invoice/ReadModels/Mysql/EnteredInvoice.php                                                                                
 ------ --------------------------------------------------------------------------------------------------------------------------- 
  93     Call to an undefined method Illuminate\Database\Eloquent\Relations\BelongsTo<Invoicr\Categories\Category>::withTrashed().  
  98     Call to an undefined method Illuminate\Database\Eloquent\Relations\BelongsTo<Invoicr\Suppliers\Supplier>::withTrashed().   
 ------ --------------------------------------------------------------------------------------------------------------------------- 
```

canvural:
Hi,

I think in previous versions this was working by "mistake". Unknown method calls on relation was treated as `DummyMethodReflection` With the recent refactorings this is not the case anymore.

So I'd say this is a future request to support `SoftDelete` trait calls on relations.

WouterSioen:
@canvural any hints on how we could implement this support for SoftDelete? We expected larastan to get into the `RelationForwardsCallsExtension` extension, but it didn't seem to get into that code.

canvural:
@WouterSioen Are checking it in the `hasMethod` method? It should get in there?

The problem in general is; [here](https://github.com/nunomaduro/larastan/blob/master/src/Methods/RelationForwardsCallsExtension.php#L38) we just try to find the method in the builder. So before that line there needs to be a check like
```php
if ($relatedModel->hasMethod($methodName)->yes()) {
    return true;
}
```

Then in `getMethod` we need to check it again and return correct method reflection.
```php
if ($relatedModel->hasMethod($methodName)->yes()) {
    return new EloquentBuilderMethodReflection(.....);
}
```

canvural:
Actually the check needs to happen if we can't find it in the builder. That makes more sense :+1:

PR #692, opened 2020-10-20
feat: support SoftDeletes methods on relations

jdrieghe:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Issue: #690

**Changes**

This introduces additional test cases to prove there is a regression issue that was introduced in version 0.6.6.
Methods from the `SoftDeletes` trait are not recognized on Relations even though this is supported through `ForwardsCalls` in Laravel.

---

Note that this PR does not yet introduce a fix but aims to clarify the issue defined in #690. I tried coming up with a solution but couldn't find it yet. Any pointers as to how to approach this would be welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- `SoftDeletes` methods on relations are no longer marked as undefined ([#692](https://github.com/nunomaduro/larastan/pull/692))
 - Generic model type is preserved when `with` method is used on a model instance.
 
 ## [0.6.6] - 2020-10-17
--- a/src/Methods/BuilderHelper.php
+++ b/src/Methods/BuilderHelper.php
@@ -6,10 +6,12 @@ namespace NunoMaduro\Larastan\Methods;
 
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
 use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Database\Eloquent\SoftDeletes;
 use Illuminate\Database\Query\Builder as QueryBuilder;
 use Illuminate\Support\Str;
 use NunoMaduro\Larastan\Reflection\EloquentBuilderMethodReflection;
 use PHPStan\Reflection\ClassReflection;
+use PHPStan\Reflection\Dummy\DummyMethodReflection;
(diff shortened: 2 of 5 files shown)


canvural:
Hi,

Thanks for the failing test case.

I tried to explain a little bit what has to be done here: https://github.com/nunomaduro/larastan/issues/690#issuecomment-712709592

Let me know if you have any more questions.

jdrieghe:
@canvural I couldn't figure out how to return the proper `MethodReflection` for the `SoftDeletes` methods as they are added through a macro. I introduced the formerly used `DummyMethodReflection` for now to fix the BC break introduced in `0.6.6`.

If you have a good solution on how to resolve the correct `MethodReflection`, I'm all ears. Thanks for your insights so far.

canvural:
Hi,

Using `DummyMethodReflection` is not a good solution. We would lose the type information after the call. 

As I wrote in https://github.com/nunomaduro/larastan/issues/690#issuecomment-712709592 we should use the `EloquentBuilderMethodReflection` as the method reflection.

So in `hasMethod` method in `RelationForwardsCallsExtension` class, if `$returnMethodReflection` is `null` we need to check the method in model instead. `$relatedModel->hasMethod($methodName)->yes()` like this, and return the result.

Then in `getMethod` method again we can check if `$returnMethodReflection` is `null` we can return the reflection for the method with something like:
```php
$originalMethodReflection = $relatedModelReflection->getMethod($methodName, new OutOfClassScope());

return new EloquentBuilderMethodReflection(
    $methodName, $classReflection, $originalMethodReflection,
    ParametersAcceptorSelector::selectSingle($originalMethodReflection->getVariants())->getParameters(),
    new GenericObjectType($classReflection->getName(), [$relatedModel]),
    false
);
```

jdrieghe:
Thanks @canvural. It's working now with the correct `MethodReflection`. The biggest issue I had was figuring out what scope to use, so your example with the `OutOfClassScope` was very helpful :clap:

canvural:
Perfect, thank you! Will merge when the build passes.

szepeviktor:
I was waiting for this as v0.6.6 was producing SoftDelete-related errors.

jdrieghe:
@szepeviktor I encountered the issue on one of our internal projects, so it was mostly out of a personal need. Happy to have helped others though.

canvural:
Released: https://github.com/nunomaduro/larastan/releases/tag/v0.6.7

PR #695, opened 2020-10-21
feat: add support for file method call on Request

### LINKED ISSUE #248: Cannot call method store() uploaded file from request

canvural:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Fixes #248 

**Changes**

This PR adds dynamic return type support for `$request->file()` calls.

### DIFF
--- a/src/ReturnTypes/RequestExtension.php
+++ b/src/ReturnTypes/RequestExtension.php
@@ -5,12 +5,16 @@ declare(strict_types=1);
 namespace NunoMaduro\Larastan\ReturnTypes;
 
 use Illuminate\Http\Request;
+use Illuminate\Http\UploadedFile;
 use PhpParser\Node\Expr\MethodCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\MethodReflection;
+use PHPStan\Type\ArrayType;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
-use PHPStan\Type\MixedType;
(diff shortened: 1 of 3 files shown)


src/ReturnTypes/RequestExtension.php
```diff
@@ -41,6 +45,16 @@ public function getTypeFromMethodCall(
...
     ): Type {
-        return new MixedType;
+        $uploadedFileType = new ObjectType(UploadedFile::class);
+
+        if (count($methodCall->args) === 0) {
+            return new ArrayType(new IntegerType(), $uploadedFileType);
+        }
+
+        if (count($methodCall->args) === 1) {
+            return TypeCombinator::addNull($uploadedFileType);
```
Daanra, line 55: I don't think this is correct.

```php
$request->file('documents')
```
may return an array as well. It depends on the request.

If the request has the keys: `documents[0]` and `documents[1]` then the `$request->file('documents')` returns an array.

canvural: Didn't know that. Good point. Thank you :+1:

canvural: Updated :+1:

PR #696, opened 2020-10-22
Exclude test providers from autoloading

### LINKED ISSUE #683: Unregistered service provider is executed during analysis
- Larastan Version: 0.6.4
- `--level` used: 8

### Description
During analysis of a package, processing stops because a LogicException is thrown in a provider. This is strange because it happens in code I wrote. That code should not be executed, just analyzed statically.

The provider is a test class that extends the actual exposed provider. It exists to test error handling of the provider and does it by creating a situation that triggers the exception. 

The test provider is not registered and not exposed through composer.json. 

Excluding the file from analysis does not solve the situation, the exception is still thrown.

kwebble:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

Resolves #683 

**Changes**
Previously an instance of all detected service providers was created during analysis. This is changed. Now  classes with namespaces defined in psr-4 paths in the _autoload-dev_ configuration of `composer.json` are skipped. 

**Breaking changes**
Since a number of classes previously autoloaded are skipped runtime behavior is changed. But I don't think it changes the outcome of the analysis.

### DIFF
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -17,7 +17,7 @@ final class ApplicationResolver
 {
     use CreatesApplication;
 
-    /** @var mixed $composer */
+    /** @var mixed */
     public static $composer;
 
     /**
@@ -99,7 +99,7 @@ final class ApplicationResolver
         // Create array of dev classes from Composer configuration.
         $devClasses = [];
         $autoloadDev = self::$composer['autoload-dev'] ?? [];
-        $autoloadDevPsr4 = $autoload4['psr-4'] ?? [];
+        $autoloadDevPsr4 = $autoloadDev['psr-4'] ?? [];
         foreach ($autoloadDevPsr4 as $path) {
             $devClasses = array_merge($devClasses, array_keys(ClassMapGenerator::createMap($path)));
         }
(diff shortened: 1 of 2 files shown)


src/ApplicationResolver.php
```diff
@@ -92,10 +96,18 @@ private static function getProjectClasses(string $namespace, string $rootDir): a
             $maps = array_merge($maps, ClassMapGenerator::createMap($dir));
         }
 
+        // Create array of dev classes from Composer configuration.
+        $devClasses = [];
+        foreach (self::$composer['autoload-dev']['psr-4'] as $path) {
```
szepeviktor, line 101: `autoload-dev` may not exist.
Could you check it?

szepeviktor: `?? []` or something...

src/ApplicationResolver.php
```diff
@@ -98,7 +98,9 @@ private static function getProjectClasses(string $namespace, string $rootDir): a
 
         // Create array of dev classes from Composer configuration.
         $devClasses = [];
-        foreach (self::$composer['autoload-dev']['psr-4'] as $path) {
+        $autoloadDev = self::$composer['autoload-dev'] ?? [];
+        $autoloadDevPsr4 = $autoload4['psr-4'] ?? [];
```
szepeviktor, line 102: @canvural Where does `$autoload4` come from?

🛰️

canvural: 🤦 I don't know..

canvural:
Thank you very much @kwebble !

PR #704, opened 2020-10-29
Remove mixin pipe

canvural:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

This PR removes the mixin pipe class. PHPStan has native support for `@mixin` annotations so this class is not needed anymore. And after removing it, there were a couple of issues, those are also fixed.


nunomaduro (approved):
Nice work

tests/Features/ReturnTypes/Facades.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Features\ReturnTypes;
+
+use Illuminate\Http\Client\PendingRequest;
+use Illuminate\Support\Facades\Http;
```
szepeviktor, line 8: Could it be that this class was not-existent in Laravel 6?

szepeviktor (changes requested):

tests/Features/ReturnTypes/Facades.php
nunomaduro: Http got introduced in Laravel 7.

PR #710, opened 2020-11-04
Add dd to list of early terminating functions

Daanra:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Adds `dd` to the list of early terminating function calls. I could possibly add `ddd` as well, but since the method definition does not reside in the Laravel package I decided to omit it.


canvural (approved):

szepeviktor:
Thank you!

PR #713, opened 2020-11-16
Model::with() accepts array of Closure

### LINKED ISSUE #712: Error when passing array<string, Closure> to Model::with()
- Larastan Version: 0.6.9
- `--level` used: 5

### Description
According to docs Laravel it's possible to pass Closure in array of params for `Model::with()`: https://laravel.com/docs/8.x/eloquent-relationships#constraining-eager-loads

Larastan reports the following error:
`Parameter #1 $relations of static method Illuminate\Database\Eloquent\Model::with() expects array<string>|string, array<int|string, (Closure(mixed): mixed)|string> given.`

```php
Taxonomy::with([
	'taxonomyTerms' => function ($relation) {
		return $relation->orderBy('_lft');
	},
	'taxonomyTerms.tag',
])

```

kvas-damian:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Fixes #712

**Changes**

Ensures that if you call a Model::with() method with array of Closures it does not result in a false positive.

### DIFF
--- a/stubs/Model.stub
+++ b/stubs/Model.stub
@@ -19,7 +19,7 @@ abstract class Model implements \JsonSerializable, \ArrayAccess
     /**
      * Begin querying a model with eager loading.
      *
-     * @param  string[]|string  $relations
+     * @param  string|array<int|string, (\Closure)|string>  $relations
      * @return \Illuminate\Database\Eloquent\Builder<static>
      */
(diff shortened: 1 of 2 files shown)


canvural:
Looks good! Thank you @kvas-damian

PR #719, opened 2020-11-28
feat: infer custom collection type on model relation properties

canvural:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Related: https://github.com/nunomaduro/larastan/issues/436#issuecomment-734814388

**Changes**

Accessing model relation as a property will return custom collection if the related model has one.


cweiske:
Thank you!

PR #722, opened 2020-11-28
Improve GHA configuration

szepeviktor:
- Proper YAML code
- Validate Composer config

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,51 +1,52 @@
-name: tests
+name: "tests"
 
 on:
   push:
   pull_request:
   schedule:
-    - cron: '0 0 * * *'
+    - cron: "0 0 * * *"
 
 jobs:
   tests:
-
-    runs-on: ubuntu-latest
     strategy:
       fail-fast: true
       matrix:
-        php: [7.2, 7.3, 7.4, 8.0]
-        laravel: [^6.0, ^7.0, ^8.0]
+        php: ["7.2", "7.3", "7.4", "8.0"]
+        laravel: ["^6.0", "^7.0", "^8.0"]
         exclude:
-          - php: 7.2
-            laravel: ^8.0
+          - php: "7.2"
+            laravel: "^8.0"
 
-    name: P${{ matrix.php }} - L${{ matrix.laravel }}
+    name: "P${{ matrix.php }} - L${{ matrix.laravel }}"
+
+    runs-on: "ubuntu-latest"
 
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


szepeviktor:
All green 🍏

.github/workflows/tests.yml
```diff
@@ -1,51 +1,52 @@
-name: tests
+name: "tests"
 
 on:
   push:
   pull_request:
```
canvural, line 5: We should also change something here. For a PR whole GA runs twice. One for push one for PR. IS there a way to avoid that @szepeviktor ?

szepeviktor: You know I'm a Travis guy. I think GHA is nowadays' hype train.

How about listing "all" our branches in `push:` ? Meaning `master`.

canvural: Worth a try I guess :man_shrugging:

.github/workflows/tests.yml
```diff
@@ -1,51 +1,52 @@
...
+          tools: "composer:v2"
+          coverage: "none"
+
+      - name: "Check Composer configuration"
+        run: "composer validate --strict"
 
-      - name: Prepares repository
-        run: test "$(find . -type f -not -path "./.git/*" -perm /111)" == ./tests/laravel-test.sh
+      - name: "Check file permissions"
+        run: "test \"$(find . -type f -not -path './.git/*' -executable)\" == ./tests/laravel-test.sh"
```
mfn, line 40: > `"test \"`

Super example, to me, to not have forced to require to add `"` suddenly everywhere?

Aka nothing was broken and now we've to escape what wasn't necessary before. My 2c :)

szepeviktor: It was not proper YAML, where YAML strings are `"quotes"`.

szepeviktor: Please take a look at e.g. https://github.com/ergebnis/php-library-template/blob/main/.github/workflows/integrate.yaml _the mother of all workflows_.

szepeviktor: https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.quoted_strings

> quoted-strings
> 
> Use this rule to forbid any string values that are not quoted, or to prevent quoted strings without needing it. You can also enforce the type of the quote used.

If you go the "prevent quoted strings without needing" way you always have to think and decide: Are quotes needed? Are quotes needed? Are quotes needed?
If you always quote string it is simply a _"yes"_.

mfn: I guess I'm just pragmatic 🤷‍♀️

PR #725, opened 2020-11-28
Add support for autoloaded arrays

### LINKED ISSUE #724: Doesn't work when autoloading multiple directories into a single namespace
- Larastan Version: Latest
- `--level` used: any

### Description

When you use:
```json
"autoload-dev": {
    "psr-4": {
        "Lean\\Tests\\": ["tests/"]
    }
},
```

Larastan will try to get all files in that namespace.

However, it won't loop through the `["tests/"]` array, so it will pass `["tests/"]` as a path to `ClassMapGenerator::createMap()`, when the method expects a string.

Screen Shot 2020-11-28 at 17 24 10

This results in:
```
Error thrown in /Users/samuel/Projects/lean/lean/vendor/composer/composer/src/Composer/Autoload/ClassMapGenerator.php on line 86 while loading bootstrap file /Users/samuel/Projects/lean/lean/vendor/phpstan/phpstan/../../nunomaduro/larastan/bootstrap.php: Call to a member function getPathname() on string
```

### Code where the issue was found

ApplicationResolver line ~103

### Solution

Loop through the paths, not just through the namespaces.

stancl:
Resolves #724

### DIFF
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -102,7 +102,7 @@ final class ApplicationResolver
         $autoloadDevPsr4 = $autoloadDev['psr-4'] ?? [];
         foreach ($autoloadDevPsr4 as $paths) {
             $paths = is_array($paths) ? $paths : [$paths];
-            
+
             foreach ($paths as $path) {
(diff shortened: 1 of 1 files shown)


nunomaduro:
Thank you.

PR #751, opened 2021-01-05
fix replacing collections in return types

### LINKED ISSUE #747: Static method on model has return type signature replaced incorrectly when containing a collection
- Larastan Version: v0.6.11
- `--level` used: 6

### Description
Defining a static method on a `Model` that contains a reference to an Eloquent `Collection` in its return type signature results in a completely different return type being parsed.

I have a legacy method that has a slightly awkward return type that can provide either a collection of a specific model, or just a single model if only one is available:
```php
use Illuminate\Database\Eloquent\Collection as DatabaseCollection;

class User extends Model
{
    /**
     * @return DatabaseCollection<Order>|Order
     */
    public static function testFunctionWithCollectionInReturnType()
    {
        return new DatabaseCollection([]);
    }
}

$orders = User::testFunctionWithCollectionInReturnType();
foreach ($orders as $order) {
    $order->some_property_belonging_to_orders;
}
```

It seems that the return type is parsed as `DatabaseCollection<User>` which loses the union type and the intended type.
As such an access to an undefined property is raised.

I'm not familiar with the internals of phpstan/larastan or I would also work on a PR to fix, but if someone can point me in the right direction, I am happy to take a stab at it. After a little debugging, it seems the return type signature is replaced here:
https://github.com/nunomaduro/larastan/blob/a036948cdcb09bc374f8be77db4c1262374cb297/src/ReturnTypes/ModelExtension.php#L89-L93

Josh-G:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

resolves #747 

**Changes**
This PR restricts the methods that `ModelExtension` will change the return type. Instead of replacing any return type containing a collection, it will now only replace the return type of the methods that can return a custom collection.
Additionally, `GenericEloquentCollectionTypeNodeResolverExtension` would inadvertently resolve `Collection|static[]` to a class with the name `static`. I have updated this to resolve the inner array type, such that static is now correctly interpreted as a `StaticType`.

This is the first time I've looked into the internals, please let me know if this is not the correct approach (apologies if so).

The test failure is unrelated and is currently affecting master, the tests [passed on my fork](https://github.com/Josh-G/larastan/actions/runs/463496559) with the unrelated failure ignored.

**Breaking changes**
I believe this was the intended behaviour.


szepeviktor (approved):

szepeviktor:
Please merge master to get #775

PR #778, opened 2021-02-03
Align whereIntegerInRaw type annotations with Eloquent.

MrMage:
See https://github.com/illuminate/database/blob/master/Query/Builder.php. This seems to be producing false positives for me; the key type in particular is irrelevant here.

### DIFF
--- a/stubs/QueryBuilder.stub
+++ b/stubs/QueryBuilder.stub
@@ -394,7 +394,7 @@ class Builder
      * Add a "where in raw" clause for integer values to the query.
      *
      * @param  string  $column
-     * @param  \Illuminate\Contracts\Support\Arrayable|array  $values
+     * @param  \Illuminate\Contracts\Support\Arrayable|array<int|string, mixed>  $values
      * @param  string  $boolean
      * @param  bool  $not
      * @return $this
@@ -406,7 +406,7 @@ class Builder
      * Add an "or where in raw" clause for integer values to the query.
      *
      * @param  string  $column
-     * @param  \Illuminate\Contracts\Support\Arrayable|array  $values
+     * @param  \Illuminate\Contracts\Support\Arrayable|array<int|string, mixed>  $values
      * @return $this
      */
     public function orWhereIntegerInRaw($column, $values)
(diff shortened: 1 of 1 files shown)


szepeviktor:
> the key type in particular is irrelevant here.

What do you think the key and value types are?

szepeviktor:
AFAIK

key -> [PHP docs](https://www.php.net/manual/en/language.types.array.php#language.types.array.syntax) say _The key can either be an int or a string._
value -> mixed is the broadest possible

MrMage:
Yes, but the previous definition of the method only allowed for string keys. This excludes integer keys.

stubs/QueryBuilder.stub
```diff
@@ -394,7 +394,7 @@ class Builder
      * Add a "where in raw" clause for integer values to the query.
      *
      * @param  string  $column
-     * @param  \Illuminate\Contracts\Support\Arrayable|array<string, mixed>  $values
+     * @param  \Illuminate\Contracts\Support\Arrayable|array  $values
```
szepeviktor, line 397:
```suggestion
     * @param  \Illuminate\Contracts\Support\Arrayable|array<int|string, mixed>  $values
```

stubs/QueryBuilder.stub
```diff
@@ -406,7 +406,7 @@ class Builder
      * Add an "or where in raw" clause for integer values to the query.
      *
      * @param  string  $column
-     * @param  \Illuminate\Contracts\Support\Arrayable|array<string, mixed>  $values
+     * @param  \Illuminate\Contracts\Support\Arrayable|array  $values
```
szepeviktor, line 409:
```suggestion
     * @param  \Illuminate\Contracts\Support\Arrayable|array<int|string, mixed>  $values
```

MrMage:
What's the point of providing `array<int|string, mixed>` if that is functionally equivalent to `array`? Also, the Laravel source that this file is based on uses `array` as well.

szepeviktor:
On PHPStan Level 6 you have to declare what is in your array.

MrMage:
Thanks, applied your fixes now.

szepeviktor (approved):

szepeviktor:
Thank you for fixing those!

PR #793, opened 2021-03-09
Add #751 to changelog

szepeviktor:
Is it all right? @Josh-G


Josh-G:
thanks @szepeviktor, may I suggest moving it to a Fixed section and possibly this as the change message, it's a little clearer I think?
> Fixed incorrectly inferring the return type of custom methods that return collections ([#751](https://github.com/nunomaduro/larastan/pull/751)) Thanks @Josh-G

szepeviktor:
Thank you!
Without _little_.

Issue #796, opened 2021-03-19
PHPStan 0.12.82 compatibility


canvural:
In [PHPStan v0.12.82](https://github.com/phpstan/phpstan/releases/tag/0.12.82) release there were changes related to how `static` keyword is resolved. This caused some issues in Larastan. Need to investigate and fix.

szepeviktor:
Was it this PR? https://github.com/phpstan/phpstan-src/pull/474

canvural:
> Was it this PR? [phpstan/phpstan-src#474](https://github.com/phpstan/phpstan-src/pull/474)

Probably.

shaffe-fr:
Most of the errors in my project can be summed up to the following snippet:
```php
$this->userOrNull = $office->user()->active()->first();
```

* Model scopes Builder parameter is not ignored
`Method Illuminate\Database\Query\Builder::active()  invoked with 0 parameters, at least 1 required.`
* Model type is lost on relations Builder: `Property \MyClass::$userOrNull (App\User|null) does not accept Illuminate\Database\Eloquent\Model|null.`

Hope it helps.

canvural:
@shaffe-fr Thank you for the information!

I also changed the `master` branch so we can see which tests are failing. [Here](https://github.com/nunomaduro/larastan/runs/2201088920?check_suite_focus=true) is the failing tests from the actions.

canvural:
I spent a little bit of time on this. But couldn't come up with a good solution. So if anyone else is interested, we can collaborate on a solution!

szepeviktor:
Should I dare to mention Ondřej Mirtes? What if he will points out bad design in Larastan??

szepeviktor:
First one being

```
Method Tests\\Features\\Methods\\ModelExtension::testLockForUpdate() should return
Illuminate\\Database\\Eloquent\\Builder<App\\User> but returns
Illuminate\\Database\\Eloquent\\Builder<Illuminate\\Database\\Eloquent\\Model>.
```

in

https://github.com/nunomaduro/larastan/blob/f74befd87b7278285cbf6f1597509cbfaca46bf8/tests/Features/Methods/ModelExtension.php#L284-L288

ondrejmirtes:
Don't hesitate to reach out to me as I'm probably the only person that can help you here :) it would be great if you could isolate what Larastan extensions started behaving differently between PHPStan 0.12.81 and 0.12.82 then we can move on to a solution.

[...]

PR #798, opened 2021-03-19
Specify version constrain for problematic PHPStan

szepeviktor:
https://getcomposer.org/doc/articles/versions.md#version-range

- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

More details in  #796

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


canvural:
I don't see why this is needed. Nothing wrong with locking the dependency to a specific version.

szepeviktor:
Okay.

szepeviktor:
> "phpstan/phpstan": "0.12.81",

Does not tell the user what is going on.

> "phpstan/phpstan": "^0.12.70 <0.12.82",

Informs the user that anything above 70 is okay, but from 82 on there is a problem.

PR #807, opened 2021-04-09
Fix when and unless return value

netpok:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**
This PR fixes the return value validation for the when and unless methods on the Models

I'm also not sure about this whole feature as it does what the `when` method should do but not what it actually can.

A more realistic (and invalid) version:
```php
    /**
     * Apply the callback's query changes if the given "value" is true.
     *
     * @template TValue
     * @template TCallbackResult
     * @template TDefaultResult
     *
     * @param TValue  $value
     * @param callable($this, TValue): TCallbackResult $callback
     * @param callable($this, TValue): TDefaultResult|null  $default
     * @return TCallbackResult|TDefaultResult|$this
     */
    public function when($value, $callback, $default = null);
```

Or a valid version: A more realistic (and invalid) version:
```php
    /**
     * Apply the callback's query changes if the given "value" is true.
     *
     * @param mixed  $value
     * @param callable($this, mixed): mixed $callback
     * @param callable($this, mixed): mixed $default
     * @return mixed|$this
     */
    public function when($value, $callback, $default = null);
```

**Breaking changes**
None

### DIFF
--- a/stubs/EloquentBuilder.stub
+++ b/stubs/EloquentBuilder.stub
@@ -189,8 +189,8 @@ class Builder
      * Apply the callback's query changes if the given "value" is true.
      *
      * @param mixed  $value
-     * @param callable($this, mixed): null|Builder<TModelClass> $callback
-     * @param callable($this, mixed): null|Builder<TModelClass>|null  $default
+     * @param callable($this, mixed): (null|Builder<TModelClass>) $callback
+     * @param callable($this, mixed): (null|Builder<TModelClass>)|null  $default
      * @return mixed|$this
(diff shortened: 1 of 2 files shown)


szepeviktor (approved):

szepeviktor:
Could you add a test for each method?

netpok:
@szepeviktor Is this test acceptable?

szepeviktor:
It seems so.

canvural:
Thank you!

PR #809, opened 2021-04-10
feat: Octane compatibility checking rule

### LINKED ISSUE #802: Rule for checking container bindings
With the Octane release, there was a [suggestion](https://twitter.com/OliverNybroe/status/1379402172253757441) from @olivernybroe if we can check the container bindings for Octane compatibility.

In this issue, I want to discuss and gather some information about what can we do in Larastan.

## Container bindings

From the Octane [docs](https://github.com/laravel/octane#container-injection):
> In general, you should avoid injecting the application service container or HTTP request instance into the constructors of other objects.

And the recommended solution is to either don't use singleton binding or pass a closure instead of the app instance. The same thing applies to [request](https://github.com/laravel/octane#request-injection) and [config](https://github.com/laravel/octane#configuration-repository-injection) injections.

This can be a rule I think. The hardest thing is here to avoid false positives. So we need a lot of examples that we can test on.

**disclaimer**: I'm not entriley sure if this rule can be written. Just brainstorming at this point.

canvural:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Resolves #802

**Changes**

This PR adds an optional rule that can check for [Octane](https://github.com/laravel/octane) compatibility.

It's an optional rule. You need to add `checkOctaneCompatibility: true` to your `parameters` array in the config file.

**Breaking changes**
none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [0.7.3] - 2021-04-12
+
+### Added
+
+- Octane compatibility checking rule ([#809](https://github.com/nunomaduro/larastan/pull/809))
+
 ## [0.7.2] - 2021-04-08
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


nunomaduro (approved):
🔥🔥🔥

olivernybroe:
Would it be worth maybe just having it as default enabled? 
The practices for Octane is in general good practices, as they might have unforeseen issues in queue also right?

canvural:
@olivernybroe I think it should be optional because this rule also complains about this code:
```php
$this->app->singleton(Service::class, function ($app) {
    return new Service($app);
});
```

which is totally valid if you don't use Octane.

olivernybroe:
@canvural Awesome! Well I guess this is good to go?

canvural:
Released: https://github.com/nunomaduro/larastan/releases/tag/v0.7.3

And [tweeted.](
https://twitter.com/can__vural/status/1381565470428499971)

PR #811, opened 2021-04-16
When and unless don't require return type

### LINKED ISSUE #799: When & Unless Validations
- Larastan Version: 0.7.1
- `--level` used: 5

### Description

```
Parameter #2 $callback of method Illuminate\Database\Eloquent\Builder<App\Resource>::when() expects
(callable(Illuminate\Database\Eloquent\Builder<App\Resource>, mixed): null)|Illuminate\Database\Eloquent\Builder<App\Resource>,
Closure(Illuminate\Database\Eloquent\Builder): void given.
```

```php
Resource::query()
    ->when($request->filled('stuff'), function ($query) use ($request) {
        $query->where('stuff', $request->input('stuff'));
    });
```

I believe instead of `null` return we should also return `void`
https://github.com/nunomaduro/larastan/commit/3b32c74a63ce36f85d5b0434725a59e0dcd74292#diff-cda4a8378b7ebd0f5aa8a5a9ae53a08e48d436e67928064bf28963f969dd93e0R192

BertvanHoekelen:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Resolves #799

### DIFF
--- a/stubs/EloquentBuilder.stub
+++ b/stubs/EloquentBuilder.stub
@@ -189,7 +189,7 @@ class Builder
      * Apply the callback's query changes if the given "value" is true.
      *
      * @param mixed  $value
-     * @param callable($this, mixed): (null|Builder<TModelClass>) $callback
+     * @param callable($this, mixed): (void|Builder<TModelClass>) $callback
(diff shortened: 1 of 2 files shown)


canvural:
Thank you!

PR #820, opened 2021-04-22
Add sole support

ThibaudDauce:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Add support for `sole` method 

**Note**

I couldn't put the correct doc blocks with the throws:

```
    /**
     * Execute the query and get the first result if it's the sole matching record.
     *
     * @param  array|string  $columns
     * @return \Illuminate\Database\Eloquent\Model|object|static|null
     *
     * @throws \Illuminate\Database\RecordsNotFoundException
     * @throws \Illuminate\Database\MultipleRecordsFoundException
     */
```

because PHPStan raises an error "PHPDoc tag @throws with type Illuminate\\Database\\MultipleRecordsFoundException|Illuminate\\Database\\RecordsNotFoundException is not subtype of Throwable". I think it's because these exceptions only exists on a newer version of Laravel?

### DIFF
--- a/tests/FeaturesTest.php
+++ b/tests/FeaturesTest.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Tests;
 
+use Illuminate\Support\Str;
 use Symfony\Component\Finder\Finder;
 
 class FeaturesTest extends TestCase
@@ -37,6 +38,10 @@ class FeaturesTest extends TestCase
 
     private function analyze(string $file): int
     {
+        if (Str::contains($file, 'Features/Laravel8') && version_compare(app()->version(), '8.0.0', '<')) {
+            return 0;
+        }
+
         $result = $this->execLarastan($file);
 
         if (! $result || $result['totals']['errors'] > 0 || $result['totals']['file_errors'] > 0) {


ThibaudDauce:
The `sole` method is only available for Laravel 8.23 or upper. Is there a way to test this feature only for these versions?

szepeviktor:
> The `sole` method is only available for Laravel 8.23 or upper. Is there a way to test this feature only for these versions?

Larastan has no version dependent conditions. So just leave it there.

canvural:
Hi,

Currently, we don't support running tests for different Laravel versions.

If you want you can make seperate PR for that. I think https://github.com/nunomaduro/larastan/blob/master/tests/FeaturesTest.php needs to be refactored. It should get the Laravel version from some env variable. Github Actions can set that I guess. And inside the https://github.com/nunomaduro/larastan/tree/master/tests/Features directory we can have new folder for example `Laravel8`. And https://github.com/nunomaduro/larastan/blob/master/tests/FeaturesTest.php will scan that directory only if env variable is also Laravel8.

szepeviktor:
I see! CI fails.

stubs/EloquentBuilder.stub
```diff
@@ -145,6 +145,14 @@ class Builder
      */
     public function firstOrFail($columns = ['*']);
 
+    /**
+     * Execute the query and get the first result if it's the sole matching record.
+     *
+     * @param  array<int, (model-property<TModelClass>|'*')>|model-property<TModelClass>|'*'  $columns
+     * @phpstan-return TModelClass
+     */
```
canvural, line 153: Now that the other PR is merged, can you add back the `@throw` tags again here?

canvural (changes requested):
Looks good! Apart from the one review comment please squash your commits and rebase on top of master again. Then we can merge!

stubs/EloquentBuilder.stub
ThibaudDauce: Not sure that I can but I will try…

ThibaudDauce:
Do you want me to rebase the branch from the upstream master?

canvural:
Yes, after you add the `throws` tags.

szepeviktor:
What happened to you @ThibaudDauce?

stubs/EloquentBuilder.stub
canvural: Why not?

ThibaudDauce:
I think I push an empty branch on the PR and Github closed it :-)

ThibaudDauce:
Reopen https://github.com/nunomaduro/larastan/pull/822/files

PR #821, opened 2021-04-22
feat: add support for Laravel 8 and upper tests

ThibaudDauce:
**Changes**

The new `tests/Features/Laravel8` folder can contains tests only executed on Laravel 8 and above.

Needed to add the `sole` method support https://github.com/nunomaduro/larastan/pull/820#issuecomment-824741968


tests/FeaturesTest.php
```diff
@@ -37,6 +38,12 @@ public function testFeatures(string $file): void
 
     private function analyze(string $file): int
     {
+        $laravelMajorVersion = (int) explode('.', app()->version())[0];
+
+        if (Str::contains($file, 'Features/Laravel8') && $laravelMajorVersion < 8) {
```
szepeviktor, line 43: Please consider using `compare_version()` with < 8.0.0

canvural: I agree with @szepeviktor

tests/FeaturesTest.php
```diff
@@ -37,6 +38,12 @@ public function testFeatures(string $file): void
 
     private function analyze(string $file): int
     {
+        $laravelMajorVersion = (int) explode('.', app()->version())[0];
```
canvural: I would rather have these checks in `getFeatures` method.

canvural (changes requested):

tests/FeaturesTest.php
ThibaudDauce: Nice! Didn't know this PHP function :-D

tests/FeaturesTest.php
ThibaudDauce: It's not possible because the framework is not initialize yet in a data provider (I think it's the reason)

canvural: Ok, good point! So let's leave it for now :+1:

tests/FeaturesTest.php
```diff
@@ -37,6 +38,10 @@ public function testFeatures(string $file): void
 
     private function analyze(string $file): int
     {
+        if (Str::contains($file, 'Features/Laravel8') && version_compare(app()->version(), '8.0.0') < 0) {
```
canvural, line 41: I think you can do `version_compare(app()->version(), '8.0.0', '<')` here

ThibaudDauce: Yes but I thought it was less readable but as you wish

canvural: IMO, `version_compare(...) < 0` is more confusing. Why do we expect it to be less than 0? (If you don't know the return types of version_compare) :smile:

canvural (approved):
Looks good! When the checks are passing you can remove the dummy test, and push again. Then we can merge! :+1:

canvural:
Thank you!

tests/FeaturesTest.php
szepeviktor: Most parts of PHP is confusing!

Issue #837, opened 2021-05-27
Nested SoftDelete not detected since 0.7.5 


IAmRGroot:
- Larastan Version: ^0.7.5
- `--level` used: 8
- Pull request with failing test: #838 

### Description
Larastan does not detect a nested SoftDeletes trait after updating from 0.7.4 to 0.7.5 or higher. 

### Laravel code where the issue was found
```php
class User extends Authenticatable
{
    public function group(): BelongsTo
    {
        return $this->belongsTo(Group::class)->withTrashed(); // Call to an undefined method Illuminate\\Database\\Eloquent\\Relations\\BelongsTo<App\\Group, App\\User>::withoutTrashed().
    }
}

class Group extends Model
{
    use NestedSoftDeletes;
}

trait NestedSoftDeletes
{
    use SoftDeletes;
}
```

canvural:
Hi,

We are using the PHPStan's MethodReflection method `hasTraitUse` [here](https://github.com/nunomaduro/larastan/blob/master/src/Methods/EloquentBuilderForwardsCallsExtension.php#L103-L106) to determine if a class uses the given trait. Looks like this needs to recursive. But I don't know it's possible.

Please feel free to create a PR that fixes it in the code I pointed out. Also you can ask in PHPStan if you need help with how to solve this.

szepeviktor:
@ondrejmirtes Could you help us whether `hasTraitUse` is recursive?

ondrejmirtes:
Try using `ClassReflection::getTraits()` instead, or send a fix to PHPStan :)

IAmRGroot:
@canvural I've added a function in #838 that collects nested traits as the PHPStan's `collectTraits` function is private. Let me know if this is acceptable.

canvural:
@IAmRGroot Did you try `ClassReflection::getTraits()` first as @ondrejmirtes suggested?

IAmRGroot:
@canvural Yes!

`ClassReflection::getTraits` gives  `App\Group: {"App\\Traits\\NestedSoftDeletes":{}}`

(new) `collectTraitNames` gives `App\Group: ["App\\Traits\\NestedSoftDeletes","Illuminate\\Database\\Eloquent\\SoftDeletes"]`

[...]

PR #838, opened 2021-05-27
feat: adds support for detecting nested trait usage for SoftDeletes trait

IAmRGroot:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Related to #837 

**Changes**

- Use recursive `$modelReflection->getTraits(true)` instead of `$modelReflection->hasTraitUse()`

Tests:
- Added `NestedSoftDeletes` trait
- Groups model now uses `NestedSoftDeletes` instead of `SoftDeletes`
- Added `->withThrashed` to the Users Group relation to add a test case

**Breaking changes**

None

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     ],
     "require": {
         "php": "^7.2 || ^8.0",
-        "ext-json": "^8.0",
+        "ext-json": "*",
         "composer/composer": "^1.0 || ^2.0",
         "illuminate/console": "^6.0 || ^7.0 || ^8.0 || ^9.0",
         "illuminate/container": "^6.0 || ^7.0 || ^8.0 || ^9.0",
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,7 @@
     ],
     "require": {
         "php": "^7.2 || ^8.0",
+        "ext-json": "^8.0",
         "composer/composer": "^1.0 || ^2.0",
         "illuminate/console": "^6.0 || ^7.0 || ^8.0 || ^9.0",
         "illuminate/container": "^6.0 || ^7.0 || ^8.0 || ^9.0",
@@ -21,9 +22,8 @@
         "illuminate/pipeline": "^6.0 || ^7.0 || ^8.0 || ^9.0",
         "illuminate/support": "^6.0 || ^7.0 || ^8.0 || ^9.0",
         "mockery/mockery": "^0.9 || ^1.0",
-        "phpstan/phpstan": "^0.12.83",
-        "symfony/process": "^4.3 || ^5.0",
-        "ext-json": "*"
+        "phpstan/phpstan": "^0.12.90",
+        "symfony/process": "^4.3 || ^5.0"
     },
     "require-dev": {
         "orchestra/testbench": "^4.0 || ^5.0 || ^6.0 || ^7.0",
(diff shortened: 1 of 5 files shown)


szepeviktor:
Spotted a `rgroot`email in your commit.

szepeviktor:
I think Larastan's SoftDeletes detection needs to be fixed.

IAmRGroot:
@szepeviktor thanks, it should be fixed now!

IAmRGroot:
@szepeviktor please let me know if you need anything from my end

szepeviktor (approved):

IAmRGroot:
@szepeviktor I've changed it to use the new phpstan recursive getTraits function. Please let me know if I need to change anything

composer.json
```diff
@@ -12,6 +12,7 @@
     ],
     "require": {
         "php": "^7.2 || ^8.0",
+        "ext-json": "^8.0",
```
canvural, line 15: Can you revert this change? As you can see from the GA, it's causing composer to fail.

szepeviktor: It could be `"ext-json": "*",`

szepeviktor (changes requested):

composer.json
canvural: It was like that already.

szepeviktor: https://github.com/nunomaduro/larastan/blob/master/composer.json#L26

😲

IAmRGroot: Ah yes I thought it only changed the order. I'll revert it

canvural:
Thank you!

tests/Application/app/User.php
```diff
@@ -63,7 +63,7 @@ public function scopeActive(Builder $query): Builder
     /** @phpstan-return BelongsTo<Group, User> */
     public function group(): BelongsTo
     {
-        return $this->belongsTo(Group::class);
+        return $this->belongsTo(Group::class)->withTrashed();
```
nunomaduro, line 66: I feel that you should add a new test, and not modify an existing one. cc @canvural

nunomaduro: Ups, looks like @canvural just merged this. Can you confirm that modifying an existing test is normal?

canvural: Actually this not a test case. It's just the dummy application that the tests are using. 

So now I think of it, this actually misses a test case 😅 Thank you @nunomaduro 

@IAmRGroot Can you add a new method to https://github.com/nunomaduro/larastan/blob/master/tests/Features/Models/Relations.php called `testNestedSoftDeletesTrait` maybe, that uses the `group` method on `User` class. And return type should be `BelongsTo<Group, User>` You can check how other methods are written in that file.

szepeviktor: What a turmoil!

IAmRGroot: I did initially change the test application to trigger a check for the nested SoftDeletes trait.

Maybe in hindsight this was not needed as the existing tests `testWithTrashedWithBelongsToRelation`, `testOnlyTrashedWithBelongsToRelation`, `testWithoutTrashedWithBelongsToRelation` already check the `$user->group()->withTrashed()` function:
https://github.com/nunomaduro/larastan/blob/master/tests/Features/Models/Relations.php#L185

PR #850, opened 2021-06-24
feat: add support for vendor-dir

rajyan:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**
Changed `$rootDir` to `$vendorDir` and use `self::$composer['config']['vendor-dir']` if exists in `ApplicationResolver.php`.
This PR adds support for https://getcomposer.org/doc/06-config.md#vendor-dir.

Without this PR I encountered an error, because we have our app files in laravel/app and the vendor dir is laravel/vendor.
Our Laravel project has a bootstrap file, but couldn't read it for the same reason. (I know it's a rare case, but I believe this PR helps other people using "vendor-dir")
```
ErrorException thrown in /var/source/app/laravel/vendor/nunomaduro/larastan/src/ApplicationResolver.php on line 134 while loading bootstrap file /var/source/app/laravel/vendor/nunomaduro/larastan/bootstrap.php: include(/var/source/app/vendor/composer/autoload_psr4.php): failed to open stream: No such file or directory
```

**Breaking changes**

This won't break any existing features, because it would fallback to the general behaviour if the `self::$composer['config']['vendor-dir']` doesn't exist.

### DIFF
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -35,7 +35,8 @@ final class ApplicationResolver
         if (file_exists($composerFile)) {
             self::$composer = json_decode((string) file_get_contents($composerFile), true);
             $namespace = (string) key(self::$composer['autoload']['psr-4']);
-            $serviceProviders = array_values(array_filter(self::getProjectClasses($namespace, dirname($composerFile)), function (string $class) use (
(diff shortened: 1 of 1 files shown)


szepeviktor (approved):

canvural:
Thank you!

PR #851, opened 2021-06-27
change bootstrap path

rajyan:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Changes `getcwd()` to `__DIR__.'/../../../'` in `bootstrap.php`

In Laravel, [basePath is configureable](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Application.php#L170), so `getcwd()` might not be the basePath.

[bootstrapPath](https://github.com/laravel/framework/blob/277c2fbd0cebd2cb194807654d870f4040e288c0/src/Illuminate/Foundation/Application.php#L368-L371) and [vendorPath](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/PackageManifest.php#L58) are hardcoded and cannot be changed, so I think it is better to find Laravel bootstrap file by relative path.

**Breaking changes**

I believe there are no breaking changes, but not sure enough.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -6,7 +6,7 @@ use NunoMaduro\Larastan\ApplicationResolver;
 
 define('LARAVEL_START', microtime(true));
 
-if (file_exists($applicationPath = getcwd().'/bootstrap/app.php')) { // Applications and Local Dev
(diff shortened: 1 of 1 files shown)


rajyan:
just changing the working dir when running phpstan would be easier. Sorry

PR #853, opened 2021-07-01
Recommend usage of composer update in CONTRIBUTING.md

spawnia:
Given there is no composer.lock file and the pipeline also runs the latest
dependencies, it seems better to have the local dependencies up to date.


canvural:
Thank you!

Issue #856, opened 2021-07-01
Fail to bootstrap application with bamarni/composer-bin-plugin


bastien-phi:
- Larastan Version: 0.7.7
- `--level` used: 8
- Pull request with failing test:

### Description

Phpstan analysis fails  with 
```
Trait "Orchestra\Testbench\Concerns\CreatesApplication" not found 
in vendor-bin/phpstan/vendor/nunomaduro/larastan/src/ApplicationResolver.php on line 16
```

Problem was introduced in https://github.com/nunomaduro/larastan/pull/852

My setup is a classic laravel application with vendor-bin composer plugin and larastan installed via vendor-bin

In fact, in `bootstrap.php`, 
```php
if (file_exists($applicationPath = dirname(__DIR__, 3) . '/bootstrap/app.php')) { // Applications and Local Dev
    $app = require $applicationPath;
} else { // Packages
    $app = ApplicationResolver::resolve();
}
```
the app is tried to be resolved from ApplicationResolver::resolve because `$applicationPath` does not exist.

The `bootstrap/app.php` does exist but not at the right place,  relatively to `bootstrap.php`.

The folders look like this :

- app
- bootstrap
  - app.php
- ...
- vendor
- vendor-bin
  - phpstan
    - vendor
      - nunomaduro
        - larastan
          - bootstrap.php
          
`dirname(__DIR__, 3)` does not give project root here

szepeviktor:
Do you have `bamarni/composer-bin-plugin` installed?

bastien-phi:
Yes

Le jeu. 1 juil. 2021 à 17:59, Viktor Szépe ***@***.***> a
écrit :

szepeviktor:
Why do you have Testbench installed?

bastien-phi:
I don't. That's why the trait `Orchestra\Testbench\Concerns\CreatesApplication` is not found :sweat_smile: 

I think something like this should do fix the issue
```
if (file_exists($applicationPath = dirname(__DIR__, 3) . '/bootstrap/app.php')) { // Applications and Local Dev
    $app = require $applicationPath;
}
elseif (file_exists($applicationPath = dirname(__DIR__, 5) . '/bootstrap/app.php')) { // larastan installed via bamarni/composer-bin-plugin
    $app = require $applicationPath;
} else { // Packages
    $app = ApplicationResolver::resolve();
}
```

rajyan:
Sorry I wasn’t aware of this kind of use case.
[...]

PR #852, opened 2021-06-30
fix bootstrap path

rajyan:
- [x] Added or updated tests
- [ ] Documented user facing changes
- [ ] Updated CHANGELOG.md

**Changes**
I want to re-open this PR https://github.com/nunomaduro/larastan/pull/851.
This PR would enable running larastan without caring about the working directory.
As I mentioned in https://github.com/nunomaduro/larastan/pull/851, Laravel's bootstrap and vendor path are hardcoded, and should always be a relative path `__DIR__.'/../../../bootstrap/app.php'` from `vendor/nunomaduro/larastan`.

I modified the test a little, and added a test to confirm the changes.

**Breaking changes**

I think there are no breaking changes.

PR #857, opened 2021-07-01
fix: leave getcwd() way of finding bootstrap file

rajyan:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

fixes https://github.com/nunomaduro/larastan/issues/856
Sorry, I wasn't aware of symlink usages.
I think it is better to restore 
```php
file_exists($applicationPath = getcwd().'/bootstrap/app.php')
```
than adding a new relative path
```php
file_exists($applicationPath = dirname(__DIR__, 5).'/bootstrap/app.php')
```
because there might be other symlink usages.

**Breaking changes**

This PR would fix the breaking changes from https://github.com/nunomaduro/larastan/pull/852.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -8,8 +8,6 @@ define('LARAVEL_START', microtime(true));
 
 if (file_exists($applicationPath = dirname(__DIR__, 3).'/bootstrap/app.php')) { // Applications and Local Dev
     $app = require $applicationPath;
-} elseif (file_exists($applicationPath = getcwd().'/bootstrap/app.php')) {
-    $app = require $applicationPath;
 } else { // Packages
     $app = ApplicationResolver::resolve();
 }
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -8,6 +8,8 @@ define('LARAVEL_START', microtime(true));
 
 if (file_exists($applicationPath = dirname(__DIR__, 3).'/bootstrap/app.php')) { // Applications and Local Dev
     $app = require $applicationPath;
+} elseif (file_exists($applicationPath = getcwd().'/bootstrap/app.php')) {
+    $app = require $applicationPath;
 } else { // Packages
     $app = ApplicationResolver::resolve();
 }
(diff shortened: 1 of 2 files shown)


szepeviktor:
To me it seems like a point-and-shoot game.

Is there an **ultimate** solution to find Laravel app's root?
(maybe there isn't)

szepeviktor:
Larastan's path or or the current working directory seems to be unrelated to where the Laravel app is.

rajyan:
> To me it seems like a point-and-shoot game.

I agree with you.
Leaving 
```php
file_exists($applicationPath = dirname(__DIR__, 3).'/bootstrap/app.php'
```
as default (because it's the default Laravel app configuration), adding something like
https://phpstan.org/config-reference#bootstrap
to make it configurable by config file might be better? I'm not sure if it's possible.

szepeviktor:
> to make it configurable

# ❤️

szepeviktor:
So we could revert #857 and introduce a configuration option.

rajyan:
I found a way to make it configurable, but when resolving parameters from reference like "%laravelBootstrapFile%", 

https://phpstan.org/config-reference#bootstrap
> Relative paths in the bootstrapFiles key are resolved based on the directory of the config file is in.

this feature seems to be broken?

We have to set a relative path from phpstan.neon as a default, which seems to be impossible.
I'll create an issue in phpstan.

szepeviktor:
> We have to set a relative path from phpstan.neon as a default

No. When `laravelBootstrapFile` is not define we should use `getcwd()`

szepeviktor:
There is a misunderstanding here!
This PR modifies our https://github.com/nunomaduro/larastan/blob/master/bootstrap.php file's path
















.

rajyan:
https://github.com/nunomaduro/larastan/pull/857#issuecomment-872674695

Did you mean to revert this one https://github.com/nunomaduro/larastan/pull/852 ?

szepeviktor:
> Did you mean to revert this one #852 ?

Yes.

rajyan:
Ok, then I'll create a revert PR and configuration PR separately 👍

PR #861, opened 2021-07-02
Laravel 8 factory support

### LINKED ISSUE #661: Support Laravel 8 factories
- Larastan Version: 0.6.4
- `--level` used: max

### Description
It's seems that states on the new factories that came with Laravel 8 are not yet supported.

It would be nice if we could deduct the return type of the `Model::factory()` method, added by the [HasFactory](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Eloquent/Factories/HasFactory.php) trait.

```php
User::factory()->expired()->create();
```

I would love to help out, but I haven't got a clue where to start. If anyone could guide me, I'm very much willing to put some time and effort into this.

canvural:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Closes #661 

This PR adds support for Laravel 8 factories.

- `User::factory()` would return `UserFactory`
- `User::factory()->createOne()` would return `User`
- Unfortunately there is no good solution to infer more precise return type for `create` method. It can either return collection of models, or single model. And this is decided by the `count()` method. This kind of fluent calls are hard to keep track with PHPStan.

I will add support for relations later (`has{Relation}` and `for{Relation}` method calls), I'm opening the PR now so if anyone wants to try over the weekend they could do so. I'd appreciate any feedback if anyone tries.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,8 +34,10 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Tests\\": "tests/",
-            "App\\": "tests/Application/app"
+            "App\\": "tests/Application/app/",
+            "Laravel8\\": "tests/Laravel8Application/app/",
+            "Database\\Factories\\": "tests/Laravel8Application/database/factories/",
+            "Tests\\": "tests/"
         }
     },
     "minimum-stability": "dev",
--- a/extension.neon
+++ b/extension.neon
@@ -4,6 +4,7 @@ parameters:
         - stubs/EloquentBuilder.stub
         - stubs/Collection.stub
         - stubs/EloquentCollection.stub
+        - stubs/Factory.stub
         - stubs/Model.stub
         - stubs/JsonResource.stub
         - stubs/Gate.stub
@@ -104,6 +105,10 @@ services:
         class: NunoMaduro\Larastan\Methods\Extension
         tags:
             - phpstan.broker.methodsClassReflectionExtension
+    -
+        class: NunoMaduro\Larastan\Methods\ModelFactoryMethodsClassReflectionExtension
+        tags:
+            - phpstan.broker.methodsClassReflectionExtension
 
     -
         class: NunoMaduro\Larastan\Properties\ModelAccessorExtension
@@ -144,6 +149,11 @@ services:
         class: NunoMaduro\Larastan\Properties\ModelRelationsExtension
         tags:
             - phpstan.broker.propertiesClassReflectionExtension
+
+    -
+        class: NunoMaduro\Larastan\ReturnTypes\ModelFactoryDynamicStaticMethodReturnTypeExtension
+        tags:
+            - phpstan.broker.dynamicStaticMethodReturnTypeExtension
 
     -
         class: NunoMaduro\Larastan\ReturnTypes\ModelExtension
--- a/src/Methods/ModelFactoryMethodsClassReflectionExtension.php
+++ b/src/Methods/ModelFactoryMethodsClassReflectionExtension.php
@@ -1,0 +1,160 @@
+<?php
+
+declare(strict_types=1);
+
+namespace NunoMaduro\Larastan\Methods;
+
+use Illuminate\Database\Eloquent\Factories\Factory;
+use Illuminate\Support\Str;
+use PHPStan\Analyser\OutOfClassScope;
+use PHPStan\Reflection\ClassMemberReflection;
+use PHPStan\Reflection\ClassReflection;
+use PHPStan\Reflection\FunctionVariant;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Reflection\MethodsClassReflectionExtension;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\TrinaryLogic;
+use PHPStan\Type\Generic\TemplateTypeMap;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\Type;
+
+class ModelFactoryMethodsClassReflectionExtension implements MethodsClassReflectionExtension
+{
+    public function hasMethod(ClassReflection $classReflection, string $methodName): bool
+    {
+        // Class only available on Laravel 8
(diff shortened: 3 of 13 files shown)


composer.json
```diff
@@ -34,8 +34,10 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Tests\\": "tests/",
-            "App\\": "tests/Application/app"
+            "App\\": "tests/Application/app",
+            "Laravel8\\": "tests/Features/Laravel8/Application/app",
+            "Database\\Factories\\": "tests/Features/Laravel8/Application/database/factories",
+            "Tests\\": "tests/"
```
szepeviktor, line 40:
```suggestion
            "App\\": "tests/Application/app/",
            "Laravel8\\": "tests/Features/Laravel8/Application/app/",
            "Database\\Factories\\": "tests/Features/Laravel8/Application/database/factories/",
            "Tests\\": "tests/"
```

tests/phpstan-tests.neon
```diff
@@ -6,3 +6,4 @@ parameters:
         - '#Return typehint of method Tests\\Features\\ReturnTypes\\Facades::testHttpBaseUrl\(\) has invalid type Illuminate\\Http\\Client\\PendingRequest\.#'
         - '#Call to static method baseUrl\(\) on an unknown class Illuminate\\Support\\Facades\\Http\.#'
         - '#testDiskGetDriver\(\) has invalid type League\\Flysystem\\FilesystemInterface.#'
+        - '#Call to an undefined method Faker\\Generator::safeEmail()#'
```
szepeviktor, line 9:
```suggestion
        - '#Call to an undefined method Faker\\Generator::safeEmail\(\)#'
```

spawnia:
Thanks for getting this going, I tried it out in our project.

> User::factory() would return UserFactory

Works beautifully.

> User::factory()->createOne() would return User

Does not work for me yet.

canvural:
> Thanks for getting this going, I tried it out in our project.
> 
> > User::factory() would return UserFactory
> 
> Works beautifully.
> 
> > User::factory()->createOne() would return User
> 
> Does not work for me yet.

Yeah, I forgot to write it in the description. Now that the `Factory` class is generic you should add `@extends Factory<User>` to your `UserFactory` class docblock.

I'll write this in release notes/changelog or upgrade guide before merging.

spawnia:
> Now that the Factory class is generic you should add @extends Factory<User> to your UserFactory class docblock.

Sounds reasonable. Is there a way to infer this type when coming from a `MyModel::factory()` call? The information is there, can it be passed on to PHPStan?

canvural:
> > Now that the Factory class is generic you should add @extends Factory to your UserFactory class docblock.
> 
> Sounds reasonable. Is there a way to infer this type when coming from a `MyModel::factory()` call? The information is there, can it be passed on to PHPStan?

No. When you extend a generic class you only have 2 options. Either use `@extends` or make the class also generic.

szepeviktor:
Is there a way to run PHP 8 tests only on PHP 8?

spawnia:
It works fine when using `@extends`, thank you.

I thought it would have to work, given from what I have experienced with inferred types of `Collection`. I do see the problem now, when making custom collection subclasses they have to use `@extends` too, right?

I am not sure I agree with PHPStan here, if that is where the limitation is coming from. Creating a subclass of a generic class does not necessarily erase its generic-ness.

canvural:
@szepeviktor Do you mean Laravel 8? Because we don't have any PHP 8 tests. For Laravel 8 yeah we are trying to run it on just Laravel 8, but seems it does not work exactly the way we wanted.

@spawnia It does not eras the genericness. If a class has a `@template` you have pass something to there when you are extending.

szepeviktor:
> Do you mean Laravel 8

Yes.

spawnia:
> It does not eras the genericness. If a class has a @template you have pass something to there when you are extending.

My language was not that clear, and generic-ness is not a well-defined term. Having to provide an argument for the generic type parameter is what I deem erasure of the generic-ness. Any subclass extending from a generic class is no longer a generic class, unless the generic is redefined. I would think that when no type argument is provided, the generic keeps being generic.

szepeviktor (approved):

canvural:
I think this is ready. I'll test it in some projects of mine. And maybe clean it a little more. Than we can merge 👍🏽

tests/Features/Laravel8/Application/database/factories/UserFactory.php
```diff
@@ -0,0 +1,52 @@
...
+declare(strict_types=1);
+
+namespace Database\Factories;
+
+use Illuminate\Database\Eloquent\Factories\Factory;
+use Illuminate\Support\Str;
+use Laravel8\Models\User;
+
+/**
+ * @extends Factory<User>
```
nunomaduro, line 12: Will our users need this?

canvural: Yes. I'll document it in release notes or upgrade guide.

nunomaduro: Wondering if we could infer this from the class name or from the `$model` property?

canvural: No. As far as I know PHPStan does not support that. If you are extending a generic class, you have to have the `@extends`

nunomaduro: @canvural Let me give this a shot tonight if you don't mind.

canvural: Yeah sure. But there is nothing you do I think. Unless you find a solution without making `Factory` class generic.

nunomaduro: You were right. There is way we make users know, that generics are missing their Laravel factories?

canvural: PHPStan will complain if `@extends` is missing, in level 6 and above.

PR #865, opened 2021-07-05
Fix getting dateClass

### LINKED ISSUE #864: Unexpected token "|", expected type at offset 0
- Larastan Version: 0.7.8
- `--level` used: 5

### Description
When running `./vendor/bin/phpstan analyse --debug` I'm getting the error 

> PHPStan\PhpDocParser\Parser\ParserException 
> Unexpected token "|", expected type at offset 0
**at phar://vendor/phpstan/phpstan/phpstan.phar/vendor/phpstan/phpdoc-parser/src/Parser/TypeParser.php:87**
```
     83▕         } else {
     84▕             $tokens->dropSavePoint();
     85▕             // because of ConstFetchNode
     86▕         }
  ➜  87▕         $exception = new \PHPStan\PhpDocParser\Parser\ParserException($tokens->currentTokenValue(), $tokens->currentTokenType(), $tokens->currentTokenOffset(), \PHPStan\PhpDocParser\Lexer\Lexer::TOKEN_IDENTIFIER);
     88▕         if ($this->constExprParser === null) {
     89▕             throw $exception;
     90▕         }
     91▕         try {
```

migration: 

```php

 Schema::create('premiums', function (Blueprint $table)
        {
        
            // ...
            $table->date('date')->nullable();
            $table->date('start_date')->nullable();
            $table->date('end_date')->nullable();

```

Model: 

```php
class Premium extends Model
{
    use HasPresentation, HasFactory;

    protected $table = 'premiums';

    protected $casts = [
        'date'       => 'date',
        'start_date' => 'date',
        'end_date'   => 'date',
    ];

    public function newPresenter(): PremiumPresenter
    {
        return new PremiumPresenter($this);
    }

```
PremiumPresenter (where I get the error)
```php
class PremiumPresenter extends Presenter
{

    public function __construct(public Premium $premium)
    {
    }

    public function startDate(): string
    {

        if ($this->premium->start_date) { // I get the error here
            return $this->premium->start_date->toUserTimezone()->isoFormat('Ll');
        }

        return '';
    }
}

```

### Debugging details
When `ModelPropertyExtension` cast the property `start_date`, the propery `$this->dateClass` has never been initialized. (And I don't know the reason)

So when it reaches the following switch case, `$realType` will be `null`
https://github.com/nunomaduro/larastan/blob/v0.7.8/src/Properties/ModelPropertyExtension.php#L242

```php
switch ($type) {
    case 'date':
    case 'datetime':
        $realType = $this->dateClass; // null
        break;

```

Then, in this following, line `$realType` will be set to `|null` resulting with the error 
> Unexpected token "|", expected type at offset 0
https://github.com/nunomaduro/larastan/blob/v0.7.8/src/Properties/ModelPropertyExtension.php#L257

```php
if ($this->tables[$modelInstance->getTable()]->columns[$name]->nullable) {
    $realType .= '|null';
}
```

szepeviktor:
Fixes #864

- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

### DIFF
--- a/src/Properties/ModelPropertyExtension.php
+++ b/src/Properties/ModelPropertyExtension.php
@@ -239,7 +239,7 @@ final class ModelPropertyExtension implements PropertiesClassReflectionExtension
                     break;
                 case 'date':
                 case 'datetime':
-                    $realType = $this->dateClass;
+                    $realType = $this->getDateClass();
                     break;
                 case 'collection':
                     $realType = '\Illuminate\Support\Collection';


szepeviktor:
@canvural Could you come up with a test?

canvural (approved):

canvural:
@szepeviktor Yeah I think so. I'll add it.

canvural:
Ok, I can't come up with a test case. Anyway we now know we shouldn't access that property directly 😄

PR #868, opened 2021-07-06
Allow any return type for Collection::each

riasvdv:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

This allows you to have the following:

```php
User::all()->each(fn(User $user, int $key) => $user->doSomething());
```

Without getting the following error:

`Parameter #1 $callback of method Illuminate\\Support\\Collection<int,App\\User>::each() expects callable(App\\User, int): bool|void, Closure(App\\User): App\\User given.`

Can't really add a test for this as it's a PHP 7.4 only thing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Added the `mixed` return type for callables within a Collection's ->each statement to allow for short closures in PHP 8 ([#868](https://github.com/nunomaduro/larastan/pull/868))
+
 ## [0.7.9] - 2021-07-05
 
 ### Fixed
--- a/stubs/Collection.stub
+++ b/stubs/Collection.stub
@@ -57,7 +57,7 @@ class Collection implements \ArrayAccess, Enumerable
     public function shift() {}
 
     /**
-     * @param callable(TValue, TKey): (void|bool) $callable
+     * @param callable(TValue, TKey): (void|mixed) $callable
      * @return static<TValue>
      */
     public function each($callable) {}


szepeviktor:
Were arrow functions introduced in PHP 7.4?
https://www.php.net/manual/en/functions.arrow.php

riasvdv:
Ah yes, but Larastan supports 7.2, so that still would break things if we added a test for it

szepeviktor:
You **can** set return typehint for an arrow function 🎉 

```php
User::all()->each(fn(User $user, int $key): int => $user->doSomething());
```

szepeviktor:
> unnecessary noise,

Please do not use these words in a PHPStan extension's comment :)
PHPStan is all about types. Types makes your code robust.

```php
User::all()->each(fn(User $user, int $key): void => $user->doSomething());
```

canvural:
Hi,

I'm not sure if this is a good idea. The callback you pass into `each` should not return anything. You want to do **something** for **each** item. If you want to return something from callback you can `map`

And if the issue is just with the arrow syntax, like @szepeviktor said you can add return typehint to it.

riasvdv:
You can't add `void` as a return type to an arrow function (this throws an exception), and casting to `bool` doesn't seem correct either.

The Collection's `each` function doesn't care about what you return, why not allow it if you're using a short closure? We use a lot of chainable methods on our models, sometimes they're just used in a function and they return themselves.

riasvdv:
Also

> I'm not sure if this is a good idea. The callback you pass into each should not return anything. You want to do something for each item. If you want to return something from callback you can map

Why does it allow a `bool` return then?

riasvdv:
> And if the issue is just with the arrow syntax, like @szepeviktor said you can add return typehint to it.

We don't really want to cast a `User` model for example to `bool` that seems a lot more wrong in a strict type environment than returning a `User` object to a function that does nothing with the result.

szepeviktor:
@riasvdv Please understand that an arrow function is not able to return void by its nature. The expression after the `=>` sign is always returned.

A simple anonymous function should be use in `each()`.

szepeviktor:
When your function returns false, `each()` will stop.

https://github.com/laravel/framework/blob/fbf25e13b7ce9847fcc336bb3125933638acb936/src/Illuminate/Collections/Traits/EnumeratesValues.php#L225-L240

riasvdv:
@szepeviktor Yes I know that, but you used `void` in your example here so I wanted to make sure that was understood: https://github.com/nunomaduro/larastan/pull/868#issuecomment-874528155 

> A simple anonymous function should be used in each().

Yes, and we want to be able to use the short syntax for that. If you want to update each user you can't do it in a short closure:

```
User::all()->each(fn(User $user) => $user->update(['foo' => 'bar'));
```

Because Larastan will complain that you return a `User`. 

We wouldn't use Collection::map there because we don't care about the result.

It will never break your application if you return something `mixed` from Collection::each, why stop people from using it like this?

szepeviktor:
Arrow functions do return stuff. Please do not spare your keyboard!

freekmurze:
FWIW I think also that it makes sense to use `mixed` for this case. Even @nunomaduro is going about it this way in Laravel 9: https://github.com/laravel/framework/blob/b0a0d667210cc560b709d9d4d0900ed08f571064/src/Illuminate/Collections/Traits/EnumeratesValues.php#L247

PR #870, opened 2021-07-06
Document and test model attribute rule.

PrinsFrank:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

When reading the docs and tests for this package, it was not immediately clear that accessing properties on models directly - outside of methods and functions - are also analyzed. I added tests as I was planning to build it, but the functionality was already there! 💯 

**Breaking changes**

None

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -72,7 +72,7 @@ parameters:
 ```
 to your `phpstan.neon` file.
 
-This rule checks every argument of a method or a function, and if the argument has the type `model-property`, it will try to check the given value against the model properties. And if the model does not have the given property, it'll produce an error.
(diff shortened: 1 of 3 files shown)


szepeviktor (approved):

szepeviktor:
Thank you for your contribution!

BTW Where is "outside methods and functions"? :) Outer space? 👨‍🚀 👽 🌍 📡

canvural:
Hi,

Thanks for the contribution! But the `ModelPropertyRule` has nothing to do with accessing model properties on the model itself. Maybe the rule name is confusing little bit.

The way it works is, Larastan scans your migrations folder and tries to understand model properties. Also this is tested in `Features/Properties/ModelPropertyExtension`

Maybe we can clarify this in an other place.

szepeviktor:
I think @PrinsFrank is trying to say that `checkModelProperties` does not suppress checking properties **outside** classes.

PR #884, opened 2021-07-19
feat: support for `orWhere` HigherOrderBuilderProxy property

shaffe-fr:
- [x] Added or updated tests
- [ ] Documented user facing changes: not relevant
- [x] Updated CHANGELOG.md

**Changes**

HigherOrderBuilderProxy property `orWhere` is not supported by Larastan yet.

Closes #885

PR #886, opened 2021-07-19
Allow Laravel 9 to fail in CI

szepeviktor:
Very dirty!!!

From https://github.com/nunomaduro/larastan/pull/884#issuecomment-882705577


szepeviktor:
I do not want to do this ...

PR #887, opened 2021-07-20
Avoid need for explicit `@extends` in factories

spawnia:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

Follow up to https://github.com/nunomaduro/larastan/pull/861

**Changes**

The type of the used model in a factory can be inferred from the `$model` attribute.
This eliminates the need for a

**Breaking changes**

One test seems to be failing: `Tests\Rules\ModelPropertyRuleTest::testModelPropertyRuleOnModelFactory`

```
Internal error: Call to undefined method PHPStan\Type\MixedType::getClassReflection()\n
Run PHPStan with --debug option and post the stack trace to:\n
https://github.com/phpstan/phpstan/issues/new?template=Bug_report.md
```

### DIFF
--- a/stubs/Factory.stub
+++ b/stubs/Factory.stub
@@ -3,10 +3,17 @@
 namespace Illuminate\Database\Eloquent\Factories;
 
 /**
- * @template TModel of \Illuminate\Database\Eloquent\Model
+ * @template TModel extends \Illuminate\Database\Eloquent\Model
  */
 class Factory
 {
+    /**
+     * The name of the factory's corresponding model.
+     *
+     * @var class-string<TModel>
+     */
+    protected $model;
+
     /**
      * Get a new factory instance for the given attributes.
      *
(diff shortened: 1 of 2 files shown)


szepeviktor:
Does it work @spawnia ?

canvural:
Hi,

This does not actually work. I think tests are passing because it returns `mixed`. You can see [here.](https://phpstan.org/r/035b20a7-fb7b-49ec-b240-39a259ad9686)

Normally you'd get an error if you don't add `@extends` when extending a generic class. You can see [here.](https://phpstan.org/r/aba5d7a5-e99c-4709-8837-9fbcb9d05ca3) But we disabled that in our config (https://github.com/nunomaduro/larastan/blob/master/extension.neon#L47)

There is really no way around this. You can read [this](https://phpstan.org/blog/generics-in-php-using-phpdocs) article for more details. Specially:

> When implementing a generic interface or extending a generic class, you have two options:
> - Preserve the genericness of the parent, the child class will also be generic
> - Specify the type variable of the interface/parent class. The child class will not be generic.

PR #908, opened 2021-07-30
Update README.md

szepeviktor:
Stricter regex + fixing 1 typo.


szepeviktor:
I knew, I knew!

The average user does not excel at regex 😢

PR #912, opened 2021-08-07
Eloquent Builder global macros support.

LastDragon-ru:
- [x] Added or updated tests
- [ ] Documented user facing changes
- [ ] Updated CHANGELOG.md

**Changes**

Eloquent `Builder` (in v8 at least) [doesn't use the `Macroable` trait](https://github.com/laravel/framework/blob/dc393980dec1b3ee4b0f36b5b0188859a399b4e4/src/Illuminate/Database/Eloquent/Builder.php#L27-L32), it uses its own realization to mimic `Macroable`. Look like Larastan supports only local macros, but not global:

```php
Builder::macro('foo', static function (): string {
    return 'foo';
});

Builder::foo() // Call to an undefined static method Illuminate\\Database\\Eloquent\\Builder::foo().
```

This PR adds support for global macros, so

```php
Builder::foo() // = 'foo'
```

**Breaking changes**

Nope.

PR #914, opened 2021-08-10
Allows checking global macros on the Eloquent Builder

ChrisExP:
The Eloquent builder allows passing methods through to the wrapped query
builder which means it has support for macros, however it also has local
macros so the API is slightly different when checking for macros.

- [x] Added or updated tests
- [x] Updated CHANGELOG.md

I came across this problem when trying to use [lampager](https://github.com/lampager/lampager-laravel) which adds a [macro](https://github.com/lampager/lampager-laravel/blob/master/src/MacroServiceProvider.php#L20-L27) to the query builder as well as the eloquent builder. Larastan successfully checks the macro on the query builder but not for eloquent. This is because the eloquent builder doesn't use the `Macroable` trait, instead it wraps around the query builder and uses that trait as well as adding custom logic for local macros.

This pull request expands the `Macros` pipe to check if the class is the eloquent builder and calls the appropriate method to check for macros.

I haven't added any test cases yet because I'm still quite new to Larastan and PHPStan and I couldn't see how to do it well. The test would need to add a macro as it's building the application, possible in a service provider but I wasn't sure the best way to do that.

### DIFF
--- a/src/Methods/Pipes/Macros.php
+++ b/src/Methods/Pipes/Macros.php
@@ -73,11 +73,9 @@ final class Macros implements PipeContract
 
             $className = (string) $className;
 
-            if ($className === Builder::class) {
-                $found = $className::hasGlobalMacro($passable->getMethodName());
-            } else {
-                $found = $className::hasMacro($passable->getMethodName());
-            }
+            $found = $className === Builder::class
+                ? $className::hasGlobalMacro($passable->getMethodName())
+                : $className::hasMacro($passable->getMethodName());
 
             if ($found) {
                 $reflectionFunction = new \ReflectionFunction($refProperty->getValue()[$passable->getMethodName()]);
--- a/src/Methods/Pipes/Macros.php
+++ b/src/Methods/Pipes/Macros.php
@@ -6,6 +6,7 @@ namespace NunoMaduro\Larastan\Methods\Pipes;
 
 use Carbon\Traits\Macro as CarbonMacro;
 use Closure;
+use Illuminate\Database\Eloquent\Builder;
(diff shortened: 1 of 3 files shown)


src/Methods/Pipes/Macros.php
```diff
@@ -72,7 +73,13 @@ public function handle(PassableContract $passable, Closure $next): void
 
             $className = (string) $className;
 
-            if ($found = $className::hasMacro($passable->getMethodName())) {
+            if ($className === Builder::class) {
+                $found = $className::hasGlobalMacro($passable->getMethodName());
+            } else {
+                $found = $className::hasMacro($passable->getMethodName());
+            }
```
szepeviktor, line 80:
```suggestion
            $found = $className === Builder::class
                ? $className::hasGlobalMacro($passable->getMethodName())
                : $className::hasMacro($passable->getMethodName());
```

szepeviktor:
Thank you!
All is 🍏 

Please add one test and it'll be ready to merge.

szepeviktor (approved):

canvural:
Isn't this same as #912?

szepeviktor:
> Isn't this same as #912?

A better one.

ChrisExP:
Is this waiting for something? Can I help?

szepeviktor:
This PR is waiting for review.
See the top right gold dot under _Reviewers_

canvural:
Hi,

Thank you. Although I don't really like to add more runtime code to deal with Laravel's magic, this change looks small enough. So I guess it'll be fine 😄 

Thank you!

PR #924, opened 2021-09-01
Add Fluent class to universal object crates

ChrisExP:
The Fluent class acts as a wrapper around an array and allows accessing
and setting values through magic methods so it should qualify as a
universal object crate.

- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Marks the `Fluent` class as a universal object that can accept any property access.


szepeviktor:
Thank you @ChrisExP !
Very nice finding.
Let's see CI.

szepeviktor:
Does your IDE support EditorConfig?
https://github.com/nunomaduro/larastan/blob/99e7aec53075882ef3c254a2671baa994b89b188/.editorconfig#L11

ChrisExP:
> Does your IDE support EditorConfig?
> https://github.com/nunomaduro/larastan/blob/99e7aec53075882ef3c254a2671baa994b89b188/.editorconfig#L11

Yes but I didn't have it enabled :facepalm:

Sorry

canvural:
Hi,

I don't think we should add this class to `universalObjectCratesClasses` by default. If the user wants they can add it themselves. Plus, maybe in the future we can write an extension to `Fluent` class itself.

Thank you!

PR #932, opened 2021-09-10
support defining columns with foreignIdFor

Josh-G:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**
This PR adds support for using [`->foreignIdFor(YourModel::class)` in migrations](https://laravel.com/docs/8.x/migrations#column-method-foreignIdFor).

**Breaking changes**
None

### DIFF
--- a/src/Properties/SchemaAggregator.php
+++ b/src/Properties/SchemaAggregator.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace NunoMaduro\Larastan\Properties;
 
+use Illuminate\Support\Str;
 use PhpParser;
 use PhpParser\NodeFinder;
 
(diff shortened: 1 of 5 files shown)


szepeviktor:
Thank you for your hard work!

szepeviktor (approved):

src/Properties/SchemaAggregator.php
```diff
@@ -160,6 +160,43 @@ private function processColumnUpdates(string $tableName, string $argName, array
...
+                            $model = $reflect->newInstanceWithoutConstructor();
+
+                            $columnName = $model->getForeignKey();
+                            $columnType = $model->getKeyType() === 'int' && $model->getIncrementing()
+                                ? 'int'
+                                : 'string';
+
+                            if ($secondArg instanceof PhpParser\Node\Scalar\String_) {
+                                $columnName = $secondArg->value;
+                            }
```
canvural, line 191: Can we replace this code here with `$columnName = Str::snake(class_basename($modelClass)) . '_id'` ? I think this will cover most of the cases, and we would avoid using reflection.

szepeviktor:
Nice to see you working, Josh!

Josh-G:
test failures are unrelated (#934)

src/Properties/SchemaAggregator.php
Josh-G: this was based on similar reflection in https://github.com/nunomaduro/larastan/blob/e35de81506c5208ad8cf923f6dae0034e51f4615/src/Properties/ModelPropertyExtension.php#L74-L79

src/Properties/SchemaAggregator.php
```diff
@@ -160,6 +161,27 @@ private function processColumnUpdates(string $tableName, string $argName, array
...
+                        } else {
+                            continue;
+                        }
+
+                        $columnName = Str::snake(class_basename($modelClass)).'_id';
+                        if ($secondArg instanceof PhpParser\Node\Scalar\String_) {
+                            $columnName = $secondArg->value;
+                        }
+
+                        $table->setColumn(new SchemaColumn($columnName, 'mixed', $nullable));
```
canvural, line 180: Why `mixed` ? We can do `int` I think. That'd cover more than %80 of the cases. If people are using UUID and have `string` primary keys, they can override this with `@property` annotations over the class.

canvural (changes requested):
Also we can add `@property string $address_id` to the `User` model in tests. Then we can change the test to use the correct type instead of `mixed`

canvural:
Thank you!

PR #934, opened 2021-09-14
fix tests

Josh-G:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**
Fix failing test introduced by line number changes in b15252672a9677e0dbb4336cde16c2407098b33c
Fix failing `CollectionStub` tests caused by changes to `checkExplicitMixed` in https://github.com/phpstan/phpstan-src/commit/6ba9ef29cb5ccd5b540815e899c95e6a1602ccad
I'm not 100% sure about the Collection related changes and the effect this will have on existing projects. I am happy to make amendments if the stub changes need adjusting.

**Breaking changes**

### DIFF
--- a/stubs/Collection.stub
+++ b/stubs/Collection.stub
@@ -71,7 +71,7 @@ class Collection implements \ArrayAccess, Enumerable
 
     /**
      * @param callable(TValue, TKey): array<mixed> $callback
-     * @return static<mixed, mixed>
+     * @return static<mixed, static<mixed>>
      */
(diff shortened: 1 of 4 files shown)


canvural:
> I'm not 100% sure about the Collection related changes and the effect this will have on existing projects. I am happy to make amendments if the stub changes need adjusting.

They look ok. We can make changes if anyone complains 😄 Also with Laravel 9 we will have this in the framework itself.

Looks good. Thank you!

Issue #831, opened 2021-05-20
Any way to register service providers?


nasyrov:
Hi there,

I'm working on a custom Laravel package, which uses quite a few other packages as dependencies. For instance, Livewire. When I try to run phpstan against my codebase it complains that classes haven't been registered in the container:

```
Note: Using configuration file /home/runner/work/laravel-custom/laravel-custom/phpstan.neon.dist.
Illuminate\Contracts\Container\BindingResolutionException thrown in /home/runner/work/laravel-custom/laravel-custom/vendor/laravel/framework/src/Illuminate/Container/Container.php on line 838 while loading bootstrap file /home/runner/work/laravel-custom/laravel-custom/vendor/nunomaduro/larastan/bootstrap.php: Target class [livewire] does not exist.
```

Is there any way we can register additional service providers when developing a custom package?

Cheers,
E

canvural:
Hello,

Do you have `orchestra/testbench` installed? If not, installing this will probably help you.

szepeviktor:
You may also want to see this.

https://github.com/szepeviktor/bazar/blob/467466c6300c4fb07eafebd020e13dd580c31e0a/.github/workflows/static-analysis.yml#L30-L32

nasyrov:
@canvural I got `orchestra/testbench` installed, doesn't help.

@szepeviktor thanks, saw it, but this one is migration-related, isn't it?

I think the problem might be here: https://github.com/nunomaduro/larastan/blob/master/src/ApplicationResolver.php#L46

When we bootstrap the application for the package we only register the actual package service provider and the rest got ignored, hence we have an error related to missing bindings in the container...

BertvanHoekelen:
I think [enabling package](https://packages.tools/testbench/basic/testcase.html#package-auto-discovery) auto-discovery in testbench should do the trick.

wijourdil:
Hi, I have the same issue (except it's with the package `nWidart/laravel-modules` and not Livewire), and I don't know how to fix it 😞 

> I think enabling package auto-discovery in testbench should do the trick.

Are you sure ? I have tried but it still does not work :/ 
But it seems logical to me, the method `ignorePackageDiscoveriesFrom` must be overrided in the `TestCase` class, which phpstan does not load or include 🤔

szepeviktor:
I'm am very curious of your findings.

wijourdil:
I think I have found a workaround using the testbench bootstrap file.

[...]

PR #943, opened 2021-09-25
Allow users to boot other ServiceProviders by adding them to composer…

antonioribeiro:
- [ ] Added or updated tests
- [X] Updated CHANGELOG.md

Resolves https://github.com/nunomaduro/larastan/issues/831

**Changes**

Currently larastan supports registering Service Providers only from the **first** namespace you put in your `composer.json` autoload object. So if you define: 

``` json
{
    ...
    "autoload": {
        "psr-4": {
            "A17\\EdgeFlush\\": "src",
            "A17\\EdgeFlush\\Database\\Factories\\": "database/factories"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "Spatie\\ResponseCache\\": "vendor/spatie/laravel-responsecache/src"
        }
    },
    ...
}
```

It will only register Service Providers from the `A17\EdgeFlush` namespace. And if you define namespaces on your `autoload-dev` it will, by design, completely ignore them too. But while developing packages, we usually need all third party Service Providers, we are using, to be registered.

This PR changes it by booting those defined in `composer.json`. And, if it's running on a package it will also register Service Providers defined in `autoload-dev`. To enable it developers have to set the Composer type to "library".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Support global macros on the eloquent query builder.
 - Partial support for defining columns with `foreignIdFor` in migrations. ([#932](https://github.com/nunomaduro/larastan/pull/932)) Thanks @Josh-G
+- Allow users to boot other ServiceProviders by adding them to composer.autoload
 
 ## [0.7.12] - 2021-07-26
 
--- a/README.md
+++ b/README.md
@@ -71,6 +71,50 @@ If you are getting the error `Allowed memory size exhausted`, then you can use t
 ./vendor/bin/phpstan analyse --memory-limit=2G
 ```
 
+## Loading other Service Providers (for package development)
+
+When facing errors like, it's possible that even Testbench is not able to load the Service Provider you need, and you may an error like:
+
+```
+Target
+    [Spatie\ResponseCache\Serializers\Serializer]
+is not instantiable while building
+    [Spatie\ResponseCache\ResponseCacheRepository]
+```
+
+To load them you have to
+
+#### Configure Composer "type" to "library"
+
+This is the name Composer uses for packages:
+
+``` json
+{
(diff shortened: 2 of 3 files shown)


szepeviktor:
Thank you for your PR!

I'm not sure `autoload-dev` is the right place to store Larastan-specific information.

antonioribeiro:
It's a matter of semantics:

- I'm developing (dev) a package for Laravel
- I need Laravel to load (autoload) up more Service Providers
- The piece of software responsible for registering namespaces and autoloading classes is Composer

Looks to me that Composer's `autoload-dev` object is the right place to configure them.

But, honestly, I just extended something [larastan was already doing](https://github.com/nunomaduro/larastan/blob/master/src/ApplicationResolver.php#L38), it's just that it was ignoring all except the first namespace.

szepeviktor:
> But, honestly, I just extended something [larastan was already doing](https://github.com/nunomaduro/larastan/blob/master/src/ApplicationResolver.php?rgh-link-date=2021-09-25T16%3A46%3A46Z#L38), it's just that it was ignoring all except the first namespace.

Larastan read the app's namespace.
This PR is putting foreign namespaces in autoload-dev.
This is not my project, I'm only a janitor! 🧹

antonioribeiro:
larastan reads the first namespace, which may not be the app/package namespace. So if I people do 

```
{
    "autoload": {
        "psr-4": {
            "Spatie\\ResponseCache\\": "vendor/spatie/laravel-responsecache/src",
            "A17\\EdgeFlush\\": "src"
        }
    },
}
```

It will not load the "app namespace". And, as told before, this change will probably really only impact package developers, which have usually other needs.

szepeviktor:
All I can think of is https://getcomposer.org/doc/04-schema.md#extra

I'm not strong enough to argue. Please wait for someone else.

canvural:
This should be now fixed in a different way in the latest commits.

PR #947, opened 2021-09-30
Fix Macro::getFileName

szepeviktor:
Fix return typehint.


szepeviktor:
To myself: [_Not so fast, Yakamoto!_](https://www.ibras.dk/montypython/episode29.htm)

Issue #863, opened 2021-07-03
Feature: Make possible to use custom bootstrap files (Don't throw fatal errors in bootstrap.php)


rajyan:
- Larastan Version: 0.7.8
- `--level` used: 5

I created this issue to know if this feature is acceptable or not.

### Description

For now, if we have a Laravel application that doesn't satisfy
```php
if (file_exists($applicationPath = getcwd().'/bootstrap/app.php'))
```
in https://github.com/nunomaduro/larastan/blob/master/bootstrap.php#L9

larastan would throw an fatal error
```
Fatal error: Trait 'Orchestra\Testbench\Concerns\CreatesApplication' not found in /var/source/app/laravel/vendor/nunomaduro/larastan/src/ApplicationResolver.php on line 16
```

I want to configure bootstrap files by my own, and not installing `orchestra\testbench` because it's not a package.

### Feature

Check if class `Orchestra\Testbench\Concerns\CreatesApplication` exists and add a notice if larastan is called without bootstrapping.
For example,
```php
elseif (class_exists('Orchestra\Testbench\Concerns\CreatesApplication')){ // Packages
    $app = ApplicationResolver::resolve();
}
else {
    // add a notice here
    return;
}
```
https://github.com/nunomaduro/larastan/blob/master/bootstrap.php#L11

The notice should be like,
"Install orchestra\testbench for Laravel packages or add bootstrapFiles to phpstan.neon to boot Laravel application"

I thought about making bootstrapFile configurable in larastan, but I think it is too much and this is a simpler and cleaner solution.

### Note

I think this feature would help solving issues like 
https://github.com/nunomaduro/larastan/issues/831
and the notice would be helpful for people who wants to use larastan on Laravel packages like 
https://github.com/nunomaduro/larastan/pull/847

nunomaduro:
Feel free to make a pull request with your proposal.

PR #858, opened 2021-07-02
Revert "fix bootstrap path (#852)"

rajyan:
This reverts commit d8d1a901428acd28c5bafe37027d1e96a8d397ac.

- [ ] Added or updated tests
- [ ] Documented user facing changes
- [ ] Updated CHANGELOG.md

reverts as it was a breaking change for people using symlinks.

fixes https://github.com/nunomaduro/larastan/issues/856

**Changes**

reverts #852 

**Breaking changes**

PR #957, opened 2021-10-09
fix: adds trait_exists to ensure orchestra/testbench is installed

rajyan:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

resolves https://github.com/nunomaduro/larastan/issues/863 

Related PRs
* https://github.com/nunomaduro/larastan/pull/852
* https://github.com/nunomaduro/larastan/pull/858

I personally think `getcwd()` is not a good solution to 
find Laravel Application bootstrap file, but could not find a better solution that solve cases like https://github.com/nunomaduro/larastan/issues/856.

so, for cases that Laravel application bootstrap file cannot be loaded by `getcwd().'/bootstrap/app.php'`, 
I just don't want 
```
Fatal error: Trait 'Orchestra\Testbench\Concerns\CreatesApplication' not found in /var/source/app/laravel/vendor/nunomaduro/larastan/src/ApplicationResolver.php on line 16
```
 and handle bootstrap by my own.

This PR adds `trait_exists` to ensure `Orchestra\Testbench\Concerns\CreatesApplication` exists, which is called in 
`ApplicationResolver::resolve()`
https://github.com/nunomaduro/larastan/blob/master/src/ApplicationResolver.php#L18

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -2,7 +2,6 @@
 
 declare(strict_types=1);
 
-use Exception;
 use Illuminate\Contracts\Console\Kernel;
 use Illuminate\Contracts\Foundation\Application;
 use Laravel\Lumen\Application as LumenApplication;


rajyan:
I'll wait for https://github.com/nunomaduro/larastan/pull/955

rajyan:
@szepeviktor 
Thank you for your fix & review.

I added a Exception.
Also, retried https://github.com/nunomaduro/larastan/pull/852 , being careful about backward compatibility this time.

szepeviktor:
>  Also, retried #852 , being careful about backward compatibility this time.

Please prioritize `getcwd()` as it was before.

PR #985, opened 2021-10-28
feat: adds new rule to check for relation existence

### LINKED ISSUE #974: Check for relation existence in Builder methods
For example `User::whereHas('foobar')->get()` should produce error `Relation 'foobar' is not found in 'User' model`

### Related builder methods are
- `has`
- `orHas`
- `doesntHave`
- `orDoesntHave`
- `whereHas`
- `orWhereHas`
- `whereDoesntHave`
- `orWhereDoesntHave`

canvural:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Resolves #974 

This PR adds a new rule to check for the existence of the relations. It also supports nested relations.


bernardwiesner:
@canvural  thanks for the feature.

Is it necessary inside Models to specify the return type?

```php
    /**
     * @return \Illuminate\Database\Eloquent\Relations\BelongsTo
     */
    public function user()
    {
        return $this->belongsTo(User::class);
    }
```

If I do not specify the return type I get an error such as:

> Relation 'user' is not found in App\Post model. 

I thought the relation is already returned in laravel HasRelationships class, but doesnt seem to work without specifying.

szepeviktor:
Hello @bernardwiesner !
Please open an issue.

`belongsTo` needs to types: https://github.com/nunomaduro/larastan/blob/6e74c6eeb0b27227c787917afd559a2d22ec241c/tests/Features/Properties/ModelRelationsExtension.php#L110-L114
https://github.com/nunomaduro/larastan/blob/f74befd87b7278285cbf6f1597509cbfaca46bf8/stubs/BelongsTo.stub#L6-L7

canvural:
@bernardwiesner Yes you need to have return types for your relations. Larastan needs to know if a relation method is in your model file. And it does that by checking the method name and return types. PHPStan does not scan method bodies, so we only see method names, comments and return types.

And no need to open a new issue.

bernardwiesner:
@canvural I see, thanks for explaining!

Issue #927, opened 2021-09-08
[Feature Request] validate Job::dispatch() args 


dianfishekqi:
### Description
Currently the only way to make sure  larastan validates Job arguments correctly is to use `dispatch(new Job(...$args))`
calling `Job::dispatch(...$args)` fails to catch missing arguments/argument types

szepeviktor:
Hello @dianfishekqi! Thank you for your report.

PR-s are welcome.

szepeviktor:
Actually this is a PHPStan feature: detecting constructor parameters.

`Job::dispatch` needs a new extension (PHP code in Larastan)

canvural:
Hi,

I'm not sure if this is even possible with PHPStan.

@ondrejmirtes Is it possible to write some kind of extension or something to say that a method should have the same signature as the constructor? [Here](https://phpstan.org/r/05b49eb6-ff9c-470d-9233-7baab9245861) is an example. A string is passed to `make`, and this will [fail in runtime](https://3v4l.org/3dbZY).

ondrejmirtes:
@canvural You can definitely write a custom rule for this :) Feel free to check out the new guide if you don't know where to start: https://phpstan.org/developing-extensions/rules

szepeviktor:
Thank you.

@ondrejmirtes I always wanted to ask: what are _extensions_ and _rules_? Is one the subset of the other?

ondrejmirtes:
@szepeviktor See: https://phpstan.org/developing-extensions/extension-types

canvural:
Implemented in #991

PR #991, opened 2021-10-30
feat: adds CheckJobDispatchArgumentTypesCompatibleWithClassConstructorRule rule

canvural:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Implements #927 

**Changes**

This PR adds a rule that can check arguments given in `dispatch` (and other variants) method against the job class constructor.

### DIFF
--- a/src/Rules/CheckJobDispatchArgumentTypesCompatibleWithClassConstructorRule.php
+++ b/src/Rules/CheckJobDispatchArgumentTypesCompatibleWithClassConstructorRule.php
@@ -1,0 +1,146 @@
+<?php
+
+declare(strict_types=1);
+
+namespace NunoMaduro\Larastan\Rules;
+
+use Illuminate\Foundation\Bus\Dispatchable;
+use PhpParser\BuilderFactory;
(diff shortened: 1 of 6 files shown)


src/Rules/CheckJobDispatchArgumentTypesCompatibleWithClassConstructorRule.php
```diff
@@ -0,0 +1,146 @@
...
+                        $jobClassReflection->getDisplayName()
+                    ))->build(),
+                ];
+            }
+
+            return [];
+        }
+
+        if (in_array($methodName, ['dispatchIf', 'dispatchUnless'], true)) {
+            if (count($node->getArgs()) < 1) {
```
szepeviktor, line 98:
```suggestion
            if ($node->getArgs() === []) {
```
Could we be explicit?

canvural: Sure. Thanks.

PR #1010, opened 2021-11-05
feat: improve return types for Collection::first, last, get and pull when default value is given.

bastien-phi:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

This PR improves return type resolution for Collection::first, last, get and pull when default value is given.

Today, 
```php
User::all()->first(default: new User);
```
is typed as `User|null`, even if it actually never returns null.

With this PR, it will be typed as `User`.

Strange calls are also supported
```php
User::all()->first(default: 42); // User|int
```
(I really don't know why someone would write this but why not ?)

This PR does not add support for callable as default
```php
User::all()->first(default: fn() => new User);
```
My guess is it will be a little bit more complicated than just a stub update.

**Breaking changes**

I hope there won't be any

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * Fix `groupBy` Collection method signature
 * Renamed `CheckJobDispatchArgumentTypesCompatibleWithClassConstructorRule` to `CheckDispatchArgumentTypesCompatibleWithClassConstructorRule` by @canvural in https://github.com/nunomaduro/larastan/pull/1003
 * `CheckDispatchArgumentTypesCompatibleWithClassConstructorRule` now checks for argument types compatibility with the class constructor for both Event and Job classes. by @canvural in https://github.com/nunomaduro/larastan/pull/1003
+* Improved return type for Collection::first, last, get, pull when giving a  default value.
 
 ### Improvements
 * Clarified "static analysis" wording in docs and logo by @asgrim in https://github.com/nunomaduro/larastan/pull/1008
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+* Improved return type for Collection::first, last, get, pull when giving a  default value.
+
 ## [1.0.1] - 2021-11-06
 
 ### Changed
 * Fix `groupBy` Collection method signature
 * Renamed `CheckJobDispatchArgumentTypesCompatibleWithClassConstructorRule` to `CheckDispatchArgumentTypesCompatibleWithClassConstructorRule` by @canvural in https://github.com/nunomaduro/larastan/pull/1003
 * `CheckDispatchArgumentTypesCompatibleWithClassConstructorRule` now checks for argument types compatibility with the class constructor for both Event and Job classes. by @canvural in https://github.com/nunomaduro/larastan/pull/1003
-* Improved return type for Collection::first, last, get, pull when giving a  default value.
 
 ### Improvements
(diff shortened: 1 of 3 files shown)


szepeviktor:
Thank you.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Renamed `CheckJobDispatchArgumentTypesCompatibleWithClassConstructorRule` to `CheckDispatchArgumentTypesCompatibleWithClassConstructorRule`
 - `CheckDispatchArgumentTypesCompatibleWithClassConstructorRule` now checks for argument types compatibility with the class constructor for both Event and Job classes.
+- Improved return type for Collection::first, last, get, pull when giving a  default value.
```
szepeviktor, line 12:
```suggestion
- Improved return type for Collection::first, last, get, pull when giving a default value.
```

stubs/Collection.stub
```diff
@@ -11,24 +11,27 @@ namespace Illuminate\Support;
...
      */
     public function last(callable $callback = null, $default = null){}
 
 
     /**
-     * @param  mixed  $key
-     * @param  mixed  $default
-     * @return TValue|null
+     * @template TDefault
+     * @param    mixed  $key
```
canvural, line 32: I think this can be just `string` or `int` isn't it? In that case the type here can be `array-key` instead of `mixed`

stubs/Collection.stub
```diff
@@ -38,9 +41,10 @@ class Collection implements \ArrayAccess, Enumerable
     public function pop() {}
 
     /**
-     * @param  mixed  $key
-     * @param  mixed  $default
-     * @return TValue|null
+     * @template TDefault
+     * @param    mixed  $key
```
canvural, line 45: I think this can be just `string` or `int` isn't it? In that case the type here can be `array-key` instead of `mixed`

canvural (changes requested):
Hi,

Thank you. It looks. Just left two tiny comments.

stubs/Collection.stub
bastien-phi: In my opinion, it should even be TKey. I didn't want to change it in this PR but if you want I can add it. I just don't know where to add tests for this

canvural: I think you can test it like `assertType('*ERROR*', $collection->get(3.14));`

bastien-phi: Unfortunately, it doesn't work... Maybe I should add tests in `tests/Rules` and a file in `tests\Rules\Data` but I don't think this is the right place...

canvural (approved):
If you can rebase and fix the conflicts we can merge! 👍🏽

bastien-phi:
Should be good !

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * Fix `groupBy` Collection method signature
 * Renamed `CheckJobDispatchArgumentTypesCompatibleWithClassConstructorRule` to `CheckDispatchArgumentTypesCompatibleWithClassConstructorRule` by @canvural in https://github.com/nunomaduro/larastan/pull/1003
 * `CheckDispatchArgumentTypesCompatibleWithClassConstructorRule` now checks for argument types compatibility with the class constructor for both Event and Job classes. by @canvural in https://github.com/nunomaduro/larastan/pull/1003
+* Improved return type for Collection::first, last, get, pull when giving a  default value.
```
canvural, line 15: This should be under the `[Unreleased]` section

canvural (changes requested):

CHANGELOG.md
bastien-phi: :facepalm: 
Changing it

canvural:
Thank you!

bastien-phi:
Glad to contribute !

PR #1014, opened 2021-11-07
Update default config to match new highest level

felixdorn:
Boilerplate wasn't really applicable. The new highest level in PHPStan is 9, this change reflects that in the documentation where it was previously stated that the max level was 8.


szepeviktor:
Thank you Félix!

PR #1017, opened 2021-11-08
Import classes in RedirectResponseMethodsClassReflectionExtension

szepeviktor:
Just importing classes.


canvural:
Thanks, but I like the `Reflection\ClassReflection` style more. Less imports.

PR #1027, opened 2021-11-13
fix: Eloquent collection has array-key for the key type

### LINKED ISSUE #1026: Eloquent and Regular collection generic types
- Larastan Version: 1.0.1
- `--level` used: 3
- Pull request with failing test: ?

### Description

When calling the method `keyBy` on an instance of `Illuminate\Database\Eloquent\Collection` and then attempting to access this collection with a string offset, the analysis complains that
"Cannot assign offset string to Illuminate\Database\Eloquent\Collection<CustomModel>"
This is due to the method returning `new static()`.
However, since the type arguments differ on both classes' stubs (Key and Value types on Support\Collection,  only value on Eloquent\Collection), cannot use `@var \Illuminate\Support\Collection<string, CustomModel>` as a forced type-cast.

I understand that I can use `@var \Illuminate\Support\Collection` but I would like to be able to strongly type this.

Using any CustomModel, do a query, then call ->keyBy('attr'). Then, access the result with any string.
```php
$keyed_col = CustomModel::get()->keyBy('name');
if(isset($keyed_col["abc"])) {}
```

canvural:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Fixes #1026 

**Changes**

Eloquent collection can have `array-key` as the `Collection` key type.

**Breaking changes**

### DIFF
--- a/stubs/Collection.stub
+++ b/stubs/Collection.stub
@@ -3,7 +3,7 @@
 namespace Illuminate\Support;
 
 /**
- * @template TKey
+ * @template TKey of array-key
  * @template TValue
  * @implements \ArrayAccess<TKey, TValue>
  * @implements Enumerable<TKey, TValue>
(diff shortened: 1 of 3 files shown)


szepeviktor:
Good morning!

PR #1037, opened 2021-11-18
fix: respect ancestor property tags in model extensions

spawnia:
- [x] Added or updated tests
- [ ] ~Documented user facing changes~
- [x] Updated CHANGELOG.md

**Changes**

Ensure the model extensions considers PHPDoc `@property` tags from ancestors, not just the model class itself.

**Breaking changes**

None.

### DIFF
--- a/src/Reflection/ReflectionHelper.php
+++ b/src/Reflection/ReflectionHelper.php
@@ -4,27 +4,25 @@ declare(strict_types=1);
 
 namespace NunoMaduro\Larastan\Reflection;
 
-use PHPStan\PhpDoc\Tag\PropertyTag;
 use PHPStan\Reflection\ClassReflection;
 
 final class ReflectionHelper
 {
     /**
-     * Returns all property tags of the given class and its ancestors.
-     *
-     * In case of duplicates, the tags of the class itself have precedence.
-     * TODO consider the hierarchy to ensure correct precedence between ancestors
-     *
-     * @return array<string, PropertyTag>
+     * Does the given class or any of its ancestors have an `@property*` annotation with the given name?
      */
-    public static function collectPropertyTags(ClassReflection $classReflection): array
+    public static function hasPropertyTag(ClassReflection $classReflection, string $propertyName): bool
     {
-        $allPropertyTags = $classReflection->getPropertyTags();
-
-        foreach ($classReflection->getAncestors() as $ancestor) {
-            $allPropertyTags += $ancestor->getPropertyTags();
+        if (array_key_exists($propertyName, $classReflection->getPropertyTags())) {
+            return true;
         }
 
-        return $allPropertyTags;
+        foreach ($classReflection->getAncestors() as $ancestor) {
+            if (array_key_exists($propertyName, $ancestor->getPropertyTags())) {
+                return true;
+            }
+        }
+
+        return false;
     }
 }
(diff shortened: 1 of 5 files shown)


src/Reflection/ReflectionHelper.php
```diff
@@ -0,0 +1,30 @@
...
+use PHPStan\PhpDoc\Tag\PropertyTag;
+use PHPStan\Reflection\ClassReflection;
+
+final class ReflectionHelper
+{
+    /**
+     * Returns all property tags of the given class and its ancestors.
+     *
+     * In case of duplicates, the tags of the class itself have precedence.
+     * TODO consider the hierarchy to ensure correct precedence between ancestors
```
spawnia, line 16: Are there existing facilities in PHPStan/Larastan to do this?

If not, I would rather not implement this perfectly right now. It seems pretty hard to get right. Don't let perfect be the enemy of improvement 🤞

szepeviktor:
Hello Benedikt!
Thank you for your work.
Are you sure you want to encourage users to manually add many-many properties?
Larastan has a lot of effort recognizing magic properties in Laravel.

spawnia:
I think it is great that Larastan can recognize magic properties, but there are still situations that require an escape hatch:
- `BelongsTo` or `HasOne` relations that are guaranteed to be non-null (backed by a non-nullable foreign key constraint)
- specializations of generic implementations in a parent class, e.g. parent defines `getIdAttribute(): int | string`, but the ID is known to always be `string` in a given subclass

Larastan seems to recognize the necessity of this, given there is already an early return if a property annotation exists. I don't see why this logic should not extend to annotations from ancestors.

src/Reflection/ReflectionHelper.php
```diff
@@ -0,0 +1,30 @@
...
+    public static function collectPropertyTags(ClassReflection $classReflection): array
+    {
+        $allPropertyTags = $classReflection->getPropertyTags();
+
+        foreach ($classReflection->getAncestors() as $ancestor) {
+            $allPropertyTags += $ancestor->getPropertyTags();
+        }
+
+        return $allPropertyTags;
+    }
```
canvural, line 29:
```suggestion
    public static function hasPropertyTagInAncestors(ClassReflection $classReflection, string $propertyName): bool
    {
        foreach ($classReflection->getAncestors() as $ancestor) {
            if (array_key_exists($propertyName, $ancestor->getPropertyTags())) {
                return true;
            }
        }

        return false;
    }
```

I'd prefer something like this. So we don't need to collect every property tag. We can return early. And of course the calling code should be changed accordingly.

And I think we shouldn't be concerned too much about the hierarchy. Defining the same property with different types is rare.

canvural (changes requested):
Thanks! I had the same thing in mind actually.

Just left one remark. And after that I will refactor the test cases for this.

szepeviktor:
> * `BelongsTo` or `HasOne` relations that are guaranteed to be non-null (backed by a non-nullable foreign key constraint)
> 
> * specializations of generic implementations in a parent class, e.g. parent defines `getIdAttribute(): int | string`, but the ID is known to always be `string` in a given subclass

I would still welcome these features in Larastan.

src/Reflection/ReflectionHelper.php
spawnia: Done: [`6f84347` (#1037)](https://github.com/nunomaduro/larastan/pull/1037/commits/6f84347628ac5855da2959d4090710928e1ed856)

We don't have to worry about the hierarchy with this approach, since the method now only cares if the property is defined anywhere at all.

canvural:
Thank you!

PR #1039, opened 2021-11-18
Call to an undefined method Illuminate\Database\Eloquent\Builder<App\User>::restore()

MatanYadaev:


szepeviktor:
We test this scenario.
https://github.com/nunomaduro/larastan/blob/72aa0afcb775ab98cdd412f7e3a9bf9ff43f8f49/tests/Features/Methods/ModelExtension.php#L202-L205

szepeviktor:
```php
User::query()->onlyTrashed()->where('id', $post_id)->restore();
```

Maybe Laravel wants to restore 1 user at a time.

PR #1043, opened 2021-11-19
feat: run tests on PHP 8.1

canvural:


canvural:
Looks like only Laravel 8 and 9 has PHP 8.1 support.

PR #1044, opened 2021-11-19
Precise return type for `Enumerable::partition()`

spawnia:
- [x] Added or updated tests
- [ ] ~Documented user facing changes~
- [x] Updated CHANGELOG.md

**Changes**

Fixes the false-positive error reported in this example:

image

**Breaking changes**

None, analysis will be more precise.

### DIFF
--- a/tests/Type/data/collection-stubs.php
+++ b/tests/Type/data/collection-stubs.php
@@ -104,3 +104,11 @@ assertType('App\User', $collection->get(1, new User()));
 
 assertType('App\User|null', $collection->pull(1));
 assertType('App\User', $collection->pull(1, new User()));
+
+$partitions = collect(['foo' => 1])->partition(function ($item, $key): bool {
+    assertType('int', $item);
+    assertType('string', $key);
+
+    return true;
+});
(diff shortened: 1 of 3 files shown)


tests/Type/data/collection-stubs.php
```diff
@@ -104,3 +104,11 @@
...
 assertType('App\User|null', $collection->pull(1));
 assertType('App\User', $collection->pull(1, new User()));
+
+$partitions = collect(['foo' => 1])->partition(function ($item, $key): bool {
+    assertType('int', $item);
+    assertType('string', $key);
+
+    return true;
+});
+assertType('Illuminate\Support\Enumerable<int<0, 1>, Illuminate\Support\Enumerable<string, int>>', $partitions);
```
spawnia, line 114: Shouldn't this be `Collection`? Does late static binding not work with `assertType()` and stubs?

canvural:
Hi,

All the collection stubs are moved to Laravel core now. And Larastan 2 will remove it from this repo. So we will no longer be accepting PRs that are modifying collection stubs. Thank you for understanding.

PR #1054, opened 2021-11-29
Add failing test case for Collection::groupBy() stub

spawnia:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Only a failing test for now.

**Breaking changes**

### DIFF
--- a/tests/Type/data/collection-stubs.php
+++ b/tests/Type/data/collection-stubs.php
@@ -41,16 +41,47 @@ assertType(
 assertType('Illuminate\Database\Eloquent\Collection<App\User>', $collection->tap(function ($collection): void {
 }));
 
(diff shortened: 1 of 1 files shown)


spawnia:
I think there might be two issues here:
- `collect()` is not equivalent to `new \Illuminate\Support\Collection()`: my guess is this can be fixed by adding an extension that does the same inference as `ReturnType\Helpers\CollectionExtension` but for direct usage
- late static binding within `stubs/Collection.stub::groupBy()` is not considered: this might be related to https://github.com/nunomaduro/larastan/pull/1044#discussion_r753321662 and I don't have ideas on how to fix it right now

canvural:
Hi,

All the collection stubs are moved to Laravel core now. And Larastan 2 will remove it from this repo. So we will no longer be accepting PRs that are modifying collection stubs. Thank you for understanding.

PR #1065, opened 2021-12-09
Fix SchemaAggregation when renaming tables through non-static method call

N-Silbernagel:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

In Laravel migrations we can rename a table by getting a reference to the table through Schema::table and renaming the table via the rename method. 
Larastan analyzes those cases incorrectly, but rather treats them the same as a "renameColumn" call, resulting in "trying to access undefined property x" errors on models whose tables were renamed through that method.
This PR aims at changing that.

The method does not seem to be well documented but when you already have migrations using the method, it's hard to change them afterwards.

**Breaking changes**

I'm not aware of any.

### DIFF
--- a/src/Properties/SchemaAggregator.php
+++ b/src/Properties/SchemaAggregator.php
@@ -76,7 +76,7 @@ final class SchemaAggregator
                         break;
 
                     case 'rename':
-                        $this->renameTable($stmt->expr);
+                        $this->renameTableThroughStaticCall($stmt->expr);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


szepeviktor:
Hello Nils! 👋 
Thank you for your work.

canvural:
Thank you!

N-Silbernagel:
Happy to contribute, love how the project is tested. Keep up the great work

PR #1079, opened 2022-01-05
Add version constrain for PHPStan <1.3

szepeviktor:
As we have multiple issues.

PR #1090, opened 2022-01-14
Change PHPStan version to minimum 1.4

bastien-phi:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Fix the tests for PHPStan 1.4

Bumped composer.json as version 1.3 introduced some bugs which were fixed in 1.4

**Breaking changes**

Not compatible anymore with PHPStan < 1.4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+* PHPStan minimum version is 1.4
+
 ## [1.0.2] - 2021-11-23
 
 ### Added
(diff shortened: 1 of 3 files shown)


szepeviktor:
I think we cannot just a skip a PHPStan version.
It'll be resolved. Please see #1079

szepeviktor:
But thank you for your PR!!

bastien-phi:
Not sure we need to support every minor version of PHPStan but this is your call !

bastien-phi:
As a fact, Larastan is only tested with latest stable version of PHPStan, not all of them : 

https://github.com/nunomaduro/larastan/blob/08c4f4f4c714b4adaa4fa2bc944db88d2d20471e/.github/workflows/tests.yml#L59-L60

canvural:
Hi,

I did not see any errors with PHPStan 1.3 So this is not necessary.

PR #1093, opened 2022-01-19
feat: enable Testbench package discovery

mewejo:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**
I was having problems with Laravel being able to resolve an interface which was bound to the container in a service provider from another package (which the package I am testing depends on). Tests with Testbench worked fine, but not via PHPStan. Turns out Testbench was not discovering packages when ran via Larastan.

I'm opening this as a draft PR, as I'm not sure how it can be tested yet, or if it should be presented as a configuration option instead of on by default. Open to criticism and discussion.

**Breaking changes**

The change didn't cause any tests to fail. There were two failing before which are still failing now (locally on 8.0 & 8.1):

####  `Tests\Rules\CheckDispatchArgumentTypesCompatibleWithClassConstructorRuleTest::testJobDispatch`
```
TypeError: PHPStan\Rules\FunctionCallParametersCheck::__construct(): Argument #5 ($propertyReflectionFinder) must be of type PHPStan\Rules\Properties\PropertyReflectionFinder, bool given
```

#### `Tests\Rules\CheckDispatchArgumentTypesCompatibleWithClassConstructorRuleTest::testEventDispatch`
```
TypeError: PHPStan\Rules\FunctionCallParametersCheck::__construct(): Argument #5 ($propertyReflectionFinder) must be of type PHPStan\Rules\Properties\PropertyReflectionFinder, bool given
```

**PS**
Thanks for your hard work on this!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+* Automatically discover packages via `Orchestra\Testbench`
+
 ## [1.0.2] - 2021-11-23
 
 ### Added
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -19,7 +19,7 @@ final class ApplicationResolver
 
     /** @var mixed */
     public static $composer;
-    
+
     /** @var bool */
     protected $enablesPackageDiscoveries = true;
 
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -20,7 +20,7 @@ final class ApplicationResolver
     /** @var mixed */
     public static $composer;
     
-    /** $var bool */
+    /** @var bool */
     protected $enablesPackageDiscoveries = true;
 
     /**
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -20,7 +20,8 @@ final class ApplicationResolver
     /** @var mixed */
     public static $composer;
     
-    protected bool $enablesPackageDiscoveries = true;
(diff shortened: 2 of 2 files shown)


mewejo:
This can be fixed by creating `bootstrap/app.php` in the packages (not Larastan) directory with the following:

```php
<?php

use Orchestra\Testbench\Concerns\CreatesApplication;

return (new class {
	use CreatesApplication;

	protected bool $enablesPackageDiscoveries = true;
})->createApplication();
```

Using an anonymous class to create the application and set the `$enablesPackageDiscoveries` property to true. I'm happy to do this if the PR isn't accepted, however I can't extend `\NunoMaduro\Larastan\ApplicationResolver` as it's final. Is there anything in there I'd benefit from? PHPStan seems happy without it.

canvural:
Hi,

I don't really understand what problem does this solve. But I guess adding this won't do any harm. We can just merge this if you mark it as ready.

mewejo:
Hi @canvural - it allows Testbench to discover packages. Useful when a package you're developing has a dependency on another packages service provider:

https://github.com/orchestral/testbench-core/blob/7.x/src/Concerns/CreatesApplication.php#L36

Thank you - I'll mark as ready :-)

szepeviktor:
@canvural Is this PR into `master` or `1.x`?

canvural:
Thank you!

crynobone:
Actually, you can write the following without using `CreatesApplication` trait.

```php
use function Orchestra\Testbench\container;

$app = container()->createsApplication();
```

mewejo:
Tidier, I'll use that if I need to change anything else, thanks @crynobone 😎

PR #1108, opened 2022-01-27
Update README.md of version 1.x

lucasgiovanny:
**Changes**

Only update the composer command on README.md from branch 1.x to match the correct version.

**Breaking changes**

None breaking changes.


canvural:
Thank you!

PR #1120, opened 2022-02-09
Do not exclude ClassGenerator

szepeviktor:


szepeviktor:
I see, @canvural you don't let me live :)

Issue #1051, opened 2021-11-25
With SoftDeletes `deleted_at` is a Carbon instance, not a string


MatanYadaev:
Larastan Version: 8.6.7
- `--level` used: max
- Pull request with failing test: https://github.com/MatanYadaev/larastan/pull/2

### Description

Larastan infers that `deleted_at` is a string and not a Carbon.

### Laravel code where the issue was found

```php
class User extends Authenticatable
{
  use SoftDeletes;

   // ...
}
```

```php
// Cannot call method addDay() on string|null.
User::onlyTrashed()->firstOrFail()->deleted_at->addDay();
```

BTW, when you add `$casts = ['deleted_at' => 'datetime'];` the issue is gone - but I believe it's not an ideal solution.

szepeviktor:
> Pull request with failing test:

Thank you for your report.
You do not need to open a PR for your issues.

szepeviktor:
Do you suggest that `onlyTrashed()` prevents `deleted_at` from being null?

MatanYadaev:
@szepeviktor 
> You do not need to open PR-s for your issues.
Sorry, what do you mean?
> Do you suggest that `onlyTrashed()` prevents `deleted_at` from being null?
It is, but it's unrelated to the issue. The issue is that Larastan doesn't cast `deleted_at` to `Carbon` by default. And in my opinion, it should do it. 

Laravel casts `created_at` and `updated_at`  to `Carbon` by default. It does the same thing to `deleted_at` (`SoftDeletes::initializeSoftDeletes`), but for some reason, Larastan doesn't understand it.

szepeviktor:
> I don't have to create a PR with failing tests when I open an issue?

Oh yess!!

szepeviktor:
Maybe here it is
https://github.com/nunomaduro/larastan/blob/7d6e564144b2f9c231d0331d4083905a1b3e872b/src/Properties/SchemaAggregator.php#L372-L378

PR #1124, opened 2022-02-10
fix: `deleted_at` column has date type rather than `string`

totov:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves https://github.com/nunomaduro/larastan/issues/1051

**Changes**

This updates the `SchemaAggregator` to read `SoftDeletes` and `SoftDeletesTz` columns as `Carbon` instances instead of strings.

### DIFF
--- a/src/Properties/ModelPropertyExtension.php
+++ b/src/Properties/ModelPropertyExtension.php
@@ -154,6 +154,22 @@ final class ModelPropertyExtension implements PropertiesClassReflectionExtension
     }
 
     /**
+     * @param Model $modelInstance
+     * @return string[]
+     * @phpstan-return array<int, string>
+     */
+    private function getModelDateColumns(Model $modelInstance): array
+    {
+        $dateColumns = $modelInstance->getDates();
+
+        if(! method_exists($modelInstance, 'getDeletedAtColumn')) {
+            return $dateColumns;
+        }
+
+        return [...$dateColumns, $modelInstance->getDeletedAtColumn()];
+    }
+
+    /**
      * @param  SchemaColumn  $column
      * @param  Model  $modelInstance
      * @return string[]
@@ -164,7 +180,7 @@ final class ModelPropertyExtension implements PropertiesClassReflectionExtension
         $readableType = $column->readableType;
         $writableType = $column->writeableType;
 
-        if (in_array($column->name, $modelInstance->getDates(), true)) {
+        if (in_array($column->name, $this->getModelDateColumns($modelInstance), true)) {
             return [$this->getDateClass().($column->nullable ? '|null' : ''), $this->getDateClass().'|string'.($column->nullable ? '|null' : '')];
         }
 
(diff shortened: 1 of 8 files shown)


BertvanHoekelen:
Nice! I think you should get the date type from `get_class(\Illuminate\Support\Facades\Date::now())` as you can have overwritten the default Date with `Date::use(CarbonImmutable::class)` in a service provider.

src/Properties/SchemaAggregator.php
```diff
@@ -372,14 +373,17 @@ private function processColumnUpdates(string $tableName, string $argName, array
...
                         case 'timestamptz':
                         case 'timetz':
                         case 'year':
-                        case 'softdeletes':
                             $table->setColumn(new SchemaColumn($columnName, 'string', true));
                             break;
 
+                        case 'softdeletestz':
+                        case 'softdeletes':
+                            $table->setColumn(new SchemaColumn($columnName, '?'.get_class(Date::now()), true));
```
canvural, line 384: I think there is no need to add the `?` there. The 3rd argument controls if the column is nullable or not.

canvural (changes requested):
Hi,

Sorry for the late response.

It looks ok. Even though I don't like the `get_class(Date::now))` call, for now it's fine I guess.

src/Properties/SchemaAggregator.php
totov: I think this PR might need a rethink - the reason the nullable parameter is ignored is because the type is not caught in the switch statement in when setting the readable and writable types later on:

https://github.com/nunomaduro/larastan/blob/6e74c6eeb0b27227c787917afd559a2d22ec241c/src/Properties/ModelPropertyExtension.php#L162-L205

Further to that, there is a bit of work on lines 167:169 above to catch model dates such as created_at and updated_at (it also figures out the correct date class to return there in `getDateClass()`).

So it may be more in line with previous work to update lines 167:169 to see if the model has the `SoftDeletes` trait, call the `getDeletedAtColumn()` method if so and then add it to the array of dates to check the provided column name against?

canvural: Yes that'd be simpler solution I guess. You can even just do `method_exists` check for `getDeletedAtColumn` I think.

totov:
I've submitted a few commits with changes to the PR in line with the earlier discussion. Let me know if I can do anything else to improve this.

src/Properties/ModelPropertyExtension.php
```diff
@@ -153,6 +153,22 @@ private function getDateClass(): string
...
+    /**
+     * @param  Model  $modelInstance
+     * @return string[]
+     * @phpstan-return array<int, string>
+     */
+    private function getModelDateColumns(Model $modelInstance): array
+    {
+        $dateColumns = $modelInstance->getDates();
+
+        if (! method_exists($modelInstance, 'getDeletedAtColumn')) {
```
canvural, line 165: I'd prefer doing a `method_exists` check here instead of `not`. So we can just add the `$modelInstance->getDeletedAtColumn()` to the `$dateColumns` array. This way we can skip array unpacking and returning new array.

totov: No problem, that's how I originally wrote it and then thought an early return may be preferable. I'll get it changed now.

canvural:
Looks good. Thank you!

PR #1136, opened 2022-02-25
feat: add return type support for `$request->user()` call

### LINKED ISSUE #1132: Rate Limiter in Fresh Laravel 9 Project Throws Error
- Larastan Version: 2.0.1
- `--level` used: max

### Description
In a fresh Laravel 9 project, I get an error in the `RouteServiceProvider` on the line regarding rate limiting.

The error is:
```
Cannot access property $id on mixed.
```

```php
protected function configureRateLimiting(): void
{
    RateLimiter::for('api', function (Request $request) {
        return Limit::perMinute(60)->by($request->user()?->id 👈 ?: $request->ip());
    });
}
```

Is there a way to remove this error?

canvural:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Fixes #1132 

**Changes**

This PR adds return type extension for `$request->user()` so the correct `User` model is resolved.

Note: In the tests we assert that it returns `Illuminate\Foundation\Auth\User` because this test does not extend our `FeaturesTest` which uses TestBench and boots up a real Laravel app. But in the user code this will resolve to correct user model.

**Breaking changes**

None.


orkhanahmadov:
@canvural  This is great that it asserts the correct model, but it also understandably asserts `null`. Is there a way to force `$request->user()` to always assert a user model? In our use case we use `$request->user()` only in places where the user is authenticated. Updating Larastan to `2.1.0` now shows a ton of errors where we access property or method of `user()`, like `$request->user()->name`.

canvural:
@orkhanahmadov No, we can't know if there will be a user or not. So the correct type is with `null`

You can narrow down the type in your code. Read [here](https://phpstan.org/writing-php-code/narrowing-types) for more information.

PR #1142, opened 2022-02-28
Date type extension

### LINKED ISSUE #421: now() helper & CarbonImmutable

bastien-phi:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #421. The code is taken from #438 and some tests were added.

I've been using initial implementation in projects for a long time and as I found a way to test it, I think it now deserves to be in this package.

**Changes**

Now, when using `Date::use(CarbonImmutable::class)`, the helpers `now()`, `today()` and models' timestamps are determined to be `CarbonImmutable`.

It also adds an extension to determine return types of `Illuminate\Support\Facade\Date` methods when a custom date Factory is used (for exemple, `Date::createFromString` returns `CarbonImmutable`)

**Breaking changes**

I don't think so


canvural:
Thank you. Sorry for the delay!

PR #1152, opened 2022-03-04
Use recent version of `composer/pcre`

orkhanahmadov:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

`composer/pcre:^1.0` is not compatible with PHP7.4 and above. According to `pcre` [docs](https://github.com/composer/pcre#requirements), version `^3.0` should be used for PHP7.4 and above, which is more relevant for Larastan.

Using `^1.0` causes conflict with the latest version of other libraries like [`PHP-CS-Fixer`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/composer.json), which requires `composer/xdebug-handler:^3.0` and that required `composer/pcre:^3.0`.

**Breaking changes**

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+* fix: Updated `composer/pcre` to new PHP relevant version
+
 ## [2.1.0] - 2022-03-04
 
 ### Added
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     "require": {
         "php": "^8.0.2",
         "ext-json": "*",
-        "composer/pcre": "^1.0",
+        "composer/pcre": "^3.0",
         "illuminate/console": "^9",
         "illuminate/container": "^9",
         "illuminate/contracts": "^9",


szepeviktor:
> `composer/pcre:^1.0` is not compatible with PHP7.4 and above.

Hello Orkhan! 👋 

Our CI installs `  - Locking composer/pcre (1.0.1)` on PHP 8.0
How could it be?

szepeviktor:
kép

I see!! :)
This is the list of **minimum** requirements not maximum.

szepeviktor (approved):

orkhanahmadov:
Hi @szepeviktor,

Sorry, I did not phrase the problem correctly. You are right, `composer/pcre:^1.0` **is compatible** with with PHP8.0 (v1 [composer.json](https://github.com/composer/pcre/blob/1.x/composer.json#L20)). But other libraries like `PHP-CS-Fixer` require `composer/pcre:^3.0`. Currently, if a project uses the latest version of `PHP-CS-Fixer` it is not possible to use `nunomaduro/larastan:^2.1`, because of `composer/pcre` change.

- One option is to use `composer/pcre:^3.0`which is the latest and more relevant version for Larastan.
- The second option would be allowing other versions of `composer/pcre` with `^1.0|^2.0|^3.0`

I think first option suits better.

canvural:
Thank you.

PR #1156, opened 2022-03-07
feat: support Laravel 9 style attribute accessors and mutators

### LINKED ISSUE #1153: Laravel 9 Casts/Attribute handling
- Larastan Version: 2.1.1
- `--level` used: 5
- Pull request with failing test: Not tech-savvy enough for that

### Description

The new approach in Laravel 9 to handle model properties' accessors and/or mutators, using the class `Illuminate\Database\Eloquent\Casts\Attribute`, needs proper handling. Right now, all such properties are interpreted as strings.

Create a model with a string/varchar column where you store space-separated list of strings (example: scopes of an OAuth token). Then, add the Attribute cast to the model, something like this:

```php
	/**
	 * The list of scopes is stored as a space-separated string.
	 */
	public function scopes() : Attribute
	{
		return Attribute::make(
			get: fn(?string $value) => is_null($value) ? [] : explode(' ', $value),
			set: function(array $value) {
				$set = array_unique($value);
				sort($set);
				return ['scopes' => implode(' ', $set)];
			}
		);
	}
```

Then, trying to use the accessor value in something like `in_array('x', $model->scopes)` raises the error 

```
Parameter #2 $haystack of function in_array expects array, string given. 
```

And trying to set the value to an array, for example `$model->scopes = ['x']`, raises the error

```
Property App\Models\CustomModel::$scopes (string) does not accept array<int, string>.
```

canvural:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Resolves #1153 

**Changes**

This PR adds support for Laravel 9 style attribute accessors and mutators.

Does this by making the `Attribute` class generic, and then reading those generic types when the property is accessed.

**Breaking changes**

### DIFF
--- a/src/Properties/ModelAccessorExtension.php
+++ b/src/Properties/ModelAccessorExtension.php
@@ -4,11 +4,15 @@ declare(strict_types=1);
 
 namespace NunoMaduro\Larastan\Properties;
 
+use Illuminate\Database\Eloquent\Casts\Attribute;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Str;
 use PHPStan\Reflection\ClassReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Reflection\PropertiesClassReflectionExtension;
 use PHPStan\Reflection\PropertyReflection;
+use PHPStan\Type\Generic\GenericObjectType;
+use PHPStan\Type\ObjectType;
 
 /**
  * @internal
@@ -21,6 +25,28 @@ final class ModelAccessorExtension implements PropertiesClassReflectionExtension
             return false;
         }
 
+        $camelCase = Str::camel($propertyName);
+
+        if ($classReflection->hasNativeMethod($camelCase)) {
+            $methodReflection = $classReflection->getNativeMethod($camelCase);
+
+            if ($methodReflection->isPublic() || $methodReflection->isPrivate()) {
+                return false;
+            }
+
+            $returnType = ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+
+            if (! $returnType instanceof GenericObjectType) {
+                return false;
+            }
+
+            if (! (new ObjectType(Attribute::class))->isSuperTypeOf($returnType)->yes()) {
+                return false;
+            }
+
+            return true;
+        }
+
         return $classReflection->hasNativeMethod('get'.Str::studly($propertyName).'Attribute');
     }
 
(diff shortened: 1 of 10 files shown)


src/Properties/ModelPropertyExtension.php
```diff
@@ -276,4 +279,29 @@ private function castPropertiesType(Model $modelInstance): void
...
+            $returnType = ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+
+            if ($methodReflection->isPublic()) {
+                return false;
+            }
+
+            if (! (new ObjectType(Attribute::class))->isSuperTypeOf($returnType)->yes()) {
+                return false;
+            }
+        }
```
szepeviktor, line 303: @canvural Why have this block? The returned value is `false` anyways.

canvural: @szepeviktor Oh, thanks! It should  return true at the end, not false.

szepeviktor: You're welcome.

src/Properties/ModelAccessorExtension.php
```diff
@@ -21,13 +25,50 @@ public function hasProperty(ClassReflection $classReflection, string $propertyNa
...
+
+            if ($methodReflection->isPublic() || $methodReflection->isPrivate()) {
+                return false;
+            }
+
+            $returnType = ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+
+            if (! $returnType instanceof GenericObjectType) {
+                return false;
+            }
```
CalebDW, line 41: @canvural, thanks for working on this! However, this is causing all of my L9 accessors (i.e., attributes that are not mutating a table column) to fail with an 'undefined property' error and it took me nearly a whole day to figure out why---I didn't know it needed the `@return Attribute` generic.

```php
    /** @return Attribute<string, never> */        <----- UNDOCUMENTED
    protected function someVar(): Attribute
    {
        return Attribute::make(
            get: fn (): string => 'fooBar',
        );
    }
```

Also, I'm still trying to get PHPStan `Level 1` to pass which is checking the following:

> Level 1: possibly undefined variables, unknown magic methods and properties on classes with __call and __get

but fails on attributes missing the generic which seems unnecessarily strict as `Level 1` is just concerned about the property *existence* and not the type. I would expect this to fail around `Level 6: report missing typehints`.

Would it be possible to only enforce the generic check at a higher level?

Thanks!

canvural: @CalebDW I agree we don't have enough documentation of the features here. This Attribute change was only in release notes I guess.

> Also, I'm still trying to get PHPStan Level 1 to pass which is checking the following:
	Level 1: possibly undefined variables, unknown magic methods and properties on classes with __call and __get
but fails on attributes missing the generic which seems unnecessarily strict as Level 1 is just concerned about the property existence and not the type. I would expect this to fail around Level 6: report missing typehints.
	Would it be possible to only enforce the generic check at a higher level?

Larastan does not have any influence on error levels. Those come from PHPStan. Which error exactly you are referring to?

CalebDW: @canvural,

The error I've been getting for Attributes without the `@return Attribute<>` generic is:

```
Access to an undefined property Path\To\Model::$foo_bar.
```

But I don't think this makes sense because Larastan can detect that the attribute exists without the generic PHPdoc right? You just wouldn't have any type information about it.

canvural: The generic return type is needed. That's how Larastan recognizes that this method is an attribute.

CalebDW: But `ModelPropertyExtension::hasAttribute` uses the following check to determine if a property is an Attribute (as does this same method).

```php
            if (! (new ObjectType(Attribute::class))->isSuperTypeOf($returnType)->yes()) {
                return false;
            }
```

Is this not enough to determine existence?

PR #1158, opened 2022-03-07
feat: add exact typehints for `Model` `casts`, `hidden` and `visible` properties

Magiczne:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Hi! When using larastan inside project in which rector is present and model classes are heavily using `$casts`  or `$hidden` attributes there is a problem with exact typehints for those properties.
First solution is to ignore rector rule that adds typehints, the second is to create stub file for Model class.

Since I've been using the stub for some time now I've decided to make this PR.

**Breaking changes**

### DIFF
--- a/stubs/Model.stub
+++ b/stubs/Model.stub
@@ -8,6 +8,21 @@ namespace Illuminate\Database\Eloquent;
 abstract class Model implements \JsonSerializable, \ArrayAccess
 {
     /**
+     * @var array<string, string|class-string>
+     */
+    protected $casts = [];
+
+    /**
+     * @var array<int, string>
+     */
+    protected $hidden = [];
+
+    /**
+     * @var array<int, string>
+     */
+    protected $visible = [];
+
+    /**
      * Update the model in the database.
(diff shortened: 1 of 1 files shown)


stubs/Model.stub
```diff
@@ -7,6 +7,21 @@ namespace Illuminate\Database\Eloquent;
  */
 abstract class Model implements \JsonSerializable, \ArrayAccess
 {
+    /**
+     * @var array<string, string|class-string>
```
canvural, line 11: `string|class-string` is just `string` Can you change this to `string`?

canvural (changes requested):

stubs/Model.stub
szepeviktor:
```suggestion
     * @var array<string, string>
```

Magiczne: Sure. Will handle it at the evening

canvural:
Thank you!

PR #1203, opened 2022-03-29
When unless void default

BertvanHoekelen:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

The default callback for when and unless can can return void.


stubs/EloquentBuilder.stub
```diff
@@ -437,7 +437,7 @@ class Builder
      *
      * @param  mixed  $value
      * @param  callable($this, mixed): (void|Builder<TModelClass>)  $callback
-     * @param  callable($this, mixed): (null|Builder<TModelClass>)|null  $default
+     * @param  callable($this, mixed): (null|void|Builder<TModelClass>)|null  $default
```
szepeviktor, line 440: Isn't it void in place of null?
```suggestion
     * @param  callable($this, mixed): (void|Builder<TModelClass>)|null  $default
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 * feat: add support for the `cursorPaginate()` paginator
+* fix: when/unless default values can return void
```
szepeviktor, line 10:
```suggestion
* fix: when/unless callback for default values can return void
```

stubs/EloquentBuilder.stub
BertvanHoekelen: Yeah, I was thinking the same. But might be a breaking change if people are already return null now. Let me know If I need to change it.

szepeviktor: @canvural I'm out. The framework uses type juggling: `elseif ($default)`.
Please lead the way!

szepeviktor: @BertvanHoekelen In PHP reality void is really null 😢 
```php
function noVoidInPhp(): void
{
    echo 'In PHP reality void is really: ';
}

var_dump(noVoidInPhp());
```

szepeviktor: Sorry, I'm paralyzed by code-liberalism :)

canvural: In this case I think it's better to have something like `(callable($this, mixed): Builder<TModelClass>)|(callable($this, mixed): void)` to say "this callable can return either `void` or `Builder`"

Can you try that? Thanks.

BertvanHoekelen: @canvural Done 💪🏼

szepeviktor (approved):

canvural:
Thank you. Sorry for the delay!

Issue #1204, opened 2022-03-30
Interaction between Mockery and Larastan


mad-briller:
- Larastan Version: 2.1.3
- `--level` used: 8

### Description

Larastan has a weird interaction with [Mockery](http://docs.mockery.io/en/latest/) and https://github.com/phpstan/phpstan-mockery where it thinks the `Expects` type is an eloquent model, due to the use of the method name `with` to register method argument expectations (i think).

It causes lots of these false positives as a result:

```
  53     Relation 'route' is not found in PHPStan\Mockery\Type\Expects model.    
```

I'm not sure how to create a pull request for that as i would need to install the `phpstan/phpstan-mockery` package to create a reproducible test case, probably not something that you want shipped by default.

However i forked larastan and did the following to recreate:

1. `composer require phpstan/phpstan-mockery`
2. add the following lines to the `./extension.neon`
```neon
includes:
  - ./vendor/phpstan/phpstan-mockery/extension.neon
```
3. add the following lines to the `./tests/Rules/Data/relation-existence-rules.php`
```php
$mock = \Mockery::mock(\StdClass::class);

$mock->expects('foo')->with('bar');
```
4. `php vendor/bin/phpunit tests/Rules/RelationExistenceRuleTest.php`
```
PHPUnit 9.5.19 #StandWithUkraine

Runtime:       PHP 8.1.3

F                                                                   1 / 1 (100%)

Time: 00:01.251, Memory: 88.00 MB

There was 1 failure:

1) Tests\Rules\RelationExistenceRuleTest::testRule
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
 50: Relation 'foo' is not found in App\User model.
 51: Relation 'foo' is not found in App\User model.
 52: Relation 'foo' is not found in App\User model.
+56: Relation 'bar' is not found in PHPStan\Mockery\Type\Expects model.
```

szepeviktor:
Here you go Bred! #1205
PR-s are free.

mad-briller:
i've had a look but unfortunately i don't understand phpstans type system well enough yet and i don't have a lot of time to spare at the moment

it seems this block in the `ModelRuleHelper`:

```
        if ((new ObjectType(Builder::class))->isSuperTypeOf($type)->no() &&
            (new ObjectType(EloquentBuilder::class))->isSuperTypeOf($type)->no() &&
            (new ObjectType(Relation::class))->isSuperTypeOf($type)->no() &&
[...]

PR #1205, opened 2022-03-30
See Mockery

szepeviktor:
Live demo of #1204


szepeviktor:
The package's README says:

> shouldReceive(), allows() and expects() methods can be called on the mock object and they work as expected.

PR #1206, opened 2022-03-30
Interaction with phpstan/phpstan-mockery

mad-briller:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Resolves https://github.com/nunomaduro/larastan/issues/1204

Inverts the check in the `ModelRuleHelper` class to reduce false positives of classes that return `maybe` for all super class checks.

So i am just creating a PR based on the `mockery` branch provided by @szepeviktor but i noticed these lines added to the `extension.neon`:

```neon
includes:
  - ./vendor/phpstan/phpstan-mockery/extension.neon
```

Wont that mean that everyone who uses Larastan will now need the `phpstan/phpstan-mockery` dependency? 

Also the `.github/workflows/tests.yml` path seems to have been changed only to run this one specific test, is that correct? 
sorry you'll have to forgive my naivety with github actions, pretty new to them!

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,11 +42,13 @@ jobs:
       - name: "Install dependencies"
         run: "composer update --with='illuminate/contracts:${{ matrix.laravel }}' --prefer-dist --no-interaction --no-progress"
 
+      - run: "composer require --dev phpstan/phpstan-mockery"
+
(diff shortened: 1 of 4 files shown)


szepeviktor:
Please open a PR with changes only for `src/Rules/ModelRuleHelper.php` 😆

PR #1207, opened 2022-03-30
fix: inverted check in the ModelRuleHelper

mad-briller:
Resolves https://github.com/nunomaduro/larastan/issues/1204

Inverting this check minimizes false positives for classes that return 'maybe' for the isSuperTypeOf check, which would then go on to be treated as a queryable Laravel class.

This breaks no test cases currently present.


mad-briller:
thanks for the guidance

PR #1219, opened 2022-04-26
Patch factory

lptn:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Add more type info to Factory class.

**Breaking changes**

- nothing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+* feat: improve model factory `state` method by @lptn in https://github.com/nunomaduro/larastan/pull/1219
 
 ## [2.1.4] - 2022-03-30
 
--- a/stubs/Factory.stub
+++ b/stubs/Factory.stub
@@ -18,7 +18,7 @@ class Factory
      * Get a new factory instance for the given attributes.
      *
      * @param  callable|array<model-property<TModel>, mixed>  $attributes
-     * @return static
+     * @return static<TModel>
      */
     public static function new($attributes = []) {}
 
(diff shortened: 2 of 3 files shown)


lptn:
>See few of the recent tests. Just one file with asserts.

done by [8a849db](https://github.com/nunomaduro/larastan/pull/1219/commits/8a849db410a848ce158db6a1405e38aaeb60a187)

canvural:
Hi,

If it's a copy paste from Laravel, why do we need it here?

lptn:
@canvural 
>If it's a copy paste from Laravel, why do we need it here?

I don't know conventions on this project, guys. Should I just omit doc for such @param?

szepeviktor:
Failing test: https://github.com/nunomaduro/larastan/blob/master/tests/Features/ReturnTypes/ModelFactory.php#L20
The return type is different.

canvural:
@lptn If it has correct docblocks in Laravel, we don't need to add it here at all. Stubs are only for overriding wrong docblocks.

szepeviktor:
> In my case I just needed to override `@return`, but I copy-pasted the whole comment for the method (inc. `@param` and text description). What I supposed to do in this case, just add `@return`?

Please create complete docblocks.

lptn:
@szepeviktor

Actually the type is valid, by PHPStan detect only a parent type. I'm new to PHPStan, I assume the issue is here: https://github.com/nunomaduro/larastan/blob/6a03ad7899f5b71133c459df0260e19491dc5ed4/src/ReturnTypes/ModelFactoryDynamicStaticMethodReturnTypeExtension.php#L55

Can you please help me to fins a solution for this problem? Or at least help me to find a direction to look for.


```json
{
    "message": "Method Tests\\Features\\ReturnTypes\\ModelFactory::testNew() should return Database\\Factories\\PostFactory but returns Illuminate\\Database\\Eloquent\\Factories\\Factory<App\\Post>.",
    "line": 75,
    "ignorable": true
},
```


Thanks!

szepeviktor:
I cannot answer but one of your modifications - maybe `Factory:new` - causes this test to fail.
https://github.com/nunomaduro/larastan/blob/486a0d5f81539a46707456f44ea0d1f281556af1/tests/Features/ReturnTypes/ModelFactory.php#L73-L76

PR #1220, opened 2022-04-26
Improve DX: specify memory limits for tests (custom composer scripts)

lptn:
explicitly specify (most likely increase) memory limits for tests on the project. Especially useful for dev env where PHP configured with strict memory limits (to catch memory issues locally), so developers will need to do less frictions to run tests locally.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,8 +65,8 @@
         }
     },
     "scripts": {
-        "test:types": "phpstan analyse --ansi --memory-limit 2G",
-        "test:unit": "phpunit --colors=always -d memory_limit=2048M",
+        "test:types": "phpstan analyse --ansi --memory-limit 256M",
+        "test:unit": "phpunit --colors=always -d memory_limit=512M",
         "test": [
             "@test:types",
             "@test:unit"
--- a/composer.json
+++ b/composer.json
@@ -65,8 +65,8 @@
         }
     },
     "scripts": {
-        "test:types": "phpstan analyse --ansi",
-        "test:unit": "phpunit --colors=always",
+        "test:types": "phpstan analyse --ansi --memory-limit 2G",
+        "test:unit": "phpunit --colors=always -d memory_limit=2048M",
         "test": [
             "@test:types",
             "@test:unit"


canvural:
Hi,

Thanks for this. But 2G for PHPStan is too much. Thanks to the recent changes in PHPStan 1.6.0 and bleeding edge, locally I can set the memory limit to 200M and it runs fine.

Can you test and lower that limit?

Also I don't think a memory limit for PHPUnit is necessary. But again, can you try to adjust and find a lower limit that both woks on CI and locally? Thanks.

lptn:
@canvural 

well, it depends on how to do you want to archive:
a) smooth DX
b) another reporting system on high memory consumption


I assumed you have chosen b) option. OK, let's go tis way.

## PHPUnit

`phpunit --colors=always -d memory_limit=512M` is working option, 256M - not (at least on my LM: PHP 8.1 on MacOS with few popular php ext, nothing special). I haven't explored options in between 256M and 512M, I think it's not such important (pls correct me if I'm wrong).

## PHPStan

`phpstan analyse --ansi --memory-limit 256M` is working option, 256M - not.


## Conclusion
It looks to me that these options can be a good start:

```json
{
        "test:types": "phpstan analyse --ansi --memory-limit 256M",
        "test:unit": "phpunit --colors=always -d memory_limit=512M",
}
```

of course, there is a chance that some potential contributors will not run tests and invest their time if tests will fail for memory limit reasons, that's why it's important to not be very greedy for these limits and thus be open to easily increase them on demand.

If you guys agree, I will commit these changes

mfn:
What is "LM"?

PS: I agree, I've had the same experience with another project: the PHP OSX default CLI has a limit of 128MB; I have not experiences this with phpunit [but phpstan](https://github.com/rebing/graphql-laravel/blob/dbda3f80ff8c51660f702f3a7bd3f1dd3d5edbb8/composer.json#L68-L69)

lptn:
>What is "LM"?

sorry, local machine

canvural:
Thank you.

PR #1221, opened 2022-04-28
fix: fix: change `QueryBuilder::newQuery()` @return from `$this` to `static`

dotdash:
newQuery(), as the name implies, creates a new instance, so the return
type should not be `$this`, but `static` instead.


szepeviktor:
https://github.com/laravel/framework/blob/v9.9.0/src/Illuminate/Database/Query/Builder.php#L3437

canvural:
Thank you.

PR #1222, opened 2022-04-28
feat: add dynamic return type extension for `App::make`

PrinsFrank:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Add dynamic return type extension for App::make the same way resolve() and app() are supported.

**Breaking changes**

None


canvural:
Thank you.

PR #1226, opened 2022-05-03
feat: rules for unnecessary uses of with() and values() functions

PrinsFrank:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**
- feat: add Rule checking for useless 'with()' function calls
- feat: add Rule checking for useless 'value()' function calls

**Breaking changes**
None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,8 +54,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Improvements
 * feat: add cursorPaginate() support by @BramVanBerkel in https://github.com/nunomaduro/larastan/pull/1198
 * feat: add App::make() support by @PrinsFrank in https://github.com/nunomaduro/larastan/pull/1222
-* feat: add Rule checking for useless 'with()' function calls by @PrinsFrank
-* feat: add Rule checking for useless 'value()' function calls by @PrinsFrank
+* feat: add Rule checking for useless 'with()' function calls by @PrinsFrank in https://github.com/nunomaduro/larastan/pull/1226
+* feat: add Rule checking for useless 'value()' function calls by @PrinsFrank in https://github.com/nunomaduro/larastan/pull/1226
 
 ### Fixes
 * fix: inverted check in the ModelRuleHelper by @mad-briller in https://github.com/nunomaduro/larastan/pull/1207
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Improvements
 * feat: add cursorPaginate() support by @BramVanBerkel in https://github.com/nunomaduro/larastan/pull/1198
 * feat: add App::make() support by @PrinsFrank in https://github.com/nunomaduro/larastan/pull/1222
+* feat: add Rule checking for useless 'with()' function calls by @PrinsFrank
 
 ### Fixes
 * fix: inverted check in the ModelRuleHelper by @mad-briller in https://github.com/nunomaduro/larastan/pull/1207
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * feat: add cursorPaginate() support by @BramVanBerkel in https://github.com/nunomaduro/larastan/pull/1198
 * feat: add App::make() support by @PrinsFrank in https://github.com/nunomaduro/larastan/pull/1222
 * feat: add Rule checking for useless 'with()' function calls by @PrinsFrank
+* feat: add Rule checking for useless 'value()' function calls by @PrinsFrank
 
 ### Fixes
 * fix: inverted check in the ModelRuleHelper by @mad-briller in https://github.com/nunomaduro/larastan/pull/1207
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
-* fix: Resolve correct model factory instance when application namespace is empty
+
+### Improvements
+* feat: add Rule checking for useless 'with()' and 'value()' function calls by @PrinsFrank in https://github.com/nunomaduro/larastan/pull/1226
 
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Improvements
 * feat: add cursorPaginate() support by @BramVanBerkel in https://github.com/nunomaduro/larastan/pull/1198
 * feat: add App::make() support by @PrinsFrank in https://github.com/nunomaduro/larastan/pull/1222
+* feat: add Rule checking for useless 'with()' function calls by @PrinsFrank in https://github.com/nunomaduro/larastan/pull/1226
+* feat: add Rule checking for useless 'value()' function calls by @PrinsFrank in https://github.com/nunomaduro/larastan/pull/1226
```
szepeviktor, line 19: You could merge these lines.

PrinsFrank: Thanks for looking at this so quickly!

szepeviktor:
[non-official] I think you won't get away with these rules enabled by default.

PrinsFrank:
> [non-official] I think you won't get away with these rules enabled by default.

@szepeviktor What would you suggest? Adding these as rules that can be enabled using configuration?
I wouldn't say these are opinionated rules as they detect code that doesn't do anything.

szepeviktor:
> I wouldn't say these are opinionated rules as they detect code that doesn't do anything.

I'm just a janitor here!

src/Rules/UselessConstructs/NoUselessValueFunctionCallsRule.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public function processNode(Node $node, Scope $scope): array
+    {
+        /** @var FuncCall $node */
+        if (! $node->name instanceof Node\Name) {
+            return [];
+        }
+
+        if ($node->name->toString() !== 'value') {
```
canvural, line 29: Please use `strtolower` here for left handside.

src/Rules/UselessConstructs/NoUselessValueFunctionCallsRule.php
```diff
@@ -0,0 +1,40 @@
...
+            return [];
+        }
+
+        $args = $node->getArgs();
+        if ($args[0]->value instanceof Node\Expr\Closure) {
```
canvural, line 34: We need to check if `count($args)` is greater than 1 before here.

Also using the `instanceof` check here with the `Closure` is wrong here. This wouldn't work for example union or intersection types. Instead we should ask the type system about this. You can use `$scope->getType($args[0]->value)` to get the type for the variable. And use `isSuperTypeOf` to compare the types. You can read more about the type system [here.](https://phpstan.org/developing-extensions/type-system)

src/Rules/UselessConstructs/NoUselessValueFunctionCallsRule.php
```diff
@@ -0,0 +1,40 @@
...
+            return [];
+        }
+
+        return ["Calling the helper function 'value()' without a closure as the first argument simply returns the first argument without doing anything"];
```
canvural, line 38: Please use `RuleErrorBuilder` to construct the error. You can check other rules to see how it is done.

src/Rules/UselessConstructs/NoUselessWithFunctionCallsRule.php
```diff
@@ -0,0 +1,44 @@
...
+        if ($node->name->toString() !== 'with') {
+            return [];
+        }
+
+        $args = $node->getArgs();
+        if (count($args) === 1) {
+            return ["Calling the helper function 'with()' with only one argument simply returns the value itself. if you want to chain methods on a construct, use '(new ClassName())->foo()' instead"];
+        }
+
+        if ($args[1]->value instanceof Node\Expr\Closure || ($args[1]->value instanceof Node\Scalar\String_ && function_exists($args[1]->value->value))) {
```
canvural, line 38: `$args` is not guaranteed to have index `1` here. Please do a check before.

Also, same as the previous comment; please use the type system to do the checks here.

src/Rules/UselessConstructs/NoUselessWithFunctionCallsRule.php
```diff
@@ -0,0 +1,44 @@
...
+            return [];
+        }
+
+        return ["Calling the helper function 'with()' without a closure as the second argument simply returns the value without doing anything"];
```
canvural, line 42: ditto. `RuleErrorBuilder` to construct the message.

tests/Rules/UselessConstructs/NoUselessValueFunctionCallsRuleTest.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rules\UselessConstructs;
+
+use Tests\RulesTest;
+
+class NoUselessValueFunctionCallsRuleTest extends RulesTest
```
canvural, line 9: Please extend `RuleTestCase` here instead of `RulesTest` And change the rest of the class according to it.

tests/Rules/UselessConstructs/NoUselessWithFunctionCallsRuleTest.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rules\UselessConstructs;
+
+use Tests\RulesTest;
+
+class NoUselessWithFunctionCallsRuleTest extends RulesTest
```
canvural, line 9: ditto. extend `RuleTestCase`

canvural (changes requested):

canvural:
Hi, are you gonna finish this?

PrinsFrank:
Sorry, was preoccupied with a job switch. Will finish this now!

szepeviktor (approved):

PrinsFrank:
@szepeviktor @canvural @nunomaduro I'm cleaning up my forks, can I assume this is not going to get merged and close this PR?

szepeviktor:
@PrinsFrank Please do not delete your fork.

We _will_ take a look at it and merge it.

PrinsFrank:
@szepeviktor Ok, np!

canvural:
Thank you! I pushed some commits to simplify some bits.

PR #1232, opened 2022-05-14
feat: Auth::user() returns all auth models when no parameter specified

### LINKED ISSUE #1230: `FormRequest::user()`always returns the model of default guard
- Larastan Version: 2.1.4
- `--level` used: 9

### Description

It seems like `FormRequest::user()` method is stated to return the model of default guard, even though multiple guards defined in the `config/auth.php`.

Providing the name of the (non-default) guard with the method argument won't help; the method is still predicted to return the default.

Let's say the default guard is `default_user` and model is `\App\Models\DefaultUser`.

```php
class ExampleRequest extends FormRequest
{
    public function authorize(): bool
    {
        return $this->user() instanceof \App\Models\AnotherUser;
        // or return $this->user('another_user') instanceof \App\Models\AnotherUser;
    }
}
```

It gives me the following error:
```
Instanceof between App\Models\DefaultUser|null and App\Models\AnotherUser will always evaluate to false.
```

mirucon:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1230

**Changes**

Currently when no $guard parameter is passed to `Auth::user()` method, it always returns the model of *default* guard.
However this behaviour doesn't match how the method actually works as it might return non-default models as well.
This PR corrected the `Auth::user()` method to return all possible auth models.

p.s. I had no idea what `src/Methods/Pipes/Auths.php` is for, so please kindly let me know how I should handle this one.

**Breaking changes**

Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixes
 
 * fix: Change QueryBuilder::newQuery() @return from `$this` to `static`
+* feat: Auth::user() returns all possible auth models when no $guard parameter specified by @mirucon in https://github.com/nunomaduro/larastan/pull/1232
 
 ## [2.1.4] - 2022-03-30
 
--- a/src/Concerns/LoadsAuthModel.php
+++ b/src/Concerns/LoadsAuthModel.php
@@ -4,21 +4,53 @@ declare(strict_types=1);
 
 namespace NunoMaduro\Larastan\Concerns;
 
+use function array_key_exists;
 use Illuminate\Config\Repository as ConfigRepository;
+use function in_array;
+use function is_array;
 
 trait LoadsAuthModel
 {
-    /** @phpstan-return class-string|null */
-    private function getAuthModel(ConfigRepository $config, ?string $guard = null): ?string
+    /** @phpstan-return array<int, class-string> */
+    private function getAuthModels(ConfigRepository $config, ?string $guard = null): array
     {
         if (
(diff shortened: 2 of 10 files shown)


szepeviktor:
Hello @mirucon!

I think `empty()` is not out friend, it is way too complex.

mirucon:
Hi @szepeviktor thanks for your response!
I replaced `empty()` with `count()` method, I hope this is what you mean.

canvural:
Hi,

Can you explain how `Auth::user()` would return non-default user model? [Looks like](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Auth/AuthManager.php#L64) it chooses the default guard if you don't pass any guard name.

calebdw:
@canvural, I'm running into this issue as well.

The default guard is overridden by Laravel when setting the current guard to use.

https://github.com/laravel/framework/blob/0c574d42e9c7d82a0bd304a5cb2ce031a914bfa6/src/Illuminate/Auth/AuthManager.php#L201-L214

I have a project with several guards and use the `shouldUse` method in a middleware depending on the route. But of course Larastan doesn't know this.

Depending on the location in the code, `Auth::user()` returns a different auth model.

calebdw:
@nunomaduro, can you  please reconsider this? This functionality is exactly how the `Auth::user()` method should work because Laravel overwrites the default guard at runtime. Calling `Auth::user()` in different locations in my codebase results in different auth models being returned.

nunomaduro:
We've closed the `master` branch. Please open a new PR and target 2.x branch if you wish to re-open this PR.

PR #1234, opened 2022-05-16
Fix Model property types

szepeviktor:
These are simple lists,
```
[
  1 => 'apple',
  3 => 'pear',
]
```
is invalid!


szepeviktor:
> Can you change it to `list` here?

Done.

canvural:
Thank you.

PR #1258, opened 2022-06-02
fix: Type hinting for `having()` and `orHaving()` QueryBuilder methods

Maxoulak:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

Fixes type hinting for `having()` and `orHaving()` QueryBuilder methods, due to [this PR on laravel/framework](https://github.com/laravel/framework/pull/42593)


szepeviktor:
Thank you @Maxoulak!

What does Laravel _say_ when you pass other types to these methods?

ClementYproximite (approved):

Maxoulak:
@szepeviktor I'm not sure to understand your question, you mean if we pass an object, an array or whatever to these methods ?

Honestly I didn't try it, I made this change to avoid phpstan error when we pass an `int` or a `float` as the value parameter. But maybe we should type hinting `string|int|float|null` instead of `mixed` ?

Maxoulak:
Well, I made some tests.
If `$value` is an array (or `$operator` if `$value` is `null`), Laravel use this to retrieve the final value :
```php
head(Arr::flatten($value));
```

So, with this array : `['foo' => '2022-06-02', 1 => '2021-01-01']`, the value will be equal to "2022-06-02".

However, if `$value` is an object, an exception is thrown : "PHP Error:  Object of class stdClass could not be converted to string in /var/www/html/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php on line 66".

Imo, the most appropriate type hint should be `string|int|float|null`, because we should not pass an array as a single value.

Are you ok with that ? If yes, I'll probably first do another PR on Laravel before merging this one.

szepeviktor:
> because we should not pass an array as a single value.

Thank you for your experiment!

szepeviktor (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixes
```
szepeviktor, line 9:
```suggestion
### Fixed
```

canvural:
If the types are gonna be same, there is no point to have a stub here. I don't even remember why were they added at the first place here.

So, if you are happy with the types coming from Laravel, we can remove those methods from our stubs. Please adjust this PR accordingly, if you agree. Thanks.

szepeviktor:
Thank you @Maxoulak!

PR #1260, opened 2022-06-02
feat: adds conditional return types to some global helper functions

canvural:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

This PR adds conditional return types to some global helper functions. Replaces some of the existing extensions we have.

**Breaking changes**
n/a


paras-malhotra:
@canvural, I believe with this PR, it is not possible to change stubs. So, if I have a `Request.stub` file in my project, it will still pick up the `Request.stub` file from Larastan.

Related: #1265

szepeviktor:
Could you share your stub and the relevant config lines?

paras-malhotra:
Here is the stub: https://github.com/enlightn/enlightn/blob/1937be5e6bd3142dfd03de90927ddd571f16477c/stubs/Request.stub

And the config lines: https://github.com/enlightn/enlightn/blob/1937be5e6bd3142dfd03de90927ddd571f16477c/phpstan.neon#L19-L20

szepeviktor:
Is it okay if you overwrite `vendor/nunomaduro/larastan/stubs/Http/Request.stub` ?
`phpstan clear-result-cache` first!

paras-malhotra:
@szepeviktor, yes it does. Overwriting the Larastan `Request.stub` file succeeds in the latest version of Larastan

szepeviktor:
The answer is here: https://github.com/nunomaduro/larastan/issues/1265#issuecomment-1147124799

paras-malhotra:
Sure, thanks. I opened https://github.com/phpstan/phpstan/issues/7461

szepeviktor:
> I opened [phpstan/phpstan#7461](https://github.com/phpstan/phpstan/issues/7461)

`cp -f Your.stub Larastan.stub` :(

paras-malhotra:
Can we not go back to the `stubFiles` configuration in Larastan rather than using a custom `stubFilesExtension`? It will fix this change in behaviour.

szepeviktor:
@canvural can tell that.

szepeviktor:
@paras-malhotra HAve you tried pushing in your custom stabs file by `stubFilesExtension`?

paras-malhotra:
> @paras-malhotra Have you tried pushing in your custom stubs file by `stubFilesExtension`?

No, I haven't tried that yet. I am assuming you mean adding a `stubFilesExtension` service in Enlightn?

[This code](https://github.com/phpstan/phpstan-src/blob/943d95ee3c36ade6233e66b92591dac08af634dc/src/PhpDoc/DefaultStubFilesProvider.php#L38-L43) seems to just append the stub files from all extensions, so I'm not sure that would work.

canvural:
Hi,

Multiple stub file for one class were never supported by PHPStan. So in previous versions your code was working "by accident"

> Can we not go back to the stubFiles configuration in Larastan rather than using a custom stubFilesExtension?

Stubs in Larastan are growing. Using `stubFiles` we'd had to keep `extension.neon` in sync with the new stubs. I do not want to maintain that anymore. `stubFilesExtension` seemed like a good solution.

On the other hand maybe we can go back to it if [this](
https://github.com/phpstan/phpstan/issues/4342) is implemented.

PR #1283, opened 2022-06-19
Update phpstan-dev.yml

szepeviktor:
# testing...

### DIFF
--- a/.github/workflows/phpstan-dev.yml
+++ b/.github/workflows/phpstan-dev.yml
@@ -1,9 +1,9 @@
 name: "PHPStan dev"
 
 on:
-  schedule:
-    - cron: "1 0 * * *"
-
+  push: null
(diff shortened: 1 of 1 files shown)


szepeviktor:
@canvural 🌟 

> Installing phpstan/phpstan (1.7.x-dev 5410365): Extracting archive

canvural:
Huh, it's weird it works. When I tested like this it installed `dev` version of every package.

What is the advantage of changing the order though?

szepeviktor:
> What is the advantage of changing the order though?

- configure 👉🏻 does nothing
- upgrade to phpstan/phpstan:@dev 👉🏻 does nothing
- do the actual installation 👉🏻 installs packages

canvural:
Makes sense. Thank you!

PR #1285, opened 2022-06-19
Fix relation types

### LINKED ISSUE #1284: `HasOneOrMany` wrongly uses `TRelatedModel` as the type for the declaring model of the relation
`HasOneOrMany` has only a single template parameter `TRelatedModel` which is bound to the model class which the relation points to. However, `HasOneOrMany` also uses `TRelatedModel` as the type for the model class which declares the relation.

https://github.com/nunomaduro/larastan/blob/ccdd599c464f9de6983a8d5e0f36daaf324b646f/stubs/HasOneOrMany.stub#L11-L20

The declaring model model (or parent or owning model) is not the same as the related model.

_Originally posted by @nagmat84 in https://github.com/nunomaduro/larastan/discussions/1275#discussioncomment-2933818_

nagmat84:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1284

**Changes**

This PR

 - heavily adds more PHPstan annotations to the `Relation`-related classes and
 - introduces two, resp. three, new template parameters besides the already existing `TRelatedModel`:
    - `TDeclaringModel`
    - `TResult`
    - `TIntermediateModel` (only in case of the `...Through...`-relations)

Therewith it also resolves issue #1284.

_Rationale_

A relation links two models with each other which are not necessary identical. The `TRelatedModel` is the type of model to which the relaton points to. The `TDeclaringModel` is the type of model which declares the relation (i.e. the "base" of the arrow in the picture).

relation

Note that the roles `TRelatedModel` and `TDeclaringModel` are independent of the definition of the foreign key on SQL layer. In the example above, only `Post` has an attribute `topic_id` to realize a (1:*)-relation. But the relations `hasMany` and `belongsTo` live on the "object space" and each points from its declaring model to the related model.

In its most general form a `Relation` has a third template type `TResult` which _is not_ necessarily a model. Typically `TResult` is either identical to `TRelatedModel` or a `Collection<int, TRelatedModel>`. This is true for the `Has...One` and `Has...Many`-relations. The latter also shows why `TResult` must not be restricted to a sub-type of `Model`.

However, for custom relations which extend `Relation` this must not be true. For example, instead of an Eloquent collection a custom relation could use another type of container, like a `FixedSizeArray<TRelatedModel>` which provides higher efficiency if one has a (1:_n_)-relation for a fixed _n_. (See https://github.com/nunomaduro/larastan/discussions/1275#discussion-4138347 for a real-world example.) However, `TResult` could also be something like `\DateTime`. In the example above, a topic could have two relations `HasDateOfLeastRecentPost` and `HasDateOfMostRecentPost` which resolve to the date/time of the least/most recent post. In other words, the result of a relation is a value which depends on the related model(s), but is not necessarily the related model(s).

_A Remark on Notation_

I chose the term `TDeclaringModel` over `TParentModel` intentionally, although Laravel calls this model the "parent model" of the relation in most cases and the corresponding variable is called `$parent`. See 

https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Relations/Relation.php#L30-L35

However, it does not do so consistently. In `BelongsTo` the names are not used according to their usage in the code but are based on the use case. The declaring model is named the "child model" and the "parent" becomes the related model. See

https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Relations/BelongsTo.php#L62-L65

Even worse `BelongsTo` does not apply this change of names consistently, i.e. the related model and the declaring model are both called the parent model. In order to avoid strange constructs like `TParentModel $child`, I decided to use `TDeclaringModel`. Moreover, the term "declaring" class has already been used inside the Larastan sources.

This gives the following table

| Template parameter | Classes `Relation`, `HasMany`, `HasOne`, ..., `BelongsToMany` | Class `BelongsTo` (only) |
|:---------------------------|:--------------------------------------------------------------------------|:--------------------------|
| `TRelatedModel`      | `$related`                                                                         | `$related`                |
| `TDeclaringModel`   | `$parent`                                                                          | `$child`                   |

**Breaking changes**

The `Relation`-related classes now require up to four template parameters.

There should be no breaking change for ordinary users which only use the provided relations and create relations via the methods `hasMany`, `hasOne`, ..., etc. provided by [`\Illuminate\Database\Eloquent\Concerns\HasRelationships`](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php). The classes [`ModelRelationsDynamicMethodReturnTypeExtension`](https://github.com/nunomaduro/larastan/commit/427c38b0fae617c735c12f43b003856321567037#diff-af11dc0a6cb05bc6063795b995b97210c60aeee56368b84f649bbb7052210d7b) and [`RelationDynamicMethodReturnTypeExtension`](https://github.com/nunomaduro/larastan/commit/d271a3297c359308b780bc072a542a9212b79323#diff-aada7b039f78eb30ba0fb377413da4c64a98e1fbc96894f427735d4831666e86) have been adopted to bound all template parameters accordingly.

A breaking change may occur for users which implement their own relation classes and extend from `Relation` or its sub-classes. They need to bind the newly added template parameters. But I assume this group to be rather small. Otherwise someone else would already have recognized this bug earlier.

### DIFF
--- a/src/Types/RelationDynamicMethodReturnTypeExtension.php
+++ b/src/Types/RelationDynamicMethodReturnTypeExtension.php
@@ -69,7 +69,7 @@ class RelationDynamicMethodReturnTypeExtension implements DynamicMethodReturnTyp
 
         if (
             // Special case for MorphTo. `morphTo` can be called without arguments.
-            ($methodName !== 'morphTo' && $numArgs < 1 ) ||
+            ($methodName !== 'morphTo' && $numArgs < 1) ||
             // Special case for "...Through". `has...Through` must be called with a 2nd parameter for the intermediate model
             (in_array($methodName, ['hasOneThrough', 'hasManyThrough']) && $numArgs < 2)
         ) {
--- a/src/Types/RelationDynamicMethodReturnTypeExtension.php
+++ b/src/Types/RelationDynamicMethodReturnTypeExtension.php
@@ -5,8 +5,8 @@ declare(strict_types=1);
 namespace NunoMaduro\Larastan\Types;
 
 use Illuminate\Database\Eloquent\Model;
-use Illuminate\Database\Eloquent\Relations\BelongsTo;
-use Illuminate\Database\Eloquent\Relations\MorphTo;
+use Illuminate\Database\Eloquent\Relations\HasManyThrough;
+use NunoMaduro\Larastan\Methods\BuilderHelper;
 use PhpParser\Node\Expr\MethodCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionVariant;
@@ -17,16 +17,19 @@ use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
 use PHPStan\Type\Generic\GenericObjectType;
+use PHPStan\Type\IntegerType;
 use PHPStan\Type\ObjectType;
-use PHPStan\Type\StaticType;
 use PHPStan\Type\Type;
+use PHPStan\Type\TypeWithClassName;
 
 class RelationDynamicMethodReturnTypeExtension implements DynamicMethodReturnTypeExtension
 {
+    private BuilderHelper $builderHelper;
     private ReflectionProvider $provider;
 
-    public function __construct(ReflectionProvider $provider)
+    public function __construct(BuilderHelper $builderHelper, ReflectionProvider $provider)
     {
+        $this->builderHelper = $builderHelper;
         $this->provider = $provider;
     }
 
@@ -53,6 +56,9 @@ class RelationDynamicMethodReturnTypeExtension implements DynamicMethodReturnTyp
         MethodCall $methodCall,
         Scope $scope
     ): Type {
+        $methodName = $methodReflection->getName();
+        $methodArgs = $methodCall->getArgs();
+        $numArgs = count($methodArgs);
         /** @var FunctionVariant $functionVariant */
         $functionVariant = ParametersAcceptorSelector::selectSingle($methodReflection->getVariants());
         $returnType = $functionVariant->getReturnType();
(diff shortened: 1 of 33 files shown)


szepeviktor:
⚠️ We have a computer scientist onboard. 👨🏻‍🔬

src/Types/RelationDynamicMethodReturnTypeExtension.php
```diff
@@ -61,38 +67,65 @@ public function getTypeFromMethodCall(
...
+        // Opposed to other `HasOne...` and `HasMany...` methods,
+        // `HasOneThrough` and `HasManyThrough` do not extend a common
+        // `HasOneOrManyThrough` base class, but `HasOneThrough` directly
+        // extends `HasManyThrough`.
+        // This does not only violate Liskov's Substitution Principle but also
+        // has the unfortunate side effect that `HasManyThrough` cannot
+        // bind the template parameter `TResult` to a Collection, but needs
+        // to keep it unbound for `HasOneThrough` to overwrite it.
+        // Hence, if `HasManyTrough` is used directly, we must bind the
+        // fourth template parameter `TResult` here.
```
szepeviktor, line 123: @nagmat84 I encourage you to send a PR to Laravel fixing this problem.

nagmat84: I will do so. That was my initial idea. However, I cannot anticipate how long it will take for Laravel to fix that bug, so I submitted this PR nonetheless. As soon as the said bug is fixed by Laravel, I will post a new PR here, which removes the then unnecessary code.

szepeviktor:
All test has passed :)

We use FQCN in docblocks.

nagmat84:
> :warning: We have a computer scientist onboard. 👨🏻‍🔬

Why the warning sign? Are you afraid of theoretical computer scientists? :astonished: (BTW: You guessed correctly, I am indeed a theoretical computer scientist.)

> All tests have passed :)

I haven't expected less. I have run them locally first. :smirk:

> We use FQCN in docblocks.

OK. I will change that, but not today. It is getting late. I also still have to update the `CHANGELOG.md`, but I wanted to see first what you think about this PR in general.

canvural:
Hi,

Thank you for this! 

While it makes sense to have the declaring class also in the generic types, I do not really see a use case for that. Even in the stubs you only used `TDeclaringModel` in one or two methods. And one of them is protected. And I don't see any new tests that are using `TDeclaringModel`

Can you give me a small code snippet where `TDeclaringModel` is useful?

nagmat84:
> While it makes sense to have the declaring class also in the generic types, I do not really see a use case for that.

First of all it fixes this bug: https://github.com/nunomaduro/larastan/issues/1284 The bug is caused by the wrong assumption (and in consequence the wrong annotations) that the declaring model is the same as the related model.

> Even in the stubs you only used `TDeclaringModel` in one or two methods. And one of them is protected.

That is simply not true. Let alone (`BelongsTo](https://github.com/nunomaduro/larastan/pull/1285/files#diff-5a0f348eeabd4d1a937b041a96250c8e884d5aeb25fa4e8f7fca84b31bc2e9ce) uses it five times.

> Can you give me a small code snippet where `TDeclaringModel` is useful?

It becomes useful in the very moment when you are working an real projects which also define their own relations because the default relations don't suffice. In that case on has to extend `Relation`. That is also the reason why I annotated all protected methods, because they become relevant as soon as on uses custom relations.

If you need examples, just have a look here [Lychee: /app/Relations](https://github.com/LycheeOrg/Lychee/tree/enable-phpstan/app/Relations). At the moment we have to write hundreds of `phpstan-ignore` comments like in [/app/Relations/HasManyPhotos::getParent()](https://github.com/LycheeOrg/Lychee/blob/a1559e37a56d6af63693608b962937e0f06152ce/app/Relations/HasManyPhotos.php#L70-L77)

nagmat84:
I updated the change log. Where and how do I document user facing changes in order to complete task 3 out of 3?

nagmat84:
I wanted to add a test which provides an real-world example that uses `TDeclaringModel`. But I was unable to figure out how. Is this somewhere documented?

I expected that it would be sufficient to create a folder below `tests\Features}Relations` and place some file there which actually us `TDeclaringModel`. However, I noticed that those files are not analyzed when I run `composer test:types`. This only tests the code of Larastan itself which is located below `src`, but how do I

 1. add models and relations for being analyzed as part of the tests
 2. run these tests locally before I push a new commit here?

szepeviktor:
> 1. add models and relations for being analyzed as part of the tests

Here is how
https://github.com/nunomaduro/larastan/blob/ab0cca9384605d5af0ac4170230849113b8344c5/tests/Type/GeneralTypeTest.php#L16

nagmat84:
I updated all PHPDoc comments to contain FQCN as requested by [this comment](https://github.com/nunomaduro/larastan/pull/1285#issuecomment-1159783587).

I also added a test case to illustrate why `TDeclaringModel` is required. However, in doing so I discovered a problem. It seems as if a similar problem to the one reported in https://github.com/nunomaduro/larastan/issues/1289 is lurking somewhere. I get the report

```
Call to an undefined method Builder<TRelatedModel of Model>::...
```

three times. Obviously, the methods `orderBy`, `union` and `newQuery` are defined. In https://github.com/nunomaduro/larastan/issues/1289 @canvural pointed out in [this post](https://github.com/nunomaduro/larastan/issues/1289#issuecomment-1171260350) that the template parameter must not be renamed between inherited classes. This advice solved the problem in ttps://github.com/nunomaduro/larastan/issues/1289, but here we have it again.

Maybe, the problem occurs, because the template parameter is called `TRelatedModel` in the `Relation`-classes, but `TModelClass` in `Builder`. However, I do not know what black magic Larastan does internally that it chokes when template parameters are renamed. Maybe, the cause of the problem is something different here. I would be really grateful, if @canvural could have a look into it. This is out of my wheelhouse.

nagmat84:
I believe I fixed the problem even though I have not fully understood what I did. In order to solve the "one-must-not-rename template-parameter-names" I stepped through the source code, watched out for the literal string `'TModelClass`' and added `'TRelatedClass'` and `'TDeclaringClass'`. See here https://github.com/nunomaduro/larastan/pull/1285/commits/b8331c18394da74ed621ec19d0e584cbc7cd080a. Previously this has already been the case for `'TRelatedClass'` at some places, but not consistently.

As a second change I had to remove a condition in `EloquentBuilderForwardsCallsExtension` which prevented a method call to be forwarded to the underlying query object if the template parameter was bound to the unspecific `Model` class instead of a more specific user-defined model class. See this commit for the change https://github.com/nunomaduro/larastan/pull/1285/commits/f660f26fe008318e58c71b61b220faf99aa6f6ab.

Obviously, this is required if one implements custom relations which shall work for all type of models extended the `Model` class and not only specific models. However, I don't know why the condition which I have removed has been there in the first place. All tests are still running, hence the condition seems not to have had any impact previously and I assume it is OK to remove it. Unfortunately, the code is missing comments explaining _why_ the original author added this or that condition. Someone with more knowledge about Larastan needs to look into that change thoroughly.

I hope the PR is now fine to be merged.

szepeviktor:
@canvural I thing we should release it and see what will custom relation developers have to say.

nagmat84:
I wonder how many there are. Obviously, I am one and I directly crashed into a lot of problems which led to this PR. So either the rest does not care and uses a lot of `phpstan-ignore` annotations or there are no other developers which use custom relations.

szepeviktor:
> there are no other developers which use custom relations.

I support one of those developers: having a big `phpstan-baseline.neon` file.

nagmat84:
What is the matter of state? When will this PR be merged? What can I do to push things forward?

szepeviktor:
I'm not allowed to answer as I'm also sitting in my lab 🧪 but Larastan must be [lazy developer](https://packagist.org/packages/nunomaduro/larastan/stats) compatible.

nagmat84:
Who is allowed to answer and has the authority to push things forward? @szepeviktor Thanks for your support.

canvural:
I'm currently really busy with work, and cannot review it now. I'll do it as soon as I can find some time.

nagmat84:
Is it possible to give a rough when this PR will be reviewed or get merged?

szepeviktor:
> a rough estimate

In two weeks.

szepeviktor:
@nagmat84 You can se how we work on the rhythm of commits in master.

nagmat84:
There have been two two-weeks cycles since the last comment. What's the progress on this PR?

canvural:
> There have been two two-weeks cycles since the last comment. What's the progress on this PR?

Hi,

We do not have any cycles of development here. 2 weeks was just an estimate of @szepeviktor I think there.

Anyway, I apologize for the lack of communication here. Past few months were crazy busy for me. Work wise and personal issues.

Only last week I found some time to get through some issues and pull requests. Made a new release with them. And still there are some issues/prs to review left. This is kinda a large PR, so first I spent time on smaller ones.

I'll try to get to this as soon as possible though. Alternatively you can always motivate me through the sponsors: https://github.com/sponsors/canvural 😊 

Thank you for understanding.

nagmat84:
I'll like to pose my frequent question. ;-)

mad-briller:
> There should be no breaking change for ordinary users 

this is not true, any users who have `@return HasOne<Model>` in phpdocs above their relations will have to update them, which could definitely be seen as a breaking change.
it's also worth noting that without such a phpdoc, phpstan will error with:

```
Method Model::relation() return type with generic class Illuminate\Database\Eloquent\Relations\HasOne does not specify its types: TRelatedModel  
``` 

which forces the addition of such a phpdoc

nunomaduro:

PR #1287, opened 2022-06-22
Updated stubs

ricklambrechts:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Added psalm types because psalm does not understand 'model-property'.

**Breaking changes**

### DIFF
--- a/stubs/BelongsToMany.stub
+++ b/stubs/BelongsToMany.stub
@@ -12,6 +12,7 @@ class BelongsToMany extends Relation
      * Find a related model by its primary key or return new instance of the related model.
      *
      * @param  mixed  $id
+     * @psalm-param  array<int, string>|string  $columns
      * @param  array<int, (model-property<TRelatedModel>|'*')>|model-property<TRelatedModel>|'*'  $columns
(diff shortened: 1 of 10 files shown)


szepeviktor:
Hello-hello @ricklambrechts! 👋🏻 

We are a PHPStan extension :)

szepeviktor:
Does https://github.com/psalm/psalm-plugin-laravel work for you?

szepeviktor:
> There are 3 options to fix that:

Could you add a 4th option:

- copy Larastan stubs to https://github.com/psalm/psalm-plugin-laravel/tree/master/stubs

canvural:
Hi,

You are correct `model-property` type is special to Larastan. So we might need to prefix it.

But I am not sure if overwriting `@param` tags is really supported by PHPStan or not. A simple example show that `@phpstan-param` takes precedence: https://phpstan.org/r/c5bdc938-9a8f-42d2-af2d-aec16237b918So I hope it can work on all cases.

So what I would like to see is to use `@phpstan-param` whereever `model-property` is used. And have `@param` for normal PHPDocs.

Can you update the PR according to that? Thanks.

canvural:
Closing this as it is out of scope currently. Psalm plugin can write it's own stubs for generics.

PR #1296, opened 2022-07-01
Translation Helper with underscore function support

### LINKED ISSUE #1227: `__()` is recognized different from `trans()`
- Larastan Version: 1.0.3
- `--level` used: 7

### Description

The `__()` function isn't allowed to be used where only string is accepted as it can return `array|string|null` which is true, same for `trans()` but because of the `TransExtension` the `trans()` helper is allowed to be used for string concatination or similar.

```php
__('my.translation').' '.$variable;
// vs
trans('my.translation').' '.$variable;
```

niekbr:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves #1227 

**Changes**

The `__()` translation helper behaves very similar to `trans()` except when calling them without arguments. The `TransExtension` has been extended to also use the mixed type when an argument is supplied for the `__()` helper, like `trans()`. 

**Breaking changes**
None.


szepeviktor (approved):

szepeviktor:
Very nice work, Niek!

canvural:
Hi,

Thank you for this. 

But looks like this could be simplified further with conditional return types. So can you do these changes:
* You can delete the `TransExtension` class, it's test, and the mention from `extension.neon`
* Then you can add the stubs for `trans` and `__` to https://github.com/nunomaduro/larastan/blob/ab0cca9384605d5af0ac4170230849113b8344c5/stubs/Helpers.stub For example for `trans` return type can be `@return ($key is null ? \Illuminate\Contracts\Translation\Translator : mixed)` Similar for `__`
* And for the tests, you can add a new file to `tests/Type/data` directory. You can check the other tests for how to test it. And don't forget to add the file to the list in `GeneralTypeTest`

If this is too confusing or too much to do, no problem, just let me know and I can do it 😊

niekbr:
I'm having trouble implementing this, getting these errors:

Implementation in `stubs/Helpers.stub`
```php
/**
 * Translate the given message.
 *
 * @param  string|null  $key
 * @param  array<string, string>  $replace
 * @param  string|null  $locale
 * @return ($key is null ? \Illuminate\Contracts\Translation\Translator : mixed)
 */
function trans($key = null, $replace = [], $locale = null)
{
    if (is_null($key)) {
        return app('translator');
    }

    return app('translator')->get($key, $replace, $locale);
}
```

Tests are failing:
```json
{
    "totals": {
        "errors": 0,
        "file_errors": 1
    },
    "files": {
        "\/home\/niek\/coding\/larastan\/src\/..\/stubs\/Helpers.stub": {
            "errors": 1,
            "messages": [
                {
                    "message": "Function trans() has invalid return type Illuminate\\Contracts\\Translation\\Translator.",
                    "line": 143,
                    "ignorable": false
                }
            ]
        }
    },
    "errors": []
}
```

stubs/Helpers.stub
```diff
@@ -131,3 +131,32 @@ function tap($value, $callback = null)
...
+ * @param  string|null  $locale
+ * @return ($key is null ? \Illuminate\Contracts\Translation\Translator : mixed)
+ */
+function trans($key = null, $replace = [], $locale = null)
+{
+    if (is_null($key)) {
+        return app('translator');
+    }
+
+    return app('translator')->get($key, $replace, $locale);
```
szepeviktor, line 149: This is a stub. You don't need function bodies.

szepeviktor:
I see! Stubs should contain all referenced classes.
Please add `namespace Illuminate\Contracts\Translation; interface Translator {}` in `Contracts/Translation.php`

interface Translator

stubs/Helpers.stub
niekbr: It was still a WIP to show my attempt :)
Fixed now

szepeviktor (approved):

canvural:
Perfect. Thank you!

PR #1305, opened 2022-07-02
Add stub files to CS config

szepeviktor:


szepeviktor:
👀 

kép

szepeviktor:
@GrahamCampbell Could you help us?

GrahamCampbell:
Those stub files are not valid PHP syntax, thus are crashing php-cs-fixer.

GrahamCampbell:
Specifically, it's the class attributes separation, which makes sense given the functions are missing their bodies. You could try turning that off.

szepeviktor:
Thank you.

PR #1319, opened 2022-07-24
fix: use Composer autoloader to determine vendor-dir

szepeviktor:
Copied from https://github.com/szepeviktor/package-path/blob/18f93ac05953340216f76e5fef0b9e1c664e7f70/src/PackagePath.php#L35-L58


szepeviktor:
@canvural It determines vendor-dir by locating `Composer\Autoload\ClassLoader`.
Please merge it.

canvural:
Thank you.

PR #1329, opened 2022-08-08
Configure Composer properly in CI

szepeviktor:
Using `composer config`


szepeviktor:
Stolen from WooCommerce.

PR #1333, opened 2022-08-15
feat(refactor): ModelProperty casting

### LINKED ISSUE #890: Missing support for inbound casting
- Larastan Version: latest
- `--level` used: 6

### Description

Using inbound casting: https://laravel.com/docs/8.x/eloquent-mutators#inbound-casting.

```php
    protected $casts = [
        'bio' => Markdown::class, // sets the attribute value to a `string`
    ];
```

Larastan thinks that `$model->bio` is a `Markdown` instance instead of a string.

erikgaal:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

Fixes #890

**Changes**

Adds better support for casting using `CastsAttributes`, `CastsInboundAttributes` and `Castable`. Also refactored the `ModelPropertyExtension` and extracted a `ModelCastHelper` that can convert Eloquent casts to PHPStan types.

**Breaking changes**

### DIFF
--- a/src/Properties/ModelCastHelper.php
+++ b/src/Properties/ModelCastHelper.php
@@ -41,7 +41,7 @@ class ModelCastHelper
             'object' => new ObjectType('stdClass'),
             'array', 'json' => new ArrayType(new MixedType(), new MixedType()),
             'collection' => new ObjectType('Illuminate\Support\Collection'),
-            'date', 'datetime' => new ObjectType('Carbon\Carbon'),
+            'date', 'datetime' => $this->getDateType(),
             'immutable_date', 'immutable_datetime' => new ObjectType('Carbon\CarbonImmutable'),
             'timestamp' => new IntegerType(),
             default => null,
--- a/src/Properties/ModelCastHelper.php
+++ b/src/Properties/ModelCastHelper.php
@@ -1,0 +1,145 @@
+<?php
+
+namespace NunoMaduro\Larastan\Properties;
+
+use Illuminate\Contracts\Database\Eloquent\Castable;
+use Illuminate\Contracts\Database\Eloquent\CastsAttributes;
+use Illuminate\Contracts\Database\Eloquent\CastsInboundAttributes;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Str;
+use PHPStan\Reflection\ParameterReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Type\Accessory\AccessoryNumericStringType;
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\BooleanType;
+use PHPStan\Type\Constant\ConstantIntegerType;
+use PHPStan\Type\FloatType;
+use PHPStan\Type\IntegerType;
+use PHPStan\Type\MixedType;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+
+class ModelCastHelper
+{
+    public function __construct(protected ReflectionProvider $reflectionProvider)
+    {
+    }
+
+    public function getReadableType(string $cast, Type $originalType): Type
+    {
+        $cast = Str::before($cast, ':');
+
+        $attributeType = match ($cast) {
+            'int', 'integer' => new IntegerType(),
+            'real', 'float', 'double' => new FloatType(),
+            'decimal' => new AccessoryNumericStringType(),
+            'string' => new StringType(),
+            'bool', 'boolean' => new BooleanType(),
+            'object' => new ObjectType('stdClass'),
+            'array', 'json' => new ArrayType(new MixedType(), new MixedType()),
+            'collection' => new ObjectType('Illuminate\Support\Collection'),
+            'date', 'datetime' => new ObjectType('Carbon\Carbon'),
+            'immutable_date', 'immutable_datetime' => new ObjectType('Carbon\CarbonImmutable'),
+            'timestamp' => new IntegerType(),
+            default => null,
+        };
+
+        if ($attributeType) {
+            return $attributeType;
+        }
+
+        if (! $this->reflectionProvider->hasClass($cast)) {
+            return new MixedType();
+        }
+
+        $classReflection = $this->reflectionProvider->getClass($cast);
+
+        if ($classReflection->isSubclassOf(Castable::class)) {
+            $methodReflection = $classReflection->getNativeMethod('castUsing');
+            $castUsingReturn = ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+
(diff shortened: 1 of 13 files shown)


szepeviktor:
@erikgaal Thank you for your contribution!

What does `DTO` stand for?

erikgaal:
> @erikgaal Thank you for your contribution!
> 
> 
> 
> What does `DTO` stand for?
> 
> 

Hi there! It stands for `DataTramsferObject`, but we can rename it to `ValueObject` so it matches the [Laravel documentation](https://laravel.com/docs/9.x/eloquent-mutators#value-object-casting).

szepeviktor:
Thank you!

(🚊 typo 🚊)

szepeviktor (approved):

canvural:
Thank you for the work here. Just FYI I'm currently on holiday and not available to review until 20th September.

erikgaal:
@canvural Hope you had a great holiday! Would you be available to review this PR?

src/Properties/ModelCastHelper.php
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    public function getReadableType(string $cast, Type $originalType): Type
+    {
+        $cast = Str::before($cast, ':');
+
+        $attributeType = match ($cast) {
+            'int', 'integer' => new IntegerType(),
+            'real', 'float', 'double' => new FloatType(),
+            'decimal' => new AccessoryNumericStringType(),
```
canvural, line 38: Accessory types cannot be used standalone. You need to do `TypeCombinator::intersect(new StringType(), new AccessoryNumericStringType())` here

src/Properties/ModelCastHelper.php
```diff
@@ -0,0 +1,145 @@
...
+            'string' => new StringType(),
+            'bool', 'boolean' => new BooleanType(),
+            'object' => new ObjectType('stdClass'),
+            'array', 'json' => new ArrayType(new MixedType(), new MixedType()),
+            'collection' => new ObjectType('Illuminate\Support\Collection'),
+            'date', 'datetime' => $this->getDateType(),
+            'immutable_date', 'immutable_datetime' => new ObjectType('Carbon\CarbonImmutable'),
```
canvural, line 45: Shouldn't we use `$this->getDateType()` here also? Or another new method.

Isn't it possible that it will return `DateTimeImmutable` instead of `CarbonImmutable`?

src/Properties/ModelPropertyExtension.php
```diff
@@ -44,270 +44,147 @@ public function hasProperty(ClassReflection $classReflection, string $propertyNa
...
+        }
 
-        $this->tables[$tableName]->columns[$propertyName] = $column;
+        return false;
+    }
 
-        return true;
+    private function migrationsLoaded(): bool
+    {
+        return ! empty($this->tables);
```
canvural, line 110: Please use `count` instead of `empty`

src/Properties/ModelPropertyExtension.php
```diff
@@ -44,270 +44,147 @@ public function hasProperty(ClassReflection $classReflection, string $propertyNa
...
 
-        $column->readableType = $readableType;
-        $column->writeableType = $writableType;
+            $returnType = ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+
+            if (! (new ObjectType(Attribute::class))->isSuperTypeOf($returnType)->yes()) {
+                return false;
+            }
+
+            return true;
```
canvural, line 102: I think we can simplify it to `return (new ObjectType(Attribute::class))->isSuperTypeOf($returnType)->yes()`

src/Properties/ModelCastHelper.php
```diff
@@ -0,0 +1,145 @@
...
+        $cast = Str::before($cast, ':');
+
+        $attributeType = match ($cast) {
+            'int', 'integer' => new IntegerType(),
+            'real', 'float', 'double' => new FloatType(),
+            'decimal' => new AccessoryNumericStringType(),
+            'string' => new StringType(),
+            'bool', 'boolean' => new BooleanType(),
+            'object' => new ObjectType('stdClass'),
+            'array', 'json' => new ArrayType(new MixedType(), new MixedType()),
```
canvural, line 42: We can do better than `mixed` for the array key here. Try `new BenevolentUnionType([new IntegerType(), new StringType()])`

src/Properties/ModelPropertyExtension.php
```diff
@@ -44,270 +44,147 @@ public function hasProperty(ClassReflection $classReflection, string $propertyNa
...
+            $propertyName === $modelInstance->getKeyName()
+            && (! array_key_exists($tableName, $this->tables) || ! array_key_exists($propertyName, $this->tables[$tableName]->columns))
         ) {
             return new ModelProperty(
-                $classReflection,
-                $this->stringResolver->resolve($modelInstance->getKeyType()),
-                $this->stringResolver->resolve($modelInstance->getKeyType())
+                declaringClass: $classReflection,
+                readableType: $this->stringResolver->resolve($modelInstance->getKeyType()),
+                writableType: $this->stringResolver->resolve($modelInstance->getKeyType()),
```
canvural, line 142: Please don't use named arguments.

canvural (changes requested):
Hi, I added some comments.

Please also add tests for the cast types that do not have any tests. Like `immutable_date`

You can do it by adding adding the casts to any model in the `Application` folder, and then add a new file to https://github.com/nunomaduro/larastan/blob/4bb0df1aee0d0b8485855f068693688c6f7ae7db/tests/Type/GeneralTypeTest.php and test with `assertType`

src/Properties/ModelCastHelper.php
erikgaal: I don't think this will ever return `DateTimeImmutable`, because [this line](https://github.com/laravel/framework/blob/9138ffdf886f130d7cf372ab165b650b201e28e2/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L781) will throw an error if `asDateTime` or `asDate` returns a `DateTime`, because it doesn't have the `toImmutable` method.

erikgaal:
Hi @canvural!

Sorry for taking so long to continue, but I've implemented most of your comments. Please also advise what to do with `immutable_datetime`.

Kind regards,
Erik

src/Properties/ModelCastHelper.php
canvural: I tested it on a project. Yeah it throws error exactly like you said. So this is fine 👍🏽

canvural:
@erikgaal That's fine. More precise types leading to errors is acceptable. It's also in PHPStan's BC promise page: https://phpstan.org/user-guide/backward-compatibility-promise#type-inference-capabilities

Just FYI: this PR discovered an issue. https://github.com/nunomaduro/larastan/blob/master/src/Properties/SchemaAggregator.php#L368 this now leads to a object with `enum` as class name. No idea how it was working before 😄 Could be also a PHPStan change. I'll try to fix it.

canvural:
Reason is that https://github.com/nunomaduro/larastan/blob/2.2.9/src/Properties/ModelPropertyExtension.php#L190-L222 this code was removed.

PR #1335, opened 2022-08-20
fix: return correct types from model factories

szepeviktor:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

**Breaking changes**


canvural:
Thank you. But I'll open it when I think it's ready.

PR #1339, opened 2022-08-23
fix: return correct types from model factories

canvural:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

Better return type inference for model factories. Now Larastan understands that `User::factory()->create()` returns a single model, whereas `User:factory(2)->create()` or `User::factory()->count(2)->create()` returns a collection.

**Breaking changes**

### DIFF
--- a/tests/Type/GeneralTypeTest.php
+++ b/tests/Type/GeneralTypeTest.php
@@ -13,14 +13,15 @@ class GeneralTypeTest extends TypeInferenceTestCase
      */
     public function dataFileAsserts(): iterable
     {
-        yield from $this->gatherAssertTypes(__DIR__.'/data/request-object.php');
-        yield from $this->gatherAssertTypes(__DIR__.'/data/eloquent-builder.php');
-        yield from $this->gatherAssertTypes(__DIR__.'/data/paginator-extension.php');
-        yield from $this->gatherAssertTypes(__DIR__.'/data/model-properties.php');
-        yield from $this->gatherAssertTypes(__DIR__.'/data/model-properties-relations.php');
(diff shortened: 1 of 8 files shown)


tests/Type/GeneralTypeTest.php
```diff
@@ -13,14 +13,15 @@ class GeneralTypeTest extends TypeInferenceTestCase
...
-        yield from $this->gatherAssertTypes(__DIR__.'/data/model-properties.php');
-        yield from $this->gatherAssertTypes(__DIR__.'/data/model-properties-relations.php');
-        yield from $this->gatherAssertTypes(__DIR__.'/data/route.php');
-        yield from $this->gatherAssertTypes(__DIR__.'/data/conditionable.php');
-        yield from $this->gatherAssertTypes(__DIR__.'/data/translate.php');
+//        yield from $this->gatherAssertTypes(__DIR__.'/data/request-object.php');
+//        yield from $this->gatherAssertTypes(__DIR__.'/data/eloquent-builder.php');
+//        yield from $this->gatherAssertTypes(__DIR__.'/data/paginator-extension.php');
+//        yield from $this->gatherAssertTypes(__DIR__.'/data/model-properties.php');
+//        yield from $this->gatherAssertTypes(__DIR__.'/data/model-properties-relations.php');
```
crissi, line 20: you need to add these back:)

szepeviktor: was fixed in https://github.com/nunomaduro/larastan/commit/ef427e6ff2badecc83b0d872acfd958e2d6f3379

PR #1343, opened 2022-08-25
Remove all unnecessary method stubs for the Query Builder

lupinitylabs:
- [ ] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

This PR updates the QueryBuilder Stub for the whereMonth, orWhereMonth, whereDay and orWhereDay methods to match the changes introduced in https://github.com/laravel/framework/pull/43668 and released in Laravel 9.25.0.

Previously, phpstan reported an error like `Parameter #3 $value of method Illuminate\Database\Eloquent\Builder<App\Models\MyModel>::whereMonth() expects DateTimeInterface|string|null, int given.` if integers were passed to any of these methods.

**Breaking changes**

Technically, these methods already took integers which were internally cast to strings, so this should not introduce any breaking changes.


szepeviktor:
Thank you!

Could you add a single line of test that passes an integer?

canvural:
If the stub is the same, we can remove it here. I copy pasted all methods for some reason, but now I think we don't need them in the stubs.

lupinitylabs:
> If the stub is the same, we can remove it here. I copy pasted all methods for some reason, but now I think we don't need them in the stubs.

I'll scan through all the methods in the stub and delete those that are identical.

lupinitylabs:
Do you want to preserve all the cases where you specified the array types, e.g. `array<mixed>`?
I only see `array<mixed>` used so far, so I don't think that is a huge benefit...

lupinitylabs:
There were some other outdated methods, like orWhereNull, or several cases where an Eloquent Builder instance is allowed additionally to the Query Builder instance, or `whereBetween`, that now takes an iterable and not just an array, `latest`, `oldest` and `reorder` now taking closures and Builder instances and many more.

One special case is `pluck`, `increment` and `decrement`: the stub defines `$column` to be a `string|\Illuminate\Database\Query\Expression`, whereas the framework class defines it to only be a string. I suppose there's a reason for doing it, so I am leaving it in.

In any case, this should improve maintainability. 
Let me know what you think.

canvural:
Thank you!

PR #1344, opened 2022-08-26
Added better typing for the Optional class and helper.

mad-briller:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Added a conditional return type to the optional() method along with template typing.
Made the Optional class a `universalObjectCrateClass` as it will return a value for all property names.


szepeviktor (approved):

canvural:
Can you rebase it on top of master branch instead of a merge commit? Thanks!

canvural:
Thank you!

canvural:
Hi @mad-briller 

This change caused some errors in my testing.

For example now `optional($this->some_carbon_date_time_field)->toIso8601String()` produces the error `Call to an undefined method Illuminate\Support\Optional::toIso8601String()`

Could you take a look at that? Thanks.

mad-briller:
i guess in the past this wasn't raised because of the explicit `mixed` present in laravel's phpdoc for `optional()`, now that phpstan knows it's `Optional` it's trying to make more guarantees about it 

it may take a bit of work for me to do that and i'm not sure when i can get to it, my idea would be to make `Optional` generic and then use some property / method reflections to see if the proxied object has that property / method

canvural:
Sure no problem. I also like the generics idea. I might look into that tonight!

mad-briller:
oh sweet, i'll keep my eye out, thanks

if we do make Optional generic, the `universalObjectCratesClasses` definition for `Optional` can be removed, because the property reflection would want to take precedence and can return `new MixedType` in the case it is not present

canvural:
@mad-briller On a second though I decided not to go generics way. There is no reason to complicate things. Since `?->` exists in PHP using `optional` is not needed anymore anyway. So I did this https://github.com/nunomaduro/larastan/commit/8cdd75dec47b7bc6d0bc0a524116da1df31ca2d3 We will still get better types when callback is used. But for calls just with the value, we assume `mixed`

PR #1347, opened 2022-08-28
docs: help copy-pasters with configuration

szepeviktor:
I see many developers using this very configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,18 +52,18 @@ includes:
 parameters:
 
     paths:
-        - app
+        - app/
 
-    # The level 9 is the highest level
+    # Level 9 is the highest level
     level: 5
 
-    ignoreErrors:
-        - '#PHPDoc tag @var#'
-
-    excludePaths:
-        - ./*/*/FileToBeExcluded.php
-
-    checkMissingIterableValueType: false
+#    ignoreErrors:
+#        - '#PHPDoc tag @var#'
+#
+#    excludePaths:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,18 +52,18 @@ includes:
...
-        - ./*/*/FileToBeExcluded.php
-
-    checkMissingIterableValueType: false
+#    ignoreErrors:
+#        - '#PHPDoc tag @var#'
+#
+#    excludePaths:
+#        - ./*/*/FileToBeExcluded.php
+#
+#    checkMissingIterableValueType: false
```
nunomaduro, line 66: Feel that, for most Laravel applications, we should start with `checkMissingIterableValueType: false`. @canvural What's your thoughts on this one?

nunomaduro (changes requested):

README.md
szepeviktor: > we should start with `checkMissingIterableValueType: fals`

That starts only on level 6, the above example has level 5.

nunomaduro (approved):
Going to let @canvural validate and merge this one.

README.md
canvural: I think I agree with both of you 😄 As @szepeviktor said that check only applies for level 6 and above. So leaving it commented here would be a good idea, if anyone wants to go above level 6.

canvural:
Thank you!

PR #1354, opened 2022-09-01
fix: abstract manager would cause larastan to crash.

mad-briller:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Encountered a situation where an `abstract` Manager class caused Larastan to crash, because it would try to resolve it from the container and then try to call a method on the `null` result, as the container cannot build an abstract class.

wasn't sure how to add tests for this as i couldn't find anything already covering the manager pipe, let me know if i missed it.


canvural:
Hi,

Yeah it seems no test covers that. You can try to add tests into the https://github.com/nunomaduro/larastan/blob/master/tests/Type/GeneralTypeTest.php

mad-briller:
@canvural i've added a test case to prove there is no crash when an `abstract` Manager is encountered

mad-briller:
the failing action seems to be caused by an unrelated change

it would seem their phpdoc doesn't state an `array<int, string>` can be given to Str::startsWith, i wonder if their `&iterable<string>` was meant to be `|iterable<string>`

canvural:
Thank you!

PR #1358, opened 2022-09-02
feat: run integration tests in Github Actions

canvural:
This PR adds a new GA that will clone some popular open source Laravel projects that uses Larastan, and run the Larastan on it.

This way we'll see if our changes will break any stuff in real world applications.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,0 +1,55 @@
+name: "Integration tests"
+
+on:
+  workflow_run:
+    workflows: ["tests"]
+    types: [completed]
+
+jobs:
+  integration-tests:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    name: "Integration Tests"
+
+    runs-on: "ubuntu-latest"
+    timeout-minutes: 30
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+#          - php-version: 8.1
+#            script: |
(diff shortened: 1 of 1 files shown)


szepeviktor:
Maybe `composer config minimum-stability dev` will do.

szepeviktor:
I think `if:` must be written in all steps, not on job level.

canvural:
> I think `if:` must be written in all steps, not on job level.

Nah, I think it was because there were also `pull_request` and `push`. `workflow_run` should be the only one. We will see.

szepeviktor:
We may need clean up that directory as the first command `rm -rf ../e2e/integration/repo`

canvural:
> We may need to clean up that directory as the first command `rm -rf ../e2e/integration/repo`

I don't think so. These are defined in the matrix so they are ran separately.

canvural:
Hmm, the `integration-tests` didn't even run last time.

I found this in the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run)

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

Maybe we need to merge this now 🤔

szepeviktor:
I would open a temporary repo with `echo 1;` and `echo 2;` as commands.

canvural:
So this is is waiting for https://github.com/monicahq/monica/pull/6305 After that I'll replace the commit hash and we can try again.

PR #1361, opened 2022-09-02
chore: run integration tests parallel to normal tests

canvural:
//

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,13 +1,20 @@
 name: "Integration tests"
 
 on:
-  workflow_run:
-    workflows: ["tests"]
-    types: [completed]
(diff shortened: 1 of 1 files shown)


szepeviktor:
May I create a tiny repo for this purpose?

canvural:
> May I create a tiny repo for this purpose?

For what exactly?

Anyway, currently integration tests workflow runs only tests are green, but is not shown in the PR checks. I read some stuff online and there is no workaround that. So I think I'll convert it to just run parallel.

szepeviktor:
> there is no workaround

Very worring.

---

kép

:)

szepeviktor:
I do not have the strength to argue.
Here we have a naming mismatch. These are E2E tests, not integrations tests.

PR #1369, opened 2022-09-06
fix: `retry` helper supports array of sleep times for $times

tpetry:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

The `retry` helper has a lesser-known additional supported type for the `$times` parameter. You can provide an array of sleep times that will be used.

**Breaking changes**

None

### DIFF
--- a/stubs/Helpers.stub
+++ b/stubs/Helpers.stub
@@ -100,7 +100,7 @@ function rescue(callable $callback, $rescue = null, $report = true)
 
 /**
  * @template TValue
- * @param int $times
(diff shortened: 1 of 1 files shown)


szepeviktor:
Thank you!

https://github.com/laravel/framework/blob/b34d67a1b1c60e0616d3a5490d00195bce08af0d/src/Illuminate/Support/helpers.php#L215-L226

canvural:
Thank you!

PR #1372, opened 2022-09-14
feat: add support for generic paginators

erikgaal:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

This PR adds support for generic paginators.

Before:

```php
$users = User::query()->paginate(); // LengthAwarePaginator
$users->items(); // array
```

After:

```php
$users = User::query()->paginate(); // LengthAwarePaginator<User>
$users->items(); // array<User>
```

**Breaking changes**

### DIFF
--- a/stubs/Contracts/Pagination.stub
+++ b/stubs/Contracts/Pagination.stub
@@ -4,9 +4,6 @@ namespace Illuminate\Contracts\Pagination;
 
 /**
  * @template TItem
- *
- * @mixin \Illuminate\Support\Collection<array-key, TItem>
- * @mixin \Illuminate\Pagination\Paginator<TItem>
  */
 interface Paginator
 {
@@ -19,8 +16,6 @@ interface Paginator
 /**
  * @template TItem
  *
- * @mixin \Illuminate\Support\Collection<array-key, TItem>
- * @mixin \Illuminate\Pagination\LengthAwarePaginator<TItem>
  * @extends Paginator<TItem>
  */
 interface LengthAwarePaginator extends Paginator
@@ -28,9 +23,6 @@ interface LengthAwarePaginator extends Paginator
 
 /**
  * @template TItem
- *
- * @mixin \Illuminate\Support\Collection<array-key, TItem>
- * @mixin \Illuminate\Pagination\CursorPaginator<TItem>
  */
 interface CursorPaginator
 {
--- a/stubs/Contracts/Pagination.stub
+++ b/stubs/Contracts/Pagination.stub
@@ -3,22 +3,39 @@
 namespace Illuminate\Contracts\Pagination;
 
 /**
- * @mixin \Illuminate\Support\Collection
- * @mixin \Illuminate\Pagination\Paginator
+ * @template TItem
+ *
+ * @mixin \Illuminate\Support\Collection<array-key, TItem>
+ * @mixin \Illuminate\Pagination\Paginator<TItem>
  */
 interface Paginator
-{}
+{
+    /**
+     * @return array<TItem>
+     */
+    public function items(): array;
+}
 
 /**
- * @mixin \Illuminate\Support\Collection
- * @mixin \Illuminate\Pagination\LengthAwarePaginator
+ * @template TItem
+ *
+ * @mixin \Illuminate\Support\Collection<array-key, TItem>
+ * @mixin \Illuminate\Pagination\LengthAwarePaginator<TItem>
+ * @extends Paginator<TItem>
  */
 interface LengthAwarePaginator extends Paginator
 {}
 
 /**
- * @mixin \Illuminate\Support\Collection
- * @mixin \Illuminate\Pagination\CursorPaginator
+ * @template TItem
+ *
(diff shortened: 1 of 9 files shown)


tests/Type/data/paginator-extension.php
```diff
@@ -7,6 +7,13 @@
...
+assertType('Illuminate\Pagination\LengthAwarePaginator<App\User>', User::paginate());
+assertType('array<App\User>', User::paginate()->items());
+
+assertType('Illuminate\Pagination\Paginator<App\User>', User::simplePaginate());
+assertType('array<App\User>', User::simplePaginate()->items());
+
+assertType('Illuminate\Pagination\CursorPaginator<App\User>', User::cursorPaginate());
+assertType('array<App\User>', User::cursorPaginate()->items());
+
+assertType('ArrayIterator<mixed, App\User>', User::query()->paginate()->getIterator());
```
szepeviktor, line 19: Array keys could only be integers or strings, not `mixed`

erikgaal: Thanks!

szepeviktor (approved):

tests/Type/data/paginator-extension.php
```diff
@@ -7,6 +7,19 @@
 use App\User;
 use function PHPStan\Testing\assertType;
 
-assertType('array', User::paginate()->all());
-assertType('array', User::simplePaginate()->all());
-assertType('array', User::cursorPaginate()->all());
```
canvural, line 12: Please don't remove these tests.

canvural (changes requested):

tests/Type/data/paginator-extension.php
erikgaal: I would rather not remove tests, but, `$paginator->all()` is not the recommended way of retrieving the items from a paginator instance. The tests will also fail, because PHPStan is unable to figure our the right return type for `all()`, because it is inherited from the `Collection` @mixin, rather than a dedicated method on the Paginator instance.

https://laravel.com/docs/9.x/pagination#paginator-instance-methods

canvural: It shouldn't matter if it is the recommended way or not. Like you said `AbstractPaginator` have `@mixin Collection` so it should be possible to do these collection calls on paginators.

> PHPStan is unable to figure our the right return type for all(), because it is inherited from the Collection @mixin
Why is that? PHPStan understands `@mixin`.

stubs/Contracts/Pagination.stub
```diff
@@ -3,22 +3,39 @@
 namespace Illuminate\Contracts\Pagination;
 
 /**
- * @mixin \Illuminate\Support\Collection
- * @mixin \Illuminate\Pagination\Paginator
+ * @template TItem
+ *
+ * @mixin \Illuminate\Support\Collection<array-key, TItem>
+ * @mixin \Illuminate\Pagination\Paginator<TItem>
```
canvural, line 9: Maybe put these mixin tags into the concrete classes. `AbstractCursorPaginator` and `AbstractPaginator`

Same for the other interfaces below.

canvural (changes requested):

tests/Type/data/paginator-extension.php
erikgaal: I reported it to PHPStan in https://github.com/phpstan/phpstan/issues/8008, and it looks like to be fixed, though unreleased.

erikgaal: Pardon me, it's already released so we can add the tests back. :)

erikgaal: Done! ✅

stubs/Contracts/Pagination.stub
erikgaal: ✅

stubs/Pagination.stub
```diff
@@ -28,8 +28,8 @@ abstract class AbstractPaginator implements \Illuminate\Contracts\Support\Htmlab
 /**
  * @template TValue
  *
- * @implements \ArrayAccess<mixed, TValue>
- * @implements \IteratorAggregate<mixed, TValue>
+ * @implements \ArrayAccess<array-key, TValue>
```
canvural, line 31: Since this class implements `ArrayAccess` we should add the `offset*` methods here with the correct return/parameter types. This will let PHPStan know `User::paginate(10)[3]` returns `User` An example can be seen [here.](https://github.com/phpstan/phpstan-src/blob/bfcef682d335bf10d5c317aad53b0771276bfbd6/stubs/ArrayObject.stub)

Also for the other classes below + tests for this 👍🏽

canvural (changes requested):
One more thing. Other than that it looks good!

stubs/Pagination.stub
erikgaal: Done! ✅

canvural:
That's great. Can you also resolve the conflict? Then if the CI is green, we can merge this!

canvural:
Errors from `monicahq/chandler` looks legit.

Thank you!

PR #1375, opened 2022-09-19
Added the ability to check for unnecessary collection calls on dynamic properties in NoUnnecessaryCollectionCallRule

zlayaAvocado:
- [X] Added or updated tests
- [X] Updated CHANGELOG.md

**Changes**

Added the ability to check for unnecessary collection calls on dynamic properties in `NoUnnecessaryCollectionCallRule`. E.g. in the following faulty example, instead of querying for all accounts for the user first, we build a query and execute it with the `count`

**Bad:**
`User::firstOrFail()->accounts->count();`

**Good:**
`User::firstOrFail()->accounts()->count();`

Also see example in [this article](https://codeburst.io/how-to-use-laravels-eloquent-efficiently-d46f5c392ca8) why it's helpful and why it can cause performance issues.

>Don’t use a collection to count the number of related entries.

**Breaking changes**


zlayaAvocado:
> Can you rebase it on top of master? Then I'll merge. Thanks!

done!

PR #1378, opened 2022-09-24
Add a rule disallowing the usage of global helper functions that depend on a booted application

PrinsFrank:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Except for a small subset of functions, most of the global helpers in laravel are dependent in some way or another on the state of a booted application. And while these functions are convenient ways to shorten the amount of code you have to write, they couple your code with the Laravel codebase. And furthermore, they decrease the testability of your codebase. This rule prevents the usage of those specific coupled helper methods, while still allowing usage of the other functions that don't rely on a booted application. for more information, see this [blog post](https://prinsfrank.nl/2022/09/20/How-to-write-decoupled-unit-tests-in-Laravel).

**Breaking changes**

Rule is disabled by default so this won't break anything


PrinsFrank:
@canvural @szepeviktor Any idea how I can fix the test run running out of memory?

szepeviktor:
a shot in the dark: what if you use FunctionTypeSpecifyingExtension?
https://github.com/nunomaduro/larastan/blob/e67a1b09fc8da255f991e52b0cd29c9ac3565a28/src/Types/AbortIfFunctionTypeSpecifyingExtension.php#L16

src/Rules/DecoupledCode/NoHelperDependingOnBootedApplicationUsageRule.php
```diff
@@ -0,0 +1,96 @@
...
+        'action'        => 'Inject Illuminate\Contracts\Routing\UrlGenerator and use method "action"',
+        'asset'         => 'Inject Illuminate\Contracts\Routing\UrlGenerator and use method "asset"',
+        'secure_asset'  => 'Inject Illuminate\Contracts\Routing\UrlGenerator and use method "asset" with the second parameter set to "true"',
+        'route'         => 'Inject Illuminate\Contracts\Routing\UrlGenerator and use method "route"',
+        'url'           => 'Inject Illuminate\Contracts\Routing\UrlGenerator and use method "url", "current" for the current url, "full" for the full url or "previous" for the previous url',
+        'secure_url'    => 'Inject Illuminate\Contracts\Routing\UrlGenerator and use method "url" with the third parameter set to "true"',
+        'redirect'      => 'Inject Illuminate\Routing\Redirector and use method "to"',
+        'to_route'      => 'Inject Illuminate\Routing\Redirector and use method "route"',
+        'back'          => 'Inject Illuminate\Routing\Redirector and use method "back"',
+        'config'        => 'Inject Illuminate\Config\Repository and use method "all" or "get"',
```
mad-briller, line 44: This rule is genius 


However, would this be better to recommend the config repository contract? As that would provide the most decoupling whilst working the same

To be honest nearly all of these cases have contracts that could be suggested rather than concrete classes, except for the managers perhaps 

I'm away from my pc atm so I can't research them all but that would provide the most decoupling and state independency

PrinsFrank: Good point! Will update them to use the contracts as much as possible!

PrinsFrank: I moved allmost all of the suggestion to contracts/interfaces, the only ones I didn't find a contract for where the Request, the Redirector and the SessionManager, as those are standalone classes.

PrinsFrank:
@szepeviktor @canvural I found the source of the memory increase. As the `now()` and `today()` function calls are resolved using the  DateExtension, and that date extension actually calls the `now()` function, the memory increases by 50mb as a container is resolved. As the idea is probably that the dateExtension is returns the correct type even if the function is overwritten, I can't really fix that. Is there something I can do here to solve this?

szepeviktor:
Maybe I am the only conservative around here but no feature is worth adding an extra half gigabyte memory.

@canvural Could we free up memory by removing something from the app container?

szepeviktor:
Actually current memory usage (unit tests/static analysis) is so huge I consider them an error.

PrinsFrank:
@szepeviktor I removed my comment within seconds of posting it because I was not entirely sure what was happening. The actual test case adds about 40mb of memory, but more things are going on the second i add any reference to `now()` or `today()`, but also the RelationExstinceRule test is failing the moment I add new tests for this rule. I am currently deep in the source code op PHPStan, the current suspect is the static $containers property in the PHPStanTestCase, but I am unsure atm.

PrinsFrank:
So to clarify, I currently have two issues: 
- Memory increases whenever a new call to `now()` or `today()` is added and analyzed by phpstan.
- A test execution order issue where the RelationExistenceRule test case is now broken because the newly added test is executed before it.

I was fed up with the memory issue as I wanted to go to sleep, so added the ridiculous new limit to see if my tests are at least passing, but that is not the case. I had a long day, so will continue debugging tomorrow. I suspect right now that part of the fix lies in the way the container is cached based on the sha1 of the additionalConfigFiles but not on the analyzed file set and that pollutes the phpstan container, but have to do some more research on that in the upcoming days.

mad-briller:
> - Memory increases by about 40mb whenever a new call to `now()` or `today()` is added and analyzed by phpstan.


Does this happen if you add a reference to Carbon::now() instead? 

Phpstan recently had a github issue whereby the Carbon class was adding a lot of overhead to analysis due to the phpdocs present in its traits etc. And I wonder if this is a symptom of that.

PrinsFrank:
@mad-briller Yes, that is also the case.

I spent a couple more evenings on the ordering issue in test, where the newly added tests break tests further on. As it looks right now; As PHPStan is not built te be run on part of the code, the container is cached only based on the additional configuration. As that is the same for all test cases in the folder, the cached container is used for all test cases, but because the individual tests are only run on single files, it doesn't run as expected. I removed the test cases for now. I will further debug the container issues in the phpstan base test case, but it will take a few weeks before I have some more time to do a deep dive.

PR #1379, opened 2022-09-25
Add @mixin Query\Builder to EloquentBuilder.stub

### LINKED ISSUE #1346: Underlying Query/Builder doesn't get resolved when using Eloquent/Builder
- Larastan Version: 2.1.12
- `--level` used: 6

### Description

Hi,  I currently have a file that looks like so:

```php

/**
 * @template TModel of \Illuminate\Database\Eloquent\Model
 */
class DailyQuery
{
    /**
     * @var Builder<TModel>
     */
    private Builder $query;

    ...
    
    public function daily(Carbon $start, Carbon $end) : DailyResults
    {
        $start = $start->startOfDay();
        $end = $end->endOfDay();
        $occurrencesPerDay = $this->query
            ->select($this->selection())
            ->whereBetween("{$this->query->getQuery()->from}.$this->column", [$start, $end])
            ->groupBy('date')
            ->orderBy('date')
            ->pluck('count', 'date');

        return new DailyResults($this->fillGaps($occurrencesPerDay, $start, $end));
    }
}
```

And I'm getting the error:
>  Call to an undefined method Illuminate\Database\Eloquent\Builder<TModel of Illuminate\Database\Eloquent\Model>::select().

I can remove the `->select(...)`, and it will instead complain that `whereBetween(...)` is undefined. I've gone through many issues here, and I'm uncertain what I'm doing wrong.

From my understanding, Laravel extends the `Query/Builder` using the mixin annotation, so the methods should be available.

You can see the entirety of the file here: https://github.com/SDU-SW-Engineering/ScalableTeaching/blob/admin-changes/domain/Analytics/DailyResults/DailyQuery.php

davidjr82:
- [ ] Added or updated tests (in progress, failing tests at the moment)
- [x] Documented user facing changes (not needed)
- [x] Updated CHANGELOG.md

**Changes**

Resolves #1346 as @xwillq is proposing in the issues comments.

As larastan is reading docblocks from stubs instead from source code, we need to add the mixins from Query\Builder in the Eloquent\Builder stub.

Reference: [Laravel 9 docblock source code of Eloquent\Builder](https://github.com/laravel/framework/blob/278c6d7fd04be005c1175a7eaea868976177600c/src/Illuminate/Database/Eloquent/Builder.php#L23)

**Breaking changes**

No breaking changes, only functionality is added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- fix: add @mixin Query\Builder to EloquentBuilder.stub by @davidjr82 in https://github.com/nunomaduro/larastan/pull/1379
+
 ## [2.2.0] - 2022-08-31
 
 ### Added
--- a/stubs/EloquentBuilder.stub
+++ b/stubs/EloquentBuilder.stub
@@ -5,6 +5,8 @@ namespace Illuminate\Database\Eloquent;
 /**
  * @template TModelClass of Model
  * @property-read static $orWhere
+ *
+ * @mixin \Illuminate\Database\Query\Builder
  */
 class Builder
 {


canvural:
Hi,

The underlying issue of #1346 is more complicated. And this fix is not correct. We already have [an extension](https://github.com/nunomaduro/larastan/blob/82d98695c097086717205f19f08e38b8b0542a63/src/Methods/EloquentBuilderForwardsCallsExtension.php) that handles the normal query builder calls on Eloquent builder.

Thank you though.

PR #1383, opened 2022-09-29
Added trans-string type which must be a defined translation key.

mad-briller:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Adds a new `trans-string` type that is similar to `view-string` but for translation keys.

Not sold on the `trans-string` name but tried to keep convention with `view-string`, perhaps `trans-key` would be a better name? open to suggestions.


canvural:
Sorry, but I'm not really convinced that this will add any value to Larastan. Plus I don't want the maintenance burden of this. 

But this would be a cool extra package!

Thanks for understanding.

Issue #1384, opened 2022-10-03
Larastan is not auto-registering service providers in packages.


mad-briller:
- Larastan Version: v2.2.0

### Description

Upon investigation i found the `getVendorDir` static method here:
https://github.com/nunomaduro/larastan/blob/master/src/ApplicationResolver.php#L39

When i do 

```
    /**
     * Creates an application and registers service providers found.
     *
     * @return \Illuminate\Contracts\Foundation\Application
     *
     * @throws \ReflectionException
     */
    public static function resolve(): Application
    {
        $app = (new self)->createApplication();

        $vendorDir = self::getVendorDir() ?? getcwd().DIRECTORY_SEPARATOR.'vendor';
        $composerConfigPath = dirname($vendorDir).DIRECTORY_SEPARATOR.'composer.json';

        var_dump($composerConfigPath, $app->basePath());
        // ...
```

i get this output:

```
string(101) "phar:///home/brad/code/esb-workspace/esb-foundation/vendor/phpstan/phpstan/phpstan.phar/composer.json"
string(84) "/home/brad/code/esb-workspace/esb-foundation/vendor/orchestra/testbench-core/laravel"
```

it appears that `getVendorDir` is returning the vendor dir of the PHPStan phar, rather than the local project. This causes larastan to read the `composer.json` from there, and fail to auto-register service providers.
This package used to work in the past (i'm not sure what version of phpstan / larastan) so along the way a change to phpstan / larastan has caused this to happen.

I suspect this is the root of my [view-string issue](https://github.com/nunomaduro/larastan/issues/1355) i posted a while ago.

Not sure how you'd go about fixing this as i'm not super familiar with the bootstrapping of larastan (yet)

mad-briller:
removing the call to `getVendorDir` and using the `getCwd` path instead makes the auto-discovery work again, but assuming that the user will always run `phpstan` from the root of their project is probably less than ideal.

szepeviktor:
Hello Brad!

`getVendorDir` locates `Composer\Autoload\ClassLoader` class.
Could it be that PHPStan PHAR file includes this class?? 🤯

canvural:
Just an idea, I added a ComposerHelper class to PHPStan a while back: https://github.com/phpstan/phpstan-src/blob/5cf39bc960b4b9dbc01be6906cbc1d71285118f7/src/Internal/ComposerHelper.php

Could we use the same logic here also?

mad-briller:
[...]

PR #1385, opened 2022-10-03
Fixes auto-discovery by resolving the correct vendorDir in package scenarios.

mad-briller:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

related: https://github.com/nunomaduro/larastan/issues/1384

**Changes**
Don't rely on the `ClassLoader` location but instead the `bootstrap.php` location instead.


szepeviktor:
This even passes "Test Laravel from other working directories"!

mad-briller:
which do you prefer? this or https://github.com/nunomaduro/larastan/issues/1384#issuecomment-1265373788 ?

either works for me, although i feel this one is the most simple.

szepeviktor:
I'm very happy that I must not answer that, as I'm just a janitor here 😌 
I have a feeling that `__DIR__` will break something in the far future.

szepeviktor (approved):

szepeviktor:
@mad-briller my own answer (not! for Larastan) is: wait 2 months and see

mad-briller:
yeah relying on filtering a list of custom autoloaders and hoping that the composer one is the first one we find that isn't in the `phar:` feels brittle aswell, whereas larastan controls the placement of the `bootstrap.php` file itself.

i suppose the one thing this doesn't handle is a custom `vendor-dir` configuration as we're assuming `vendor` is always in the root of the project.

szepeviktor:
> hoping that the composer one is the first one we find

`ClassLoader::getRegisteredLoaders` could be our primary and `__DIR__` our fallback.

mad-briller:
i'll wait for @canvural's input before continuing on this

mad-briller:
may be superceded by this: 
https://github.com/nunomaduro/larastan/pull/1391

canvural:
Closing because it is fixed in another way with #1392

PR #1389, opened 2022-10-03
feat:dont use ExtendedMethodReflection

### LINKED ISSUE #1289: Method calls on custom query builder to public parent methods are wrongly reported as private
- Larastan Version: 2.1.11
- `--level` used: 7
- Pull request with failing test: _\<none\>_

### Description

If one creates a trait to equip a model with a custom query builder, then methods calls on the custom query builder which are defined in `Illuminate\Database\Query\Builder` are wrongly reported as `private`.

I prepared the following GITHub repository with a minimal Laravel application: https://github.com/nagmat84/larastan-example. Step to reproduce the problem:

 1. Checkout the repository
 2. Install all dependencies via `composer install`
 3. Run PHPStan via `composer phpstan`

You will get the following output

```
 ------ ---------------------------------------------------------------------------------------------------------- 
  Line   Http/Controllers/Controller.php                                                                           
 ------ ---------------------------------------------------------------------------------------------------------- 
  18     Call to private method whereBetween() of parent class Illuminate\Database\Eloquent\Builder<TNodeModel of  
         App\Models\Contracts\Node&Illuminate\Database\Eloquent\Model>.                                            
 ------ ----------------------------------------------------------------------------------------------------------
```

However the method is not private.

### LINKED ISSUE #1386: Do not implement ExtendedMethodReflection in EloquentBuilderMethodReflection
I'd like to resurface my concern here :) We've finally taken advantage of ExtendedMethodReflection and added a new method there in https://github.com/phpstan/phpstan-src/pull/1317. So when PHPStan 1.9.0 is shipped, Larastan as in the current state is not going to be compatible with it.

Context: https://github.com/nunomaduro/larastan/commit/d233e1c3efe5de35430f5b1b1ea3a5a545e4c966

canvural:
This PR fixes compatibility issues with PHPStan 1.9.x.

1.9.x is needed because [this](https://github.com/phpstan/phpstan-src/commit/c052aace4efa157d9594f13fb748965cbfc43be2) commit is needed here.

Closes #1386 
Closes #1289

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "illuminate/support": "^9",
         "mockery/mockery": "^1.4.4",
         "phpmyadmin/sql-parser": "^5.5",
-        "phpstan/phpstan": "^1.9.x-dev"
+        "phpstan/phpstan": "^1.8.7"
     },
     "require-dev": {
         "nikic/php-parser": "^4.13.2",
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "illuminate/support": "^9",
         "mockery/mockery": "^1.4.4",
         "phpmyadmin/sql-parser": "^5.5",
-        "phpstan/phpstan": "^1.8.3"
+        "phpstan/phpstan": "^1.9.x-dev"
     },
     "require-dev": {
         "nikic/php-parser": "^4.13.2",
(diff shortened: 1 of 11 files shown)


ondrejmirtes:
Do you imply that strange commit I did actually enabled this change? Wow 😊

canvural:
> Do you imply that strange commit I did actually enabled this change? Wow blush

Yes! I tested multiple times before you released 1.8.7. So, 1.8.6 would fail but 1.9.x-dev would pass.

And not related to `ExtendedMethodReflection` issue, that commit fixed another thing. It's a really strange thing to explain, but it was causing issues for us even 2 years ago. I think [here](https://github.com/phpstan/phpstan/issues/4063) I tried to describe the issue. Basically our extension would get called multiple times with different types. And seeing that commit changed `getMethod` to `getNativeMethod` I think these are related 😄

ondrejmirtes:
Wow, what a coincidence, I could not have foreseen this :)

PR #1392, opened 2022-10-05
feat: use `Orchestra\Testbench\Foundation\Application` and allow custom `laravel` path based on `testbench.yaml`

crynobone:
- ~~Added or updated tests~~
- ~~Documented user facing changes~~
- [ ] Updated CHANGELOG.md

**Changes**

Allow developers to utilize custom Laravel stub via `testbench.yaml` if needed.

As an example, `laravel/nova` uses `laravel/nova-dusk-suite` as it's application stub and `laravel/nova` does have `testbench.yaml` with following:

```yaml
laravel: ./vendor/laravel/nova-dusk-suite
```

In tests and using Testbench Commander, everything points to `vendor/laravel/nova-dusk-suite` but Larastan still uses `vendor/orchestra/testbench-core/laravel` as the application stub. This fix that.

**Breaking changes**

None, all functionally provided from `orchestra/testbench`

### DIFF
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -9,21 +9,17 @@ use Composer\ClassMapGenerator\ClassMapGenerator;
 use const DIRECTORY_SEPARATOR;
 use Illuminate\Contracts\Foundation\Application;
 use function in_array;
-use Orchestra\Testbench\Concerns\CreatesApplication;
+use Orchestra\Testbench\Foundation\Application as Testbench;
 use ReflectionClass;
+use Symfony\Component\Yaml\Yaml;
 
 /**
  * @internal
  */
 final class ApplicationResolver
 {
-    use CreatesApplication;
-
     /** @var mixed */
     public static $composer;
-
-    /** @var bool */
-    protected $enablesPackageDiscoveries = true;
 
     /**
      * Creates an application and registers service providers found.
@@ -34,7 +30,15 @@ final class ApplicationResolver
      */
     public static function resolve(): Application
     {
-        $app = (new self)->createApplication();
+        if (file_exists(getcwd().'/testbench.yaml')) {
+            $appBasePath = transform(Yaml::parseFile(getcwd().'/testbench.yaml'), function ($config) {
+                return $config['laravel'];
+            });
+        }
+
+        $appBasePath ??= Testbench::applicationBasePath();
+
+        $app = Testbench::create($appBasePath, null, ['enables_package_discoveries' => true]);
 
         $vendorDir = self::getVendorDir() ?? getcwd().DIRECTORY_SEPARATOR.'vendor';
         $composerConfigPath = dirname($vendorDir).DIRECTORY_SEPARATOR.'composer.json';
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -38,7 +38,6 @@ final class ApplicationResolver
             $app = Testbench::create(options: ['enables_package_discoveries' => true]);
         }
 
-
         $vendorDir = self::getVendorDir() ?? getcwd().DIRECTORY_SEPARATOR.'vendor';
         $composerConfigPath = dirname($vendorDir).DIRECTORY_SEPARATOR.'composer.json';
 
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -5,9 +5,9 @@ declare(strict_types=1);
 namespace NunoMaduro\Larastan;
 
 use Illuminate\Contracts\Foundation\Application;
+use Illuminate\Foundation\PackageManifest;
 use Orchestra\Testbench\Foundation\Application as Testbench;
 use Orchestra\Testbench\Foundation\Config;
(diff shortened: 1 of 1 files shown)


src/ApplicationResolver.php
```diff
@@ -34,7 +30,15 @@ final class ApplicationResolver
...
     public static function resolve(): Application
     {
-        $app = (new self)->createApplication();
+        if (file_exists(getcwd().'/testbench.yaml')) {
+            $appBasePath = transform(Yaml::parseFile(getcwd().'/testbench.yaml'), function ($config) {
+                return $config['laravel'];
+            });
+        }
+
+        $appBasePath ??= Testbench::applicationBasePath();
```
szepeviktor, line 41: This could be so simple.
```suggestion
        $appBasePath = file_exists(getcwd().'/testbench.yaml')
            ? transform(Yaml::parseFile(getcwd().'/testbench.yaml'), function ($config) {
                return $config['laravel'];
            })
            : Testbench::applicationBasePath();
```

szepeviktor:
I think that Larastan should be able to enumerate service providers on its own.
A guy with harder balls will press <kbd>Close</kbd> below.

crynobone:
> A guy with harder balls will press Close below.

Is this comment even necessary? What a way to welcome contribution.

szepeviktor:
That explains why I don't dare to close this PR.

canvural:
> I think Larastan should be able to enumerate service providers on its own. 

This PR is not about finding service providers at all.

@crynobone , Only thing I don't like is the `symfony/yaml` dependency. Is there no way that this could be done in Testbench itself?

crynobone:
If you want to bump Testbench ^7.10 (upcoming version) then yes I can do this in Testbench. This would remove support for Laravel 9 releases lower than what latest Testbench consider as minimum, which is Laravel Framework 9.34.0

canvural:
We don't depend on `orchestra/testbench` anyway. We just [suggest it](https://github.com/nunomaduro/larastan/blob/52448d76a8b2372b600d1c52a6683126522abd2b/composer.json#L35) to the package developers if they want to use Larastan in their packages.

So, no problem in our side.

crynobone:
Sure, will work on it tomorrow.

crynobone (commented):
wip

canvural:
> > This PR is not about finding service providers at all.
> 
> Latest commit solved this.

I tested this on a simple Laravel package. And the root service provider is still not detected.

You can test it here: https://github.com/canvural/larastan-test When you run `vendor/bin/phpstan analyse` it shouldn't output any error.

canvural:
Also BTW, this is my solution that works: https://github.com/nunomaduro/larastan/blob/811ea63f4a6b6a41a3408e9ecf443e342a10c78e/src/ApplicationResolver.php#L30-L43 But it feels really hacky 😅

crynobone:
I tested with Nova source with above changes and everything work as expected.

https://user-images.githubusercontent.com/172966/195034433-e0a7387e-5ccf-407e-8ce0-b94fccf1fa3c.mp4

canvural:
> I tested with Nova source with above changes and everything work as expected (`bootstrap/cache/packages.php` regenerated just by running `phpstan analyse`)

Yes it generates the `bootstrap/cache/packages.php` file. But that file is missing the root service provider. That is the problem.

crynobone:
> Yes it generates the bootstrap/cache/packages.php file. But that file is missing the root service provider. That is the problem.

Fixed via https://github.com/nunomaduro/larastan/pull/1392/commits/04295bae8bffcadd20c7a03a7ce06b29110b45bf

canvural:
> > Yes it generates the bootstrap/cache/packages.php file. But that file is missing the root service provider. That is the problem.
> 
> Fixed via [04295ba](https://github.com/nunomaduro/larastan/commit/04295bae8bffcadd20c7a03a7ce06b29110b45bf)

Yep that fixes it! But now I see another problem 😄  

Now `bootstrap/cache/packages.php` only contains the root package from my testing. Can you confirm you still see the all packages (for example `spatie/laravel-ray`) in that file with the commit https://github.com/nunomaduro/larastan/commit/04295bae8bffcadd20c7a03a7ce06b29110b45bf ?

crynobone:
> Now bootstrap/cache/packages.php only contains the root package from my testing. Can you confirm you still see the all packages (for example spatie/laravel-ray) in that file with the commit 

Not able to replicate that.

canvural:
I found my issue. The dummy Laravel app under the `vendor/orchestra/testbench-core/laravel` doesn't have a `vendor` folder. Do you know what might have caused that?

canvural:
Ok, I think I found out. `vendor/bin/testbench package:discover` needs to be run at least one time. Either right after installation or in a Composer's after dump-autoload hook.

crynobone:
> Ok, I think I found out. vendor/bin/testbench package:discover needs to be run at least one time. Either right after installation or in a Composer's after dump-autoload hook.

Fixed via https://github.com/nunomaduro/larastan/pull/1392/commits/f98023717e5629bbaacaef07b0c8537b2c1a470f

PR #1394, opened 2022-10-10
feat: refined return type of Request::header method.

mad-briller:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Refines return type of the `\Illuminate\Http\Request::header` method.


canvural:
Thank you!

Issue #1397, opened 2022-10-11
False positives from NoUnnecessaryCollectionCallRule


mad-briller:
- Larastan Version: dev-master
- `--level` used: 8
- Pull request with failing test: https://github.com/nunomaduro/larastan/pull/1398

### Description

Since updating to `dev-master` to try out the recent fix for packages, i encountered some false positives with this rule, possibly related to this recent merged pr:
https://github.com/nunomaduro/larastan/pull/1375

canvural:
I reverted that PR. Thanks.

PR #1399, opened 2022-10-11
Revert #1375

canvural:
@zlayaAvocado I need to revert your PR unfortunately. It causes this issue: #1397 

Now I checked the code again, and yes this change would disallow all the calls on Collection if it's a property of a class. Ideally we would want to just disallow them on relations, but I couldn't find a clean way to do it. Please feel free to investigate more and send a new PR! Thanks.


zlayaAvocado:
@canvural no worries and thank you so much!

PR #1400, opened 2022-10-11
Delete `bootstrap/cache/packages.php` on vendor symlink.

crynobone:
- [ ] Updated CHANGELOG.md

**Changes**

Match changes to Testbench Core v7.10.1, this would avoid running `$packageManifest->build();` unless vendor requires re-running via `$filesystem->link()` 

**Breaking changes**


canvural:
Thank you!

PR #1402, opened 2022-10-13
Do not overwrite variable in getComposerJsonPath

szepeviktor:
robust and readable


canvural:
> @canvural Is an improvement a `fix:` or a `feat:`?

I think `feat:` would be fine.

szepeviktor:
Thank you.

PR #1404, opened 2022-10-14
fix: don't attempt to discover any packages during symlink.

crynobone:
solves nunomaduro/larastan#1345

- ~~[ ] Added or updated tests~~
- ~~[ ] Documented user facing changes~~
- [ ] Updated CHANGELOG.md

**Changes**
We shouldn't attempt to autoload root package providers during `vendor` symlink, it causes an issue when the root package depends on another service provider to be available.

**Breaking changes**

None

### DIFF
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -74,7 +74,7 @@ final class ApplicationResolver
         if (class_exists(Config::class)) {
             $config = Config::loadFromYaml($workingPath);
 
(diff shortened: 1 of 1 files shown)


szepeviktor:
Hello!

This is our own method.
https://github.com/nunomaduro/larastan/blob/dedb62bfa85b9a87b67760a5a8079c67140dd827/src/ApplicationResolver.php#L25

canvural:
Awesome! I can confirm it fixes the issue in the `laravel-table` package. Thank you!

Issue #1366, opened 2022-09-04
non-falsy-string breaks Attribute testing


joostdebruijn:
- Larastan Version: 2.2.0 (on phpstan v1.8.4)
- `--level` used: 5

### Description

Since phpstan v1.8.3, the following error occurs:
```
Parameter $get of static method Illuminate\Database\Eloquent\Casts\Attribute<mixed,mixed>::make() expects 
  (callable(mixed, mixed=): string)|null,
  Closure(mixed, mixed): non-falsy-string given.
```
The `non-falsy-string` is introduced in phpstan v1.8.3. I think this test should pass, but it doesn't.

### Laravel code where the issue was found

```php

    /**
     * Returns the full name of a person.
     *
     * @return \Illuminate\Database\Eloquent\Casts\Attribute<string, never>
     */
    protected function fullname(): Attribute
    {
        return Attribute::make(
            get: function ($value, $attributes): string {
                if (! is_null($attributes['prefix'])) {
                    return $attributes['firstname'].' '.$attributes['prefix'].' '.$attributes['lastname'];
                }

                return $attributes['firstname'].' '.$attributes['lastname'];
            },
        );
    }

```

szepeviktor:
Hello Joost! 👋🏻
Thank you for your report.

joostdebruijn:
@szepeviktor Thanks for you blazing fast reply! I reported over there. Would you prefer to close this one for now or just leave it open untill they've responded upstream?

ondrejmirtes:
The problem isn't the non-falsy-string part, the problem is that the code expects `callable(mixed, mixed=)` (one required parameter, one optional parameter), but your Closure requires two parameters.

This is a problem because the code that accepts the callable might call it only with one argument but when you give it a callable with two required parameters, PHP will crash.

mreduar:
So this is a phpstan problem? I am experiencing exactly this problem as well.
```php
protected function fullImageUrl(): Attribute
{
    return Attribute::make(
        get: fn ($value, $attributes) => global_asset($attributes['image_url']),
    );
}

szepeviktor:
AFAIK the given callback may not get a second argument `$attributes`.
https://github.com/nunomaduro/larastan/blob/8fbe7423782dac778f2383f8cc895b0da0faf733/stubs/Attribute.stub#L30

szepeviktor:
Theory says it should have a default value `fn ($value, $attributes = [])`
[...]

PR #1410, opened 2022-10-17
fix: `Attribute::make` callable signature

szepeviktor:
@mreduar Let see what CI has to say.

From https://github.com/nunomaduro/larastan/issues/1366#issuecomment-1281375778


mreduar:
Failed two checks, what does it mean?

szepeviktor:
> Failed two checks, what does it mean?

If you take a look at the output. It is about this method.

canvural:
@mad-briller Makes sense.

Thank you!

PR #1415, opened 2022-10-19
feat: test on PHP 8.2

canvural:
Test on PHP 8.2


szepeviktor:
Lowest deps on PHP 8.2 caused problems?

canvural:
> Lowest deps on PHP 8.2 caused problems?

Yeah. Of course Laravel 9.0.0 is not compatible with PHP 8.2

PR #1418, opened 2022-10-19
feat: run StyleCI on stubs also

szepeviktor:
@GrahamCampbell Could you help us why stub files cause internal errors?
https://github.styleci.io/analyses/JGg3pV

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,1 +1,5 @@
 preset: laravel
+finder:
+  name:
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Because those stub files are not valid PHP syntax.

GrahamCampbell:
They pass the token_get_all linter, but not `php -l`.

PR #1419, opened 2022-10-19
feat: conditional return types for `Eloquent\Collection::find()`

sebdesign:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

This was originally submitted here: https://github.com/laravel/framework/pull/44459

**Changes**

Thanks to PHPStan's conditional return types, we can now infer what does the `find` method of the Eloquent Collection returns.
When the `$key` argument is an `array` or an `Arrayable` object, it returns a new collection with the found items.
In all other cases, it returns a single item, or the `$default` value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - feat: Added stub for `optional()` helper and class by @mad-briller in https://github.com/nunomaduro/larastan/pull/1344
 - fix: abstract Manager class causing Larastan to crash by @mad-briller
 - fix: Eloquent builder `whereRelation()` losing the TModelClass generic type by @mad-briller.
+- feat: conditional return types for `Eloquent\Collection::find()` by @sebdesign
 
 ## [2.2.0] - 2022-08-31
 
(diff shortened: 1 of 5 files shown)


szepeviktor:
> we can now infer what does the `find` method of the Eloquent Collection returns

Oh yes.
Thank you for this PR.

szepeviktor (approved):

canvural (changes requested):
Thank you.

But also can you add tests to [this file](https://github.com/nunomaduro/larastan/blob/507386544b81009724dadf56956b16de164b5940/tests/Type/data/collection-generic-static-methods.php)? So we'll make sure it also works fine with the custom eloquent collections. If the tests doesn't pass, you might need to add `find` to the array [here.](https://github.com/nunomaduro/larastan/blob/507386544b81009724dadf56956b16de164b5940/src/ReturnTypes/CollectionGenericStaticMethodDynamicMethodReturnTypeExtension.php#L34-L44)

sebdesign:
Hey @canvural, I've tested the return type for custom Eloquent collections.

Instead of adding the method name on the list for the base collection class, we need to check the method name on the eloquent collection class, because that's where the method is declared.

canvural:
Thank you! Can you also resolve the conflict? Then if the CI is green, we can merge!

canvural:
Thank you!

PR #1420, opened 2022-10-21
feat: refined return type of DB::transaction method

jdjfisher:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Refines the return type of the `\Illuminate\Support\Facades\DB::transaction` method.

### DIFF
--- a/stubs/Database/ManagesTransactions.stub
+++ b/stubs/Database/ManagesTransactions.stub
@@ -10,7 +10,7 @@ trait ManagesTransactions
      * @template TCallbackReturnType
      *
      * @param  \Closure($this): TCallbackReturnType  $callback
-     * @param  int  $attempts
+     * @param  positive-int  $attempts
      * @return TCallbackReturnType
      *
      * @throws \Throwable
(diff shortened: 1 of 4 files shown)


jdjfisher:
It would be great to also do something similar for `Cache::remember`php  as I find myself having to explicitly annotate the return type for that quite often. That said, I'm not sure if its possible to refine it while being entirely true to the implementation. e.g

```php
Cache::put('the-key', 'not-an-int');

$probably_an_int_ = Cache::rememberForever('the-key', fn () => rand());
```

Thoughts?

stubs/Database/ManagesTransactions.stub
```diff
@@ -0,0 +1,20 @@
...
+
+trait ManagesTransactions
+{
+    /**
+     * Execute a Closure within a transaction.
+     *
+     * @template TCallbackReturnType
+     *
+     * @param  \Closure($this): TCallbackReturnType  $callback
+     * @param  int  $attempts
```
mad-briller, line 13: You could make this positive-int as making 0 attempts makes no sense

jdjfisher: Good idea 👍

szepeviktor (approved):

canvural (approved):

canvural:
Thank you!

PR #1423, opened 2022-10-24
feat: unused views collector rule

canvural:
- [x] Added or updated tests
- [ ] Updated CHANGELOG.md

This PR adds a new rule that can find the unused views, using PHPStan's collector feature. It'll be an optional rule.

### Some concerns

- Collectors for this feature tries to find used views in 2 places currently: `view('foo')` calls. And `$this->markdown('foo')` in any `Mailable` class. Are there more places that we can check?
- Currently there are false positives if a view is only used in another view. For example with `@extends('foo')` I'd say this is an edge case, and can be ignored by the user.

### Todos
- [ ] [This](https://github.com/nunomaduro/larastan/actions/runs/3312375217/jobs/5468975281) error should not happen. The mentioned views are from the dummy Laravel project Testbench created, not from the actual package.
- [x] Support `View::make('foo')` and `$viewFactory->make()`
- [x] Support `$this->view('foo')` in `Mailable`s

## How to test it?
Run `composer require --dev nunomaduro/larastan:dev-unused-views` Then run your PHPStan analysis again.
And report here if there are any false positives or errors that we can fix!


mad-briller:
> Are there more places that we can check?
 
Alongside the view facade, the `View\Factory` [contract](https://laravel.com/api/9.x/Illuminate/Contracts/View/Factory.html) may be injected and `$viewFactory->make()` called instead, so may be worth checking there too

canvural:
Thanks, updated the list.

PR #1430, opened 2022-10-30
Less arguments passed to ScopeFactory

ondrejmirtes:
The `$methodReflection` is immediately rewritten anyway...

I'm changing the ScopeFactory a bit in PHPStan 1.9.0 but with this change the BC break isn't as concerning :)


szepeviktor (approved):

canvural:
Thank you!

PR #1431, opened 2022-11-01
fix: PHPStan 1.9.0 compatibility

canvural:
This PR adds PHPStan 1.9.x compatibility.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "illuminate/support": "^9",
         "mockery/mockery": "^1.4.4",
         "phpmyadmin/sql-parser": "^5.5",
-        "phpstan/phpstan": "^1.8.7"
+        "phpstan/phpstan": "^1.9.0"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


agi--:
Is it possible to apply that patch also to 1.0.3? We still need to use that version because we haven't yet switched to Laravel 9.x.

canvural:
No. 1.x releases are not maintained any more. It lacks a lot of fixes, not just this one. 

You can use your own fork with this fix though.

agi--:
Ok, we can do that, but I think the better approach would be to upgrade Laravel as soon as possible. Thanks for the fast reply.

NickStallman:
I've also got a few Laravel 8 projects and I'm hitting this issue. It would be appreciated if it could be backported, or at least make Larastan 1.x block PHPStan 1.9 so it doesn't completely break.

szepeviktor:
@NickStallman Hello! Thank you for commenting.
Please add `"phpstan/phpstan": "< 1.9",` to your Composer configuration.

PR #1433, opened 2022-11-02
Organize E2E tests

szepeviktor:
- keep data only in matrix
- tricky trick: `composer exec phpstan`
- removed unnecessary `composer update` on Larastan
- removed `--ignore-platform-reqs`


canvural:
Thank you!

PR #1439, opened 2022-11-06
feat: support for encrypted and parameterized casts

jdjfisher:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Added support for resolving [encrypted eloquent cast](https://github.com/laravel/framework/blob/c76c3bdddf3bac41f92bbcaddb50d1053d0011f2/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1546) types.

Added support for resolving underlying eloquent cast types with cast parameters (e.g decimal with precision, datetime with format etc).

### DIFF
--- a/src/Properties/ModelPropertyExtension.php
+++ b/src/Properties/ModelPropertyExtension.php
@@ -237,12 +237,16 @@ final class ModelPropertyExtension implements PropertiesClassReflectionExtension
                 $type = Str::after($type, 'encrypted:');
             }
 
+            // Truncate cast parameters
+            $type = Str::before($type, ':');
+
             switch ($type) {
                 case 'boolean':
                 case 'bool':
                     $realType = 'boolean';
                     break;
                 case 'string':
+                case 'decimal':
                     $realType = 'string';
                     break;
                 case 'array':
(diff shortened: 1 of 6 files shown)


src/Properties/ModelPropertyExtension.php
```diff
@@ -232,12 +232,21 @@ private function castPropertiesType(Model $modelInstance): void
                 continue;
             }
 
+            // Reduce encrypted castable types
+            if (in_array($type, ['encrypted', 'encrypted:array', 'encrypted:collection', 'encrypted:json', 'encrypted:object'], true)) {
+                $type = Str::after($type, 'encrypted:');
+            }
+
+            // Truncate cast parameters
+            $type = Str::before($type, ':');
```
jdjfisher, line 241: Looking through the framework, it appears to be explicitly looking for `decimal:`, `date:`, `datetime:` etc for the built-in parameterized casts. That said, I think for the interest maintainibility here, we can reasonably assume that everything following the `:` is parameters and not part of the underlying cast type.

canvural:
Thank you! Looks good.

PR #1443, opened 2022-11-09
Removed duplicated method pop() in Collection.stub

miloshavlicek:
Reason: PHPStan reports that pop() method is already defined in Collection.stub

- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

- Removed duplicated method pop() in Collection.stub

**Breaking changes**

- No breaking changes


canvural:
We do not accept any fixes for 1.x branch. Sorry. Please upgrade Larastan or downgrade PHPStan version to 1.8.x

PR #1444, opened 2022-11-11
Improve handling of read types on custom model casts

Aberdeener:
- [x] Added or updated tests
- [ ] Documented user facing changes (I don't think this is needed?)
- [x] Updated CHANGELOG.md

**Changes**
This PR fixes a pretty annoying issue while using Larastan and model casts.
For example, when using the [Laravel Money](https://github.com/cknow/laravel-money) package, it has a custom cast class `\Cknow\Money\Casts\MoneyIntegerCast`.
This casts `get` method returns a `\Cknow\Money\Money` object, but Larastan only detected the `MoneyIntegerCast` as the return type.
This causes dozens of errors similar to:
```
Parameter #1 $price of static method                              
         `App\Helpers\TaxHelper::calculateFor()` expects `Cknow\Money\Money`
         `Cknow\Money\Casts\MoneyIntegerCast` given.
```
To fix this issue, I have used reflection to determine the return type of the casts underlying `get` method, and (if it is defined), use that as the read type for the attribute.

I am aware of this PR https://github.com/nunomaduro/larastan/pull/1333, but I thought a smaller PR to fix this problem faster would be nice, until that one gets merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - feat: conditional return types for `Eloquent\Collection::find()` by @sebdesign
 - feat: add support for generic paginators by @erikgaal
 - feat: add support for encrypted and parameterized eloquent casts by @jdjfisher
+- fix: improve handling of read types on custom model casts by @Aberdeener
 
 ## [2.2.0] - 2022-08-31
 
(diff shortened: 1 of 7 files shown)


canvural:
Since #1333 is merged, we no longer need this I guess. Please send a new PR if you see that something can be improved. Thanks.

PR #1447, opened 2022-11-15
feat: container binding implements abstract rule

jdjfisher:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Adds a new collector and rule to check that container bindings make sense.

Questions:
- Should this rule be optional?
- Add support for `$bindings` and `$singletons` defined on Service Providers?
- Should this check that the concrete type is either an instance or instantiable?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .idea/*
+.vscode/
 .idea/codeStyleSettings.xml
 composer.lock
 /vendor/
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -137,6 +137,31 @@ And if you call the function above with a property that does not exist in User m
 takesOnlyUserModelProperties('emaiil');
 ```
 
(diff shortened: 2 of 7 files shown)


canvural:
Why is this done with collector? Looks like it only checks method calls, so it can be done with a normal rule.

Also, I'm not sure if this is what we really want. Like you see in the failing E2E test, it also complains when `mixed` is passed as the concrete type.

What about:
1) Make it an optional rule. I just worry it'd annoy people if it was required.
2) Do the check only if the abstract type is an interface. And check if the given concrete type implements that interface. Although binding completely different class to another class is not recommended, I think some use cases can be valid. So just checking interfaces makes more sense.

What do you think?

canvural:
Hi,

I don't think this is a very useful rule. So I'll not merge it this time.

Thank you.

PR #1450, opened 2022-11-20
Fix CS

szepeviktor:
We have 100 `function foo();`
and 27 `function bar() {}`

What to do now?


canvural:
I think every stub should use `;`

So can you update this PR to fix everything?

szepeviktor:
Oh no! Only interfaces may have `function foo();`

canvural:
Not only interfaces but also methods. So only functions should have `{}`

szepeviktor:
> Not only interfaces but also methods.

I've tried in a commit above. Everything went berserk :x:

canvural:
Thank you!

Issue #1441, opened 2022-11-07
Illuminate\Foundation\Application in Service provider closure type not resolving


XbNz:
- Larastan Version: 2.2.9
- `--level` used: Max

### Description

Return type of `$app->make(Foo::class)` is not inferred, so PHPStan throws a "mixed given" error

### Laravel code where the issue was found

```php
$this->app->bind(ExampleService::class, function (Application $app) {
      return new ExampleService(
          $app->make(Foo::class),
      );
});

```

mad-briller:
how have i never noticed this is not supported, spent an hour today where i gaslit myself into thinking it was

PR #1451, opened 2022-11-21
fix: `make`, `makeWith` and `resolve` methods not resolving to correct type on Application and Container.

mad-briller:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

resolves https://github.com/nunomaduro/larastan/issues/1441

**Changes**
Added return type extension for Application and Container `make`, `makeWith` and `resolves` method.

Abstracts repeated behavior into a `AppMakeHelper` and uses that in places i could find it being duplicated.

### DIFF
--- a/src/ReturnTypes/AppMakeHelper.php
+++ b/src/ReturnTypes/AppMakeHelper.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace NunoMaduro\Larastan\ReturnTypes;
+
+use NunoMaduro\Larastan\Concerns\HasContainer;
+use PhpParser\Node\Expr\ClassConstFetch;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Name\FullyQualified;
+use PhpParser\Node\Scalar\String_;
+use PHPStan\Type\ErrorType;
+use PHPStan\Type\NeverType;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\Type;
+use Throwable;
+
+final class AppMakeHelper
+{
+    use HasContainer;
+
+    public function resolveTypeFromCall(FuncCall|MethodCall|StaticCall $call): Type
+    {
+        if (count($call->getArgs()) === 0) {
+            return new ErrorType();
+        }
+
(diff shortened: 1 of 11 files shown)


src/ReturnTypes/AppMakeHelper.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        if (count($call->getArgs()) === 0) {
+            return new ErrorType();
+        }
+
+        $expr = $call->getArgs()[0]->value;
+        if ($expr instanceof String_) {
+            try {
+                /** @var object|null $resolved */
+                $resolved = $this->resolve($expr->value);
```
mad-briller, line 32: it's not really guaranteed that `resolve` will return an object, scalars can be bound into the container using names too and i've seen that in the wild, so it might be worth expanding this code to anticipate more types to be returned in the future

mad-briller:
not sure if the `koel/koel` e2e failures are caused by this change

JanMikes:
Hi, i have checked the failing E2E test and actually this is correct that test is failing.

https://github.com/koel/koel/blob/master/tests/Traits/CreatesApplication.php#L25

```
    use App\Console\Kernel;
    use Illuminate\Contracts\Console\Kernel as Artisan;

    private Kernel $artisan;

    $this->artisan = $app->make(Artisan::class);
```

Because this PR adds phpstan functionality that makes it understand that `$app->make(Artisan::class)` will return instance of `Artisan` (the interface) and the property is typed to specific instance. 

I will try send a PR to fix this issue in original repo.

mad-briller:
@canvural is there anything i can do to help get this over the line? thanks

JanMikes:
Pls can you re-run the E2E pipeline? Should be passing now.

szepeviktor:
> Pls can you re-run the E2E pipeline?

Here you go. This is the rerun.
https://github.com/nunomaduro/larastan/actions/runs/3684486433/jobs/6280772909

JanMikes:
I see - though the dependent PR is merged, there is fixed git reference: https://github.com/nunomaduro/larastan/blob/master/.github/workflows/e2e-tests.yml#L40

We would need to change it to `9e864c937f5f060aa0192596d4e0776e68dc16e5` (references to https://github.com/koel/koel/commit/9e864c937f5f060aa0192596d4e0776e68dc16e5)

canvural:
Thank you!

PR #1462, opened 2022-11-29
feat: add `createOneQuietly` stub

amooij:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

The factory stub did not have createOneQuietly defined, which prevented PHPStan/Larastan from resolving the type correctly.

**Breaking changes**

None

### DIFF
--- a/tests/Type/data/model-factories.php
+++ b/tests/Type/data/model-factories.php
@@ -8,6 +8,7 @@ use function PHPStan\Testing\assertType;
 assertType('Database\Factories\UserFactory', User::factory());
 assertType('Database\Factories\UserFactory', User::factory()->new());
 assertType('App\User', User::factory()->createOne());
+assertType('App\User', User::factory()->createOneQuietly());
 assertType('Illuminate\Database\Eloquent\Collection<int, App\User>', User::factory()->createMany([]));
 assertType('App\User', User::factory()->makeOne());
 
--- a/tests/Type/data/model-factories.php
+++ b/tests/Type/data/model-factories.php
@@ -35,11 +35,6 @@ assertType('App\User', User::factory(2)->count(null)->createQuietly());
 assertType('App\User', User::factory(2)->count(null)->createOneQuietly());
 assertType('App\User', User::factory(2)->count(null)->make());
 
-assertType('App\User', User::factory(2)->count(null)->create());
-assertType('App\User', User::factory(2)->count(null)->createQuietly());
-assertType('App\User', User::factory(2)->count(null)->createOneQuietly());
-assertType('App\User', User::factory(2)->count(null)->make());
-
 assertType('App\User', User::factory(2)->state(['foo'])->count(null)->create());
 assertType('App\User', User::factory(2)->state(['foo'])->count(null)->createQuietly());
 assertType('App\User', User::factory(2)->state(['foo'])->count(null)->createOneQuietly());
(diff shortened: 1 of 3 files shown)


canvural:
This needs tests added in https://github.com/nunomaduro/larastan/blob/0c9c70ff21e70ac8a90c0b1680cb5a7e5352d0ca/tests/Type/data/model-factories.php

amooij:
> This needs tests added in https://github.com/nunomaduro/larastan/blob/0c9c70ff21e70ac8a90c0b1680cb5a7e5352d0ca/tests/Type/data/model-factories.php

Like so?

szepeviktor (approved):

tests/Type/data/model-factories.php
```diff
@@ -8,6 +8,7 @@
 assertType('Database\Factories\UserFactory', User::factory());
 assertType('Database\Factories\UserFactory', User::factory()->new());
 assertType('App\User', User::factory()->createOne());
+assertType('App\User', User::factory()->createOneQuietly());
```
canvural, line 11: Instead of adding this here, please add similar tests after every `createQuietly`

canvural (changes requested):

tests/Type/data/model-factories.php
amooij: Can you help me understand why you requested this change?
My thinking was that createOneQuietly is more like createOne than createQuietly.

canvural: Yes, it is more like `createOne`. I just wanted you to add more tests. Which you can do it by adding them just below of every `createQuietly`

Like `assertType('Illuminate\Database\Eloquent\Collection<int, App\User>', User::factory(2)->createOneQuietly());` and `assertType('App\User', User::factory()->createOneQuietly());`

amooij:
```
 assertType('Illuminate\Database\Eloquent\Collection<int, App\User>', User::factory(2)->createOneQuietly());
```
Are you sure about this? I think this is broken usage, why would you tell the factory to create two items and then ask it for one?
The latter of the two is already there, if do you want to duplicate it?

canvural: While it doesn't make sense, it is a valid usage. It's not broken. We would test that `createOneQuietly` still returns single model even after calling `User::factory(2)`

The two I wrote above are just examples. I want you to add new test cases based on the previous test cases. If you read the file you'll notice a pattern. So I just want you to add `createOneQuietly` tests like the rest of them/

amooij: I'm not sure which expectations we need to verify in these additional tests. The createQuietly method has two return types, Model and Collection<int,Model>, I see how you would want to test both. createOneQuietly only has one return type. 
I also see that there is only one test for createOne, possibly for the same reason. 
I hope I'm not annoying with all my questions, but I'm trying to understand the why.

canvural: It is just additional safety. While unlikely, I don't want to break something in the future accidentally. So having these tests would give us a confidence and doesn't cost anything 😄

amooij: I added the tests. I saw that lines 33-37 are equal to lines 38-41. Is this intentional?

canvural: No, it's not intentional. Must be mistake. You can delete one of them.

amooij: I deleted them.

canvural:
Thank you!

Issue #1463, opened 2022-11-30
Passing string to view() function generates warning that the type should be view-string|null


robopzet:
- Larastan Version: 2.2.7
- `--level` used: 9
- Pull request with failing test: -

### Description
PHPStan generates this warning for the given code:

    Parameter #1 $view of function view expects view-string|null, string given.

The definition of the Laravel `view()` function describes the first parameter as `string|null`. The warning says it should be of type `view-string|null`. I do not understand why passing a `string` is not valid here.

### Laravel code where the issue was found

```php
    /**
     * @param string $fil
     * @param array<string, mixed> $dta
     * @return View
     */
    private function vw (
        string  $fil,
        array   $dta
    ): View {
        return view(sprintf(Settings::VW_BAS_TPL, $fil), $dta);
    }

    // Definition of Settings::VW_BAS_TPL:
    public const VW_BAS_TPL = 'path/to/templates/%s';
```

szepeviktor:
Hello @robopzet! 👋🏻 
view-string is our specialty :)
https://github.com/nunomaduro/larastan/blob/master/docs/custom-types.md

Please suppress the error. It does not support generated view names.

szepeviktor:
Or you have force the type.
```php
        /** @var view-string $viewName */
        $viewName = sprintf(Settings::VW_BAS_TPL, $fil);
        return view($viewName, $dta);
```

mad-briller:
if this is merged you can use `view()->exists($viewName)` to check it is a `view-string`:
https://github.com/nunomaduro/larastan/pull/1464

robopzet:
Thanks for the suggestions, and the quick PR!

PR #1464, opened 2022-11-30
Assert view-string with view()->exists().

mad-briller:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

Resolves https://github.com/nunomaduro/larastan/issues/1463 https://github.com/nunomaduro/larastan/discussions/1461

Asserts a string is a `view-string` when `exists()` is called on it.


canvural:
Thank you!

PR #1483, opened 2022-12-16
Prevent false positive for NoUnnecessaryCollectionCallRule when using arrow functions

JeppeKnockaert:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

Using arrow functions in `contains` and `first` no longer results in false positives for NoUnnecessaryCollectionCallRule.


canvural:
Thank you!

PR #1485, opened 2022-12-16
fix: support extended schema classes in schema aggregator

tpetry:
- [ ] Added or updated tests
- [ ] Updated CHANGELOG.md

**Changes**

The `SchemaAggregator` inspects all migrations to learn which Eloquent attributes do exist. But the schema and blueprint classes had been hardcoded so any class extending them was not supported. There are some packages extending the Laravel's migration classes and the second they are used larastan was not able to infer the model attributes anymore (e.g. tpetry/laravel-postgresql-enhanced#43). I've changed the implementation slightly to check whether the `Schema` and `Blueprint` class are the original Laravel classes or any class extending them.

### DIFF
--- a/src/Properties/SchemaAggregator.php
+++ b/src/Properties/SchemaAggregator.php
@@ -67,14 +67,14 @@ final class SchemaAggregator
                 && $stmt->expr->var->class instanceof PhpParser\Node\Name
                 && $stmt->expr->var->name instanceof PhpParser\Node\Identifier
                 && ($stmt->expr->var->name->toString() === 'connection' || $stmt->expr->var->name->toString() === 'setConnection')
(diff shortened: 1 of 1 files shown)


canvural:
I fixed it like this and added a test: 3eb62a4

Thank you!

tpetry:
That looks a lot cleaner than my solution!

tpetry:
@canvural The fix is not complete. The `Blueprint` class also needs to have that more generic handling:  [SchemaAggregator.php#L128](https://github.com/nunomaduro/larastan/blob/3eb62a4398130b2684fc78f590b30ef9690c0315/src/Properties/SchemaAggregator.php#L128)

canvural:
Ah I missed that. I'll add later. Thanks.

PR #1486, opened 2022-12-23
fix: incorrect EloquentBuilder::where() return type

taka-oyama:
- [x] Added or updated tests
- [x] Updated CHANGELOG.md

**Changes**

This changes the return type of `where` in the `EloquentBuilder.stub` which did not match the original file.

This was throwing false positives in my custom Builder class.

For example, the following code.
```
{
    /** @return $this */
    public function visible()
    {
        return $this->where('visible', true);
    }
}
```

Was giving me this error.

```
Method App\Models\Builders\UserBuilder::visible() should return $this(App\Models\Builders\UserBuilder) but returns static(App\Models\Builders\UserBuilder).
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


canvural:
`static` is correct here. You can just remove your return doctag from your method.

Issue #1487, opened 2022-12-23
Error with new rule `checkUnusedViews`


rcerljenko:
- Larastan Version: 2.3.0
- `--level` used: 6

### Description

After updating to v2.3 and enabling rule `checkUnusedViews` this happens:

```
In ViewFileHelper.php line 70:

Undefined array key 1
```

### Laravel code where the issue was found

This is from your code that's reported above:

```php
if (str_contains($view->getPathname(), 'views/vendor')) {
       continue;
 }

$viewName = explode($viewDirectory.'/', $view->getPathname());

yield str_replace(['/', '.blade.php'], ['.', ''], $viewName[1]); // THIS IS LINE 70
```

rcerljenko:
myabe it has something to do with windows - different directory separator  => `DIRECTORY_SEPARATOR` constant

rcerljenko:
yep.. just tried it with that constant and it worked... should i provide a PR for that?

canvural:
> yep.. just tried it with that constant and it worked... should i provide a PR for that?

Yes please.

rcerljenko:
here it is @canvural https://github.com/nunomaduro/larastan/pull/1488

canvural:
Released in https://github.com/nunomaduro/larastan/releases/tag/2.3.1 Thank you

PR #1488, opened 2022-12-23
Fix for issue #1487

rcerljenko:
Fixes https://github.com/nunomaduro/larastan/issues/1487

### DIFF
--- a/src/Support/ViewFileHelper.php
+++ b/src/Support/ViewFileHelper.php
@@ -65,9 +65,9 @@ final class ViewFileHelper
                     continue;
                 }
 
-                $viewName = explode($viewDirectory.'/', $view->getPathname());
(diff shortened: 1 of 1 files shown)


canvural:
Thank you!

rcerljenko:
np.. since we're here could you somehow exclude a folder from that rule? eg: I want to check my views but i dont want to check `resources/views/vendor` views because mostly i don't use them but they get published witrh everything else...

canvural:
`views/vendor` should be skipped already. Thanks to this [line.](https://github.com/nunomaduro/larastan/blob/master/src/Support/ViewFileHelper.php#L64) Does it not work for you?

rcerljenko:
again... directory separator haha  :) i'll provide another PR for that

rcerljenko:
here it is => https://github.com/nunomaduro/larastan/pull/1489

PR #1494, opened 2022-12-28
feat: handle native enum casts

erikgaal:
Extends [my previous PR](https://github.com/nunomaduro/larastan/pull/1333), where I forgot to implement support for native enum casts.

**Changes**

**Breaking changes**

### DIFF
--- a/src/Properties/ModelCastHelper.php
+++ b/src/Properties/ModelCastHelper.php
@@ -60,6 +60,10 @@ class ModelCastHelper
 
         $classReflection = $this->reflectionProvider->getClass($cast);
 
+        if ($classReflection->isEnum()) {
+            return new ObjectType($cast);
+        }
+
         if ($classReflection->isSubclassOf(Castable::class)) {
             $methodReflection = $classReflection->getNativeMethod('castUsing');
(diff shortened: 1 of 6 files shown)


szepeviktor (approved):

canvural:
Weird, how did this pass in the CI with PHP 8.0? 🤔 

Anyway, can you also add a test with backed enums?

levu42:
What's the status of this PR? Is there something missing that I can help to provide? The bug fixed by this PR is still causing a bunch of false positives on my side, which is quite annoying, so: how can I help accelerate it to get it merged & released?

erikgaal:
@canvural I've added a test for backed enums.

@levu42 The PR is ready now. 😄

szepeviktor:
@erikgaal Could it be that `enum BackedEnumeration : string` is not executed but analyzed on PHP 8.0 that is why it passes?

erikgaal:
@szepeviktor Yes, that's completely correct. It's not using the native PHP reflection (which would crash on 8.0), but the `BetterReflection` which comes from the PHPStan parser. 😄

canvural:
This looks good. I'm merging it now, and gonna add some more test cases for the writeable type later.

Thank you.

levu42:
@canvural Do you have an estimation for when a release will be made with this included?

PR #1496, opened 2022-12-29
Ignore "errors" folder in unused views rule

rcerljenko:
This is an update to my previous [PR](https://github.com/nunomaduro/larastan/pull/1488) which resolved a windows-style directory separator.

Now the situration is where you have a custom HTTP error views just like in the [laravel documentation](https://laravel.com/docs/9.x/errors#custom-http-error-pages). You don't have a reference to these views in the code (controllers etc.) but you still need them and laravel uses them for the exception rendering.

So the easiest approach is to just ignore `errors` folder just like we ignore the `vendor` one.

### DIFF
--- a/src/ReturnTypes/BuilderModelFindExtension.php
+++ b/src/ReturnTypes/BuilderModelFindExtension.php
@@ -81,7 +81,11 @@ final class BuilderModelFindExtension implements DynamicMethodReturnTypeExtensio
         MethodReflection $methodReflection,
         MethodCall $methodCall,
         Scope $scope
-    ): Type {
+    ): ?Type {
+        if (count($methodCall->getArgs()) < 1) {
+            return null;
+        }
+
         /** @var ObjectType $model */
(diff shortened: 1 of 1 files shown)


szepeviktor:
Thank you Ricardo!

Could you rearrange your commit to avoid merge commits?

rcerljenko:
@szepeviktor how can i do that exactly? (never done this before...)

szepeviktor:
Something like `git reset HEAD~5` and `git commit -a`

But do not listen to me! https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

rcerljenko:
thanks i just tried it but it's giving me a headache...  :/

szepeviktor:
Thank you for trying.

rcerljenko:
Feel free to adjust it if you can... i'm kinda stuck here

szepeviktor:
Yes, it is really-really hard.

kép

commits from master were brought in by merge.

Fixing now ... 🏃🏻

rcerljenko:
Thanks man!

PR #1512, opened 2023-01-09
feat(refactor): remove Mockery dependency

canvural:
This PR removes Mockery dependency

### DIFF
--- a/src/Methods/Passable.php
+++ b/src/Methods/Passable.php
@@ -6,12 +6,11 @@ namespace NunoMaduro\Larastan\Methods;
 
 use Illuminate\Contracts\Pipeline\Pipeline;
 use LogicException;
-use Mockery;
 use NunoMaduro\Larastan\Concerns;
 use NunoMaduro\Larastan\Contracts\Methods\PassableContract;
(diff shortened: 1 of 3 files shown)


src/Methods/Passable.php
```diff
@@ -189,14 +187,7 @@ function (PassableContract $passable) use ($originalClassReflection) {
...
         if ($result = $this->hasFound()) {
-            $methodReflection = $this->getMethodReflection();
-            if (get_class($methodReflection) === PhpMethodReflection::class) {
-                $methodReflection = Mockery::mock($methodReflection);
-                $methodReflection->shouldReceive('isStatic')
-                    ->andReturn($this->isStaticAllowed());
-            }
-
-            $this->setMethodReflection($methodReflection);
+            $this->setMethodReflection($this->getMethodReflection());
```
nunomaduro, line 190: And tests are passing? If yes, why I was using it in the first place? 🧐

canvural: Yeah I just wanted to see what is failing. Then I'll try to find a way to fix it.

PR #1522, opened 2023-01-17
make nightly CI workflow Laravel 10 compatible

szepeviktor:
@canvural After the release of Laravel 10 we will revert this.


szepeviktor:
Current nightly problems: https://github.com/nunomaduro/larastan/actions/runs/4001564048/jobs/6867919998

canvural:
I don't think we need to run this job with Laravel 10 also. Point of the job is to see if the next version of PHPStan gonna break Larastan or not.

szepeviktor:
I see.
Removing L10 from this workflow....

PR #1526, opened 2023-01-22
Removed forced type from callIsQuery

szepeviktor:
from nightly dev workflow


szepeviktor:
3 failed test with stable PHPStan

PR #1527, opened 2023-01-23
feat: add `Builder::chunk` method stub and test

amooij:
- [x] Added or updated tests

Adds type hints for the chunk method on the Eloquent Builder.

More code checked

### DIFF
--- a/stubs/EloquentBuilder.stub
+++ b/stubs/EloquentBuilder.stub
@@ -513,6 +513,15 @@ class Builder
      * @return static
      */
     public function whereRelation($relation, $column, $operator = null, $value = null);
+
+    /**
+     * Gets the result of a query in chunks.
+     *
+     * @param  int  $count
+     * @param  callable(\Illuminate\Database\Eloquent\Collection<int, TModelClass>, int): bool $callback
+     * @return bool
+     */
+    public function chunk($count, $callback);
 }
 
 class Scope {}
--- a/stubs/EloquentBuilder.stub
+++ b/stubs/EloquentBuilder.stub
@@ -518,7 +518,7 @@ class Builder
      * Gets the result of a query in chunks.
      *
      * @param  int  $count
-     * @param  callable(\Illuminate\Database\Eloquent\Collection<int, TModelClass>, int): bool $callback
+     * @param  callable(\Illuminate\Database\Eloquent\Collection<int, TModelClass>, int): mixed $callback
      * @return bool
      */
     public function chunk($count, $callback);
--- a/tests/Type/data/eloquent-builder.php
+++ b/tests/Type/data/eloquent-builder.php
@@ -432,6 +432,11 @@ function testPaginateItems()
     assertType('array<App\User>', User::query()->paginate()->items());
 }
 
+function testChunkOnEloquentBuilder()
+{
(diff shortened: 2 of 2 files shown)


szepeviktor:
Thank you @amooij !!

stubs/EloquentBuilder.stub
```diff
@@ -513,6 +513,15 @@ class Builder
      * @return static
      */
     public function whereRelation($relation, $column, $operator = null, $value = null);
+
+    /**
+     * Gets the result of a query in chunks.
+     *
+     * @param  int  $count
+     * @param  callable(\Illuminate\Database\Eloquent\Collection<TModelClass>, int): bool $callback
```
canvural, line 521: `\Illuminate\Database\Eloquent\Collection` needs 2 generic types. I think the key here can be `array-key`

tests/Type/data/eloquent-builder.php
```diff
@@ -432,6 +432,11 @@ function testPaginateItems()
     assertType('array<App\User>', User::query()->paginate()->items());
 }
 
+function testChunkOnEloquentBuilder()
+{
+    User::chunk(1000, fn ($collection) => assertType('Illuminate\Database\Eloquent\Collection<App\User>'));
```
canvural, line 437: After the above change, this needs to be updated also for the 2 generic types.

canvural (changes requested):

stubs/EloquentBuilder.stub
amooij: Done!

tests/Type/data/eloquent-builder.php
amooij: And done!

tests/Type/data/eloquent-builder.php
```diff
@@ -432,6 +432,11 @@ function testPaginateItems()
     assertType('array<App\User>', User::query()->paginate()->items());
 }
 
+function testChunkOnEloquentBuilder()
+{
+    User::chunk(1000, fn ($collection) => assertType('Illuminate\Database\Eloquent\Collection<int, App\User>'));
```
szepeviktor, line 437: I'm not a developer but after reading the docblock
```suggestion
    assertType('Illuminate\Database\Eloquent\Collection<int, App\User>', User::chunk(1000, fn ($collection, $someInteger) => $someBoolean));
```

szepeviktor, line 437: Is this the problem?
```suggestion
    User::chunk(1000, fn ($collection) => assertType('Illuminate\Database\Eloquent\Collection<int, App\User>', $collection));
```

szepeviktor:
Thank you.
Now it is okay.
Do not care about that merge commit and e2e tests.

szepeviktor (approved):

canvural:
I changed the return type of the callback to `mixed` Callback doesn't need to return anything, it's just if it returns false, the internal loop will stop. So `mixed` makes more sense.

Otherwise, it's good. Thank you!

amooij:
Thanks, can you explain something to me? All the unit tests fail on my machine on master. Is this a temporary issue or something in my setup? I'm running php 8.1 on MacOS with a regular composer.

canvural:
Can you show the errors?

PR #1530, opened 2023-01-24
Fake PR to test CI ✔

szepeviktor:
@canvural Some of our dependencies got _too updated_.


canvural:
I fixed the issue. But now there is something new: https://github.com/nunomaduro/larastan/actions/runs/3998803969/jobs/6861905800

@szepeviktor Can you take a look at that? I didn't understand the issue.

PR #1532, opened 2023-01-24
fix: installation in laravel

szepeviktor:
From https://github.com/nunomaduro/larastan/pull/1530#issuecomment-1402325374


szepeviktor:
@canvural After Laravel 10 release we will remove that `if`.

canvural:
Sounds good :+1

Thank you!

PR #1533, opened 2023-01-24
feat (chore): run phpstan with 1 gb memory

szepeviktor:
It is possible because the recent improvement of tests.

> Time: 00:25.091, Memory: 1.14 GB

In the future this reduced memory limit will allow us to notice when tests consume more memory.


canvural (approved):

canvural:
Makes sense. Thank you!

PR #1534, opened 2023-01-25
fix: update deps in e2e tests with less thrust

szepeviktor:
@canvural I'm not sure this will work on the long-term.

Let us be reckless cowboys! 🤠


canvural:
Both chandler and koel now seems to use Laravel 9.50 or higher. So I think we don't need this anymore.

szepeviktor:
Yes!

PR #1537, opened 2023-02-02
Support `@mixin` on custom Eloquent Model

rene-bos:
- [x] Added or updated tests

**Changes**
Add support for `@mixin CustomEloquentBuilder<static>` on a custom Model base class. Adding the mixin greatly improves autocompletion inside IDE's.

```php
/**
 * @mixin CustomEloquentBuilder<static>
 */
class CustomEloquentModel extends Model
{
    /**
     * @param Builder $query
     * @return CustomEloquentBuilder
     */
    public function newEloquentBuilder($query): CustomEloquentBuilder
    {
        return new CustomEloquentBuilder($query);
    }
}
```
_Why?_
We have a big application (500+ models) and use a custom Eloquent Model, Query Builder and Eloquent Builder. Larastan handles this fantastic (for which we thank you very much!!). To make PhpStorm regonize all pieces together we needed to add the mixin. Unfortunally this broke Larastan, so I decided to propose an upstream fix so others can take advantage of it too. I think it is a pattern which is usable for others too.

**Breaking changes**
None.

**Note**
Some checks fail, they also fail on the master branch and are (if I'm right) therefore out of scope for this PR.

### DIFF
--- a/src/ReturnTypes/ModelDynamicStaticMethodReturnTypeExtension.php
+++ b/src/ReturnTypes/ModelDynamicStaticMethodReturnTypeExtension.php
@@ -64,11 +64,11 @@ final class ModelDynamicStaticMethodReturnTypeExtension implements DynamicStatic
             return true;
         }
 
-        if (! $methodReflection->getDeclaringClass()->hasNativeMethod($name)) {
+        if (! $methodReflection->getDeclaringClass()->hasMethod($name)) {
             return false;
         }
 
-        $method = $methodReflection->getDeclaringClass()->getNativeMethod($methodReflection->getName());
+        $method = $methodReflection->getDeclaringClass()->getMethod($methodReflection->getName(), new OutOfClassScope());
 
         $returnType = ParametersAcceptorSelector::selectSingle($method->getVariants())->getReturnType();
 
--- a/src/ReturnTypes/ModelDynamicStaticMethodReturnTypeExtension.php
+++ b/src/ReturnTypes/ModelDynamicStaticMethodReturnTypeExtension.php
@@ -11,9 +11,11 @@ use Illuminate\Database\Query\Builder as QueryBuilder;
 use Illuminate\Support\Str;
 use NunoMaduro\Larastan\Methods\BuilderHelper;
 use PhpParser\Node\Expr\StaticCall;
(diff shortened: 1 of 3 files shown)


rene-bos:
Hi @szepeviktor! Not that I'm aware of, we tried quite a few of them. We only didn't try the paid one, as the per user costs are a bit high. This mixin gives us very good results on most of our Eloquent related code. With the right generics we have types for a big part of our code 😄.

szepeviktor:
> We only didn't try the paid one

That was I trick question. Paid ones come with a massage chair.

rene-bos:
Then I definitely have to give the paid one a try ☺️ 

How does reviewing work here? Do I just have to wait for someone to check the PR or do I have to assign someone?

rene-bos:
I just rebased on the latest changes of the master branch. E2E test failures are the known issues from the `chunk()` method. This PR is thus ready for review @canvural 🥳.

canvural:
I'm really surprised this didn't break anything else. Because by putting `mixin` to the model, PHPStan will try to find methods in the custom and Eloquent builder. Not through our extensions.

I'll need some time to understand what this change exactly does.

rene-bos:
I actually had very good results with `@mixin` and these changes in Larastan on a codebase of 7000+ PHP files. But what I didn't see was that it slowed down the scanning A LOT, it is around 3 times slower. When removing the `@mixin`s the scan is back fast again. This may also be due to changes in new PHPStan or Larastan versions, I haven't tested that.

I still like my idea of this pull request, as adding the `@mixin` is very developer friendly because IDE's will recognize a lot more functionality. I have some issues figuring out where the performance decrease came from, do you have tips for troubleshooting problems like this?

When I look at my code again, by replacing `hasNativeMethod` with the reflection method, I think I work around the problem you describe. I think that instead of Larastan which dynamically adds the "native methods", now the mixin is used which is read using reflection. If that's true, this change is just bad 😉.

calebdw:
@rene-bos, I think `@mixin` is the wrong approach---you can take a look at how I've set up my generics: https://github.com/larastan/larastan/issues/1777

Essentially I add update the return type on the `query` method, and then I begin all queries with `Model::query` which properly enables auto-completion on the custom builder

nunomaduro:

PR #1539, opened 2023-02-04
fix: schema dump could have .sql extension

joostdebruijn:
- [x] Added or updated tests

**Changes**
Since laravel/framework#45805 has merged (v9.50.0), a dumped database schema has a `.sql` extension instead of `.dump`. This PR searches for both type of file extensions.

**Breaking changes**
There are no breaking changes.

### DIFF
--- a/src/Properties/SquashedMigrationHelper.php
+++ b/src/Properties/SquashedMigrationHelper.php
@@ -107,7 +107,7 @@ final class SquashedMigrationHelper
                 $schemaFiles += iterator_to_array(
                     new RegexIterator(
                         new RecursiveIteratorIterator(new RecursiveDirectoryIterator($absolutePath)),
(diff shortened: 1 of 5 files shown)


szepeviktor:
Thank you @joostdebruijn!

joostdebruijn:
No problem, @szepeviktor. I see the E2E-tests failing, but I believe this is not related to this change - as those checks are failing on the master branch as well.

szepeviktor (approved):

canvural:
Looks good, thanks! But maybe we can add one more test that shows both `.dump` and `.sql` files could be used at the same time.

joostdebruijn:
@canvural I've added a test for that.

canvural:
Thank you!

PR #1543, opened 2023-02-05
fix: carry over collection class name for HigherOrderCollectionProxy

canvural:
- [x] Added or updated tests

This PR fixes an issue where method calls and property accesses on HigherOrderCollectionProxy resulting in wrong collection class.


szepeviktor:
@canvural Are these problems related to this PR?
https://github.com/nunomaduro/larastan/actions/runs/4195712300/jobs/7275647641

crynobone:
Hi, 

This PR just caused PHPStan to fail on our existing project:

CleanShot 2023-02-17 at 08 46 22

To fix the issue we had to make the following changes:

image

canvural:
@crynobone Should be fixed with new 2.5.0 release

PR #1551, opened 2023-02-16
feat: add stub for transform helper method

erikgaal:
- [x] Added or updated tests

**Changes**

Adds a stub method for the `transform` helper method. Previously it would always return `mixed`. Now it has the correct return type based on the passed value and callbacks.

**Breaking changes**

### DIFF
--- a/stubs/Helpers.stub
+++ b/stubs/Helpers.stub
@@ -155,3 +155,17 @@ function __($key = null, $replace = [], $locale = null) {}
  * @return ($callback is null ? mixed : ($value is null ? null : TReturn))
  */
 function optional($value = null, callable $callback = null) {}
+
+/**
+ * Transform the given value if it is present.
+ *
+ * @template TValue of mixed
+ * @template TReturn of mixed
+ * @template TDefault of mixed
+ *
+ * @param TValue $value
+ * @param callable(TValue): TReturn $callback
+ * @param null|TDefault|callable(TValue): TDefault $default
+ * @return ($value is empty ? ($default is null ? null : TDefault) : TReturn)
+ */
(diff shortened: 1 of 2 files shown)


szepeviktor:
We have some problems on master branch.
https://github.com/nunomaduro/larastan/actions/runs/4216179452/jobs/7317836255
@canvural Could you help sort it out?

canvural:
Yes, I will check and fix them.

szepeviktor (approved):

canvural:
Thank you!

stubs/Helpers.stub
```diff
@@ -155,3 +155,17 @@ function __($key = null, $replace = [], $locale = null) {}
...
+ * Transform the given value if it is present.
+ *
+ * @template TValue of mixed
+ * @template TReturn of mixed
+ * @template TDefault of mixed
+ *
+ * @param TValue $value
+ * @param callable(TValue): TReturn $callback
+ * @param null|TDefault|callable(TValue): TDefault $default
+ * @return ($value is empty ? ($default is null ? null : TDefault) : TReturn)
```
lptn, line 169: Isn't `$default is null ? null : TDefault` the same as just `TDefault`?

lptn: Also, transform has another logic, different from empty: it uses `blank()` under the hood. Unlike `empty`, these values a not blank:
 - `false`
 - `"0"`
 - `0`
 - `0.0`

Some values are blank, but not empty:
 - `Countable` with `count() === 0` are empty also (e.g. empty Collection)
 - strings contains whitespaces only

lptn (commented):
It seems like there is a room for improvements here

PR #1559, opened 2023-02-24
chore :clean up development setup

spawnia:
Just a few things that irritated me while working on a pull request.


.github/workflows/tests.yml
```diff
@@ -73,5 +73,5 @@ jobs:
       - name: "Execute unit tests"
         run: "composer run-script test:unit"
 
-      - name: "Run Larastan on Laravel sample applications"
+      - name: "Run Larastan on a sample Laravel application"
```
szepeviktor, line 76: 👍🏻

szepeviktor:
I wish you win the battle with unit tests.

tests/IntegrationTest.php
```diff
@@ -12,9 +12,9 @@ class IntegrationTest extends PHPStanTestCase
      */
     public function dataIntegrationTests(): iterable
     {
-        yield [__DIR__.'/integration/data/test-case-extension.php'];
-        yield [__DIR__.'/integration/data/model-properties.php'];
-        yield [__DIR__.'/integration/data/blade-view.php'];
+        yield [__DIR__.'/data/test-case-extension.php'];
+        yield [__DIR__.'/data/model-properties.php'];
+        yield [__DIR__.'/data/blade-view.php'];
```
canvural, line 17: Why this change? `integration` folder is good because then it's clear these are data for integration tests.

canvural (changes requested):
Looks good! More consistency is always better! 😄 

I have only one question for now. Other than that it looks good 👍🏽

tests/IntegrationTest.php
spawnia: It follows the pattern established in other layers of the test folder hierarchy. There, it is not clear to which test class the data belongs to either:

image

spawnia:
> I wish you win the battle 🥊 against unit tests.

Thanks, just did. The remaining static analysis issues seem related to the PHPStan 1.10 update. I would appreciate if you could merge this regardless.

tests/IntegrationTest.php
canvural: So maybe it's better to also put the `IntegrationTest` into some `Integration` folder. And in there have `data` folder.

spawnia: Just did that, thanks. I also renamed `tests/Application` to `tests/application` to signify it is not just another namespace for tests to live in.

canvural:
Thank you!

PR #1560, opened 2023-02-25
Update E2E tests repositories

rene-bos:
- [X] Added or updated tests

The E2E tests are failing for some time now, I found this very annoying. That's why I fixed [the errors at koel/koel](https://github.com/koel/koel/pull/1659) side and update the refs for all this and all other repositories. This fixes the E2E tests which are run for every PR.

**Changes**

Update E2E tests repositories to latest commit.

**Breaking changes**

No.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -37,7 +37,7 @@ jobs:
             checkout-with: |
               {
                 "repository": "koel/koel",
-                "ref": "545a303e0a3eca6a10909cdda73dd04264683727",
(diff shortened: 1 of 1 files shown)


szepeviktor:
Thank you René 🐝

canvural:
Thank you!

I want to implement baseline for e2e tests like PHPStan [has](https://github.com/phpstan/phpstan/blob/1.10.x/e2e/integration/larastan-baseline.neon), in the future. So we wouldn't need to wait for fixes from repos.

rene-bos:
Yeah, that would be great! 👌🏻

PR #1561, opened 2023-02-25
Support PHPStan 1.10

rene-bos:
- [x] Added or updated tests

After the PHPStan 1.10 release, Larastans builds fail because of some newly introduced rules. See [this job](https://github.com/nunomaduro/larastan/actions/runs/4269278910/jobs/7432359604) for example. 

**Changes**

I added the new errors to the baseline so all our builds will be green again. We then can fix the errors when we got time for it.

**Breaking changes**

None.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,12 +1,227 @@
 parameters:
 	ignoreErrors:
 		-
+			message: "#^PHPDoc tag @var with type string is not subtype of native type non\\-empty\\-string\\|false\\.$#"
+			count: 1
+			path: src/ApplicationResolver.php
+
+		-
+			message: "#^Doing instanceof PHPStan\\\\Type\\\\Generic\\\\GenericObjectType is error\\-prone and deprecated\\.$#"
+			count: 1
+			path: src/Methods/BuilderHelper.php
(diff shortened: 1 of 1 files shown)


szepeviktor:
Thank you!
I think we should fix those errors.

rene-bos:
Yes we should for sure! But for now it would be good to have the pipelines running without errors. We can then fix the errors without any pressure 😄 Shall I create an issue for it?

szepeviktor:
This is really bad: #1562
I'm ashamed of this.

szepeviktor:
Others may have another opinion.
Errors should not be ignored but resolved.

rene-bos:
I totally agree with you, I see adding errors to the baseline just as a temporary workaround. In fact the baseline is one big pile of technical debt, which should be paid of as soon as possible 😄

szepeviktor:
This is my first step: #1563

canvural:
Hi,

Thank you for this. I understand as a contributor seeing a failing CI is not good. But in the end, this issues don't affect the end user. I started to work on this. So, if it is fine with you, I'm gonna close this one. And will try to finish my fix until Wednesday.

PR #1562, opened 2023-02-25
fix: disable instanceofType toggle for now

szepeviktor:
The remaining 4 errors are from PHPStan 1.9.x


szepeviktor:
> Unexpected item 'parameters › featureToggles › instanceofType'.

It makes older PHPStan-s fail.

rene-bos:
Why would you like to disable the instanceofType check? When it is disabled we won't see mistakes we make in new code. Also it makes it more difficult to find the errors and fix them.

I think adding the errors to the baseline like I did in #1561 is better because of that.

szepeviktor:
Others may have another opinion.
Errors should not be ignored but resolved.

PR #1563, opened 2023-02-25
fix: fix type in ApplicationResolver

szepeviktor:
Just that.


szepeviktor:
@canvural So we have some tech debt from PHPStan 1.9 (Node errors) and from 1.10 (type detection).

canvural:
Thank you!

PR #1573, opened 2023-03-05
feat: create options to disable migration and schema scanning

CalebDW:
- [x] Added or updated tests

**Changes**
I added options to disable migrations and schema scanning. As mentioned in #1569, Postgres schemas are scanned (which may take upwards of 7 minutes) even when they produce invalid results. This will enable the end user to simply disable scanning if desired.

**Breaking changes**
N/A

### DIFF
--- a/docs/custom-config-parameters.md
+++ b/docs/custom-config-parameters.md
@@ -20,6 +20,17 @@ parameters:
         - app/Domain/DomainB/migrations
 ```
 
+## `disableMigrationScan`
+**default**: `false`
+
+You can disable use this config to disable migration scanning.
+
+#### Example
+```neon
+parameters:
+    disableMigrationScan: true
+```
+
 ## `squashedMigrationsPath`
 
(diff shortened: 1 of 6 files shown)


szepeviktor:
Thank you for this @CalebDW!

I think these switches affect very small number of users ... I could count them on one hand.

CalebDW:
> I think these switches affect a very small number of users ... I could count them on one hand.

That may be so, but I don't think it hurts to have them (and they're `false` by default so everyone else wouldn't have to worry about setting them).

canvural:
I understand the need to disable squashed migration parsing. But why the need to disable migration scanning?

CalebDW:
:man_shrugging: it was just kind of the sister option so I figured it wouldn't hurt to add it. Maybe some folks use PHPdocs on their models and wouldn't want the migrations scanned.

CalebDW:
Is there anything blocking this merge?

canvural:
Thank you!

PR #1576, opened 2023-03-05
feat: console command analysis

canvural:
- [x] Added or updated tests

**Changes**
This PR adds console command analysis. 
- It infers more specific types for `argument`, `arguments`, `option` and `options` methods.
- Adds 2 new rules to find undefined argument or option usages.

**Breaking changes**

N/A


tests/Type/GeneralTypeTest.php
```diff
@@ -52,6 +52,13 @@ public function dataFileAsserts(): iterable
         yield from $this->gatherAssertTypes(__DIR__.'/data/gate-facade.php');
         yield from $this->gatherAssertTypes(__DIR__.'/data/helpers.php');
         yield from $this->gatherAssertTypes(__DIR__.'/data/custom-eloquent-collection.php');
+
+        //##############################################################################################################
```
szepeviktor, line 56: This is poetry 🙃

PR #1583, opened 2023-03-11
feat: add support for relationship count property

### LINKED ISSUE #466: Relationship count should be recognized as property

canvural:
- [x] Added or updated tests

Fixes #466 

**Changes**

This PR adds support for relationship count property access on models. For example `$user->accounts_count`

**Breaking changes**
N/A

### DIFF
--- a/src/Properties/ModelRelationsExtension.php
+++ b/src/Properties/ModelRelationsExtension.php
@@ -17,6 +17,7 @@ use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Reflection\PropertiesClassReflectionExtension;
 use PHPStan\Reflection\PropertyReflection;
 use PHPStan\Type\Generic\GenericObjectType;
+use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\IntersectionType;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\NeverType;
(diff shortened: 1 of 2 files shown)


mbardelmeijer:
@nunomaduro would you mind tagging a new release of Larastan to include this feature :)

szepeviktor:
March was a long time ago ⏳

szepeviktor:
Here you go @mbardelmeijer 🚀 
https://github.com/nunomaduro/larastan/releases/tag/2.5.2

canvural:
@szepeviktor The new release wasn't suppose to be 2.5.2 but 2.6

I know I didn't dedicate too much time into oss in the last couple of months, but please don't make new releases before discussing it with me in the future.

canvural:
Also this feature has a bug in it. I was planning on fixing that before making a release. So, I'll make a new one with that fix later.

szepeviktor:
@canvural All right.
At least 2.6.0 can be still released.

> Also this feature has a bug in it.

Please consider making this knowledge public in a PR or in an issue.
Thank you.

PR #1585, opened 2023-03-12
fix: `transform` callable argument is never null

erikgaal:
- [x] Added or updated tests
- [x] ~Documented user facing changes~

**Changes**

Fixes an issue where the `$callback` parameter in `transform` is still expected to have a nullable argument. It surfaced during testing in our project where it gave the following error:

```
 Parameter #2 $callback of function transform expects callable(\App\User): array, Closure(\App\User): array given.
```

Also simplifies the conditional return type as suggested by @lptn in #1551.

**Breaking changes**

None.


szepeviktor:
Thank you.

PR #1590, opened 2023-03-16
Fix stub for Builder::havingBetween()

lptn:
havingBetween accepts `iterable`, not arrays only:

https://github.com/laravel/framework/blob/136ed754f2d80e5c7cd407a4262eb2eac478e7e8/src/Illuminate/Database/Query/Builder.php#L2221-L2243

### DIFF
--- a/stubs/QueryBuilder.stub
+++ b/stubs/QueryBuilder.stub
@@ -149,12 +149,12 @@ class Builder
      * Add a "having between " clause to the query.
      *
      * @param  string  $column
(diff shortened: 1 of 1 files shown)


canvural:
Hi,

This looks like it got changed in 10.x. In 9.x it is still `array` Since we still support both we'd need to add different stub files for different Laravel versions.

But in this specific case the docblocks looks exactly the same between our stubs and Laravel. I honestly forgot why I added all those method to stubs 😅  I'll try to refactor the stubs instead.

PR #1594, opened 2023-03-20
Add translation return type extensions.

mad-briller:
- [x] Added or updated tests

**Changes**
Adds a few return type extensions for translation related features to provide more contextual typing, meaning less false positives


mad-briller:
monicahq/chandler failure seems to be related to the improved console command analysis released recently, not a result of this change

szepeviktor:
I like your TranslationHelper usage 💝 much netter than we write here.

mad-briller:
I think that's fine, it will solve the false positives which is my major concern with this change 

I appreciate the desire to avoid dynamic analysis as it adds complexity

I'll try to revise this soon

mad-briller:
can a benevolent union be specified in a phpdoc? i don't think it can?

canvural (approved):

canvural:
> can a benevolent union be specified in a phpdoc? i don't think it can?

It's kinda possible, but it's highly internal stuff used in PHPstan. So extension is the right way to go 👍🏽 

Thank you!

PR #1596, opened 2023-03-21
feat: add generics to more methods in Relation stub

rene-bos:
- [x] Added or updated tests

**Changes**

Add generics to more public methods in the Relation stub file.

**Breaking changes**

None.


szepeviktor (approved):

tests/Type/data/relation.php
```diff
@@ -0,0 +1,15 @@
+<?php
```
canvural, line 1: We already have `model-relations.php` so we can just move these to that file.

canvural (changes requested):
Overall looks ok. But I have a question about `getQuery` Would it return the custom query builder, if the model uses one? If so, we might need a new extension that handles that.

tests/Type/data/relation.php
rene-bos: Fixed.

rene-bos:
> Overall looks ok. But I have a question about `getQuery` Would it return the custom query builder, if the model uses one? If so, we might need a new extension that handles that.

@canvural Yeah I think it will. The relation will create a new instance of the related class, and will call `newQuery()` on that. I think this change is already an nice improvement for most setups, are you fine with merging this first and extending it with an extension for custom query builders when somebody has time or the need for it?

canvural (approved):
Looks good 👍🏽 But we'd need to iterate on it to support custom builders.

canvural:
Thank you!

PR #1599, opened 2023-03-27
Fix dropColumn() migrations when they contain arrays

MaxGiting:
- [x] Added or updated tests
~~Documented user facing changes~~

**Issue**
Larastan parses migrations but only handles`dropColumn()` if a string is passed in.
```php
$table->dropColumn('column') //This line works
$table->dropColumn(['address1','address2']); //This line is ignored
```

**Changes**
This PR has fixed that issue and added a test to check.

### DIFF
--- a/src/Properties/SchemaAggregator.php
+++ b/src/Properties/SchemaAggregator.php
@@ -207,6 +207,14 @@ final class SchemaAggregator
                     }
 
                     if (! $firstArg instanceof PhpParser\Node\Scalar\String_) {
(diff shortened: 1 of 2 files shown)


canvural (changes requested):
Looks ok. But there is no test here actually. Please look at [here](https://github.com/nunomaduro/larastan/blob/11bdd132565d37e808e4ac985a580b492c2d8887/tests/Unit/MigrationHelperTest.php), and try to add a new test case that verifies both columns are dropped.

canvural:
Thank you. I'll add the tests myself later.

PR #1604, opened 2023-03-30
Filter out null from callback in `when` method

rovansteen:
- [x] Added or updated tests

**Changes**

When passing a value that has a type of `string|null` to Laravel's `when` method I'd expect to be able to narrow that type to `string` for the second parameter.

**Breaking changes**
—

### DIFF
--- a/tests/Type/data/conditionable.php
+++ b/tests/Type/data/conditionable.php
@@ -25,6 +25,11 @@ assertType('int', (new Foo())->when(true, function (Foo $foo): int {
     return rand();
 }));
 
+// Test to make sure the callback has a non-null value.
+(new Foo())->when(User::first(), function (Foo $foo, $user): void {
+    assertType(User::class, $user);
+});
+
 assertType('ConditionableStubs\Foo', (new Foo())->unless(true, function (Foo $foo) {
     // do nothing
 }));
--- a/stubs/Conditionable.stub
+++ b/stubs/Conditionable.stub
@@ -12,7 +12,7 @@ trait Conditionable
      * @template TWhenParameter
      * @template TWhenReturnType
      *
-     * @param  (\Closure($this): TWhenParameter)|TWhenParameter  $value
+     * @param  (\Closure($this): TWhenParameter)|TWhenParameter|null  $value
      * @param  (callable($this, TWhenParameter): TWhenReturnType)|null  $callback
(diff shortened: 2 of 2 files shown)


tests/Type/data/conditionable.php
```diff
@@ -24,6 +24,12 @@ class Foo
     return rand();
 }));
 
+/** @var string|null */
+$var = 'string'
+(new Foo())->when($var, function (Foo $foo, string $var) {
```
canvural, line 29: I think this test would always pass because of the `string` typehint here. Can you remove it and see if still passes?

rovansteen: Done!

canvural: Yeah, as you can see tests are failing now. So this is not the correct way to handle this I guess.

erikgaal: Yeah, this test doesn't work because PHPStan will ignore the `@var` annotation and still think it's a `string`. In the PR I did before to support `transform` I used `User::first()` which is definitely `App\User|null`.

erikgaal: Use this test instead and it should work:

```php
// Test to make sure the callback has a non-null value.
(new Foo())->when(User::first(), function (Foo $foo, $user): void {
    assertType(User::class, $user);
});
```

tests/Type/data/conditionable.php
```diff
@@ -24,6 +24,12 @@ class Foo
     return rand();
 }));
 
+/** @var string|null */
+$var = 'string'
+(new Foo())->when($var, function (Foo $foo, $var) {
+    assertType('string', $var);
+});
+
```
erikgaal, line 32:
```suggestion
// Test to make sure the callback has a non-null value.
(new Foo())->when(User::first(), function (Foo $foo, $user): void {
    assertType(User::class, $user);
});

```

tests/Type/data/conditionable.php
rovansteen: Thanks @erikgaal!

rovansteen: @canvural could you have another look at this? Need your approval to run the tests

canvural:
Even though I don't understand how exactly this works, I'm gonna merge it 😄 We can revert it if it causes issues.

Thank you!

PR #1612, opened 2023-04-03
fix: correct return type of `Illuminate\Database\Query\Builder::pluck()` in stubs/QueryBuilder.stub

Quarkay:
- [x] Added or updated tests

### DIFF
--- a/tests/Type/data/query-builder.php
+++ b/tests/Type/data/query-builder.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EloquentBuilder;
+
+use Illuminate\Support\Facades\DB;
+
+use function PHPStan\Testing\assertType;
+
+function testPluckReturnType()
+{
+    $record = DB::table('user')->pluck('email', 'id');
+    assertType('Illuminate\Support\Collection<(int|string), mixed>', $record);
+}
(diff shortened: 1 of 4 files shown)


Quarkay:
Hello guys I'm new here.

I found a issue of the stub file while using larastan in my repo ( or maybe, I made a wrong usage ) . 

In the test case the specific return type of `pluck()` method should be `Illuminate\Support\Collection<int,string>` which is a counterexample of the current version doc type `Illuminate\Support\Collection<string,mixed>`

Thanks for you guys wonderful work! Larastan really helps a lot.

tests/Type/data/query-builder.php
```diff
@@ -0,0 +1,17 @@
...
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\DB;
+
+use function PHPStan\Testing\assertType;
+
+function testPluckReturnType()
+{
+    /** @var Collection<int, string> $record */
+    $record = DB::table('user')->pluck('email', 'id');
+    assertType('Illuminate\Support\Collection<int, string>', $record);
```
canvural, line 16:
```suggestion
    assertType('Illuminate\Support\Collection<array-key, string>', $record);
```

canvural (changes requested):

canvural:
Hi,

It looks fine overall, but just need to update the test. Also rebase the branch on top of the latest master 👍🏽

tests/Type/data/query-builder.php
Quarkay: @mad-briller  Yes exactly! I modified the code with reference to existed test cases.

By the way I found phpstan will not warn inproprite type overwritting in test case environment but it will do this in my personal repo (with same restrict level).

canvural:
Thank you!

PR #1618, opened 2023-04-11
fix: Http/Request.stub types are incorrect

tpetry:
The documented type is incorrect. When not passing a key the result will be a map of headers:

```php
[
  "accept": ["text/html, application/xhtml+xml"],
  "user-agent": ["Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/110.0.0.0 Safari/537.36"],
  // ...
]
```

### DIFF
--- a/stubs/Http/Request.stub
+++ b/stubs/Http/Request.stub
@@ -12,7 +12,7 @@ class Request {
      * @param  string|null  $key
      * @param  TDefault $default
      *
-     * @return ($key is null ? array<string, string> : string|TDefault)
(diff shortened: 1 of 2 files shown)


szepeviktor:
Thank you for the fix.

Could you add a test that fails without this PR?

szepeviktor:
We have that!

https://github.com/nunomaduro/larastan/blob/69f1ec7c0b9404a515c0bf8704a1da7bbbc1beda/tests/Type/data/request-header.php#L10

canvural:
Thank you!

Issue #1367, opened 2022-09-05
"Todo" hunt. Up for grabs!


szepeviktor:
https://github.com/nunomaduro/larastan/search?q=todo 👀 

There are only three todos.

maximal:
~~four~~ seven :-)

PR #1634, opened 2023-05-01
Rules review

joelclermont:
- [ ] Added or updated tests

I've used Larastan for a while and I love it! This is my first foray into Larastan code, and I started by reviewing the Rules.

I'm opening this as a Draft PR just to get some feedback and see if you're open to these types of changes.

I also wrote a failing test to address one of the todos from #1367 but I'm not quite clear how to get it to pass. Inside `isCalledOnModel`, the `$class` is an instance of `FullyQualified`, but the `isSuperTypeOf` check is `maybe()` not `yes()`. I'll keep digging, but I mention it in case you can point me in the right direction.

**Changes**

- Fixed an incorrect generic type on one rule
- I standardized imports of PHPStan classes (I didn't see any detailed style rules, so I kept this in a separate commit in case it's not welcome)
- Add a psr-4 mapping for the Tests namespace
- Added a failing test for the `todo` inside the `NoModelMakeRule`

**Breaking changes**

None

### DIFF
--- a/src/Rules/NoModelMakeRule.php
+++ b/src/Rules/NoModelMakeRule.php
@@ -7,7 +7,6 @@ namespace NunoMaduro\Larastan\Rules;
 use Illuminate\Database\Eloquent\Model;
 use PhpParser\Node;
 use PhpParser\Node\Expr;
-use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Identifier;
 use PhpParser\Node\Name\FullyQualified;
@@ -27,7 +26,7 @@ use PHPStan\Type\ObjectType;
  * It is functionally equivalent to simply use the constructor:
  * new User()
  *
- * @implements Rule<MethodCall>
+ * @implements Rule<StaticCall>
  */
 class NoModelMakeRule implements Rule
 {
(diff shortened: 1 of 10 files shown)


szepeviktor:
> I standardized imports of PHPStan classes

Thank you for this PR.

FQCN should be used in
- docblocks
- exceptions
- in `extends`/`implements`

szepeviktor:
> `"Tests\\": "tests/"`

It is a big work to make tests directory PSR-4 compliant.
First remove `classmap` mapping and issue
`composer dump-autoload --optimize --strict-psr`

src/Rules/NoModelMakeRule.php
```diff
@@ -27,12 +26,12 @@
  * It is functionally equivalent to simply use the constructor:
  * new User()
  *
- * @implements Rule<MethodCall>
+ * @implements Rule<StaticCall>
```
szepeviktor, line 29: Could you add a test that fails without this change?

szepeviktor: This is very interesting.

joelclermont: Thanks for the feedback on the other items. I'll revise and update. I'll think on how to make a failing test for this one too.

canvural:
Hi,

This PR does too many stuff inside. I'd appreciate if each PR is focused on only one thing. Also, for code style we use StyleCI, so if it doesn't complain we don't need to do manual changes. Hope you understand.

For the NoModelMakeRule change please send a separate PR, that looks interesting!

Thank you.

joelclermont:
Sure thing, I'll re-submit after Laracon. Thanks!

PR #1635, opened 2023-05-01
fix: use FQCN in docblocks

szepeviktor:
This is what PHPStan source code does.
Currently we have **mixed** relative names and FQCN.

Done automatically.
```bash
vendor/bin/phpcbf --standard=SlevomatCodingStandard --sniffs=SlevomatCodingStandard.Namespaces.FullyQualifiedClassNameInAnnotation src/ tests/
```

Larastan got 50 lines shorter! (removed imports only for docblocks)


canvural:
Hi,

Instead of manually changing style in different PRs, I'd prefer to stick to one tool.

At this point we can even migrate to PHP_CodeSniffer altogether ditching StyleCI. If you are interested you can send a PR for that. (If you do that I'd like to use https://github.com/doctrine/coding-standard also on top of it)

szepeviktor:
All right.
If you change your mind you could issue the above command in a minute.

PR #1660, opened 2023-06-14
fix: proper type resolving for `foreignIdFor` fields

### LINKED ISSUE #1653: Incorrect type for foreignIdFor when primary key is uuid
- Larastan Version: 2.6.0
- `--level` used: 5

### Description
Reference type for a model with `\Illuminate\Database\Eloquent\Concerns\HasUuids` trait resolves to `int` instead of `string`

Here is a minimal repo that reproduces this issue: https://github.com/sanfair/larastan-playground/commit/35d934f0b258fe5c27bbe3a14aa514f4d61f4aee

### The root cause of the problem
Type hardcoded in [src/Properties/SchemaAggregator.php#L204](https://github.com/nunomaduro/larastan/blob/ccac5b25949576807862cf32ba1fce1769c06c42/src/Properties/SchemaAggregator.php#L204)
```php
if ($firstMethodCall->name->name === 'foreignIdFor') {
    if ($firstArg instanceof PhpParser\Node\Expr\ClassConstFetch
        && $firstArg->class instanceof PhpParser\Node\Name
    ) {
        $modelClass = $firstArg->class->toCodeString();
    } elseif ($firstArg instanceof PhpParser\Node\Scalar\String_) {
        $modelClass = $firstArg->value;
    } else {
        continue;
    }

    $columnName = Str::snake(class_basename($modelClass)).'_id';
    if ($secondArg instanceof PhpParser\Node\Scalar\String_) {
        $columnName = $secondArg->value;
    }

    $table->setColumn(new SchemaColumn($columnName, 'int' /** <---- HERE */, $nullable));

    continue;
}
```

#### Lookup type in parsed tables
 Problems:
* If the table name is not default need to use reflection or class instance to find the correct table name
* Need to know the primary column name

Partial implementation:
```php
private function getIdTypeFromRelationModel(string $modelClass): ?string
{
    $tableName = Str::snake(Str::pluralStudly(class_basename($modelClass)));
   
    if (!array_key_exists($tableName, $this->tables)) {
         /* Resolve table using reflection. */
        return null;
    }
 
    $table = $this->tables[$tableName];

    $column = 'id';

    if (!array_key_exists($column, $table->columns)) {
         /* Resolve column using reflection. */
        return null;
    }

    return $table->columns[$column]->readableType;
}

sanfair:
- [x] Added or updated tests
- [ ] Documented user-facing changes

Resolves #1653

**Changes**

Type for `foriginIdFor` fields will be set using a primary key column from the table for the related model, with a fallback to the previous implementation (hardcoded `int` type) when a table or column is not found.

**Breaking changes**

Type for `foriginIdFor` fields might change.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -413,6 +413,7 @@ services:
             databaseMigrationPath: %databaseMigrationsPath%
             disableMigrationScan: %disableMigrationScan%
             parser: @currentPhpVersionSimpleDirectParser
+            reflectionProvider: @reflectionProvider
 
     -
         class: NunoMaduro\Larastan\Properties\SquashedMigrationHelper
--- a/src/Properties/MigrationHelper.php
+++ b/src/Properties/MigrationHelper.php
@@ -7,6 +7,7 @@ namespace NunoMaduro\Larastan\Properties;
 use PHPStan\File\FileHelper;
 use PHPStan\Parser\Parser;
 use PHPStan\Parser\ParserErrorsException;
+use PHPStan\Reflection\ReflectionProvider;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
 use RegexIterator;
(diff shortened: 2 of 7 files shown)


canvural:
Thank you!

Issue #1656, opened 2023-06-13
Generic return type from query builder


wslawski-printify:
- Larastan Version: 2.6.2
- `--level` used: 8
- Pull request with failing test:

### Description

When calling `query` on model instances and getting results from it there is generic return type returned instead of specific one. This happens when $model is assigned to property. When trying to do `SomeModel::query` it works fine

### Laravel code where the issue was found

```php
final class SomeModelRepository implements SomeModelRepositoryInterface
{
    private SomeModel $model;

    public function __construct()
    {
        $this->model = new SomeModel(); // extending laravel model
    }

    public function findById($id): ?SomeModel
    {
        return $this->model::query()
            ->where('id', $id)
            ->get()
            ->first();
    }

    public function findByIds(array $ids): array
    {
        return $this->model::query()
            ->whereIn('id', $ids)
            ->get()
            ->all();
    }
}
```

canvural:
Hi,

I did not understand the issue. Can you give what is the error and what is the expected output?

wslawski-printify:
Sure, the error is:

```
  21     Method                                                                                                   
        SomModelRepository::findById()   
         should return SomModel|null but returns           
         Illuminate\Database\Eloquent\Model|null.
```

Interface looks like this:

```
interface PrintProviderRoutingRepository
{
    public function findById($id): ?PrintProviderRouting;

    /**
     * @return array<int, PrintProviderRouting>
     */
    public function findByIds(array $ids): array;
}
```

When using `SomeModel::query` everything works fine.

canvural:
Fixed: eb31ffa

wslawski-printify:
Thanks a lot :)

canvural:
Released: https://github.com/nunomaduro/larastan/releases/tag/v2.6.3

wslawski-printify:
One more thing, there is also an issue if calling `->query()` vs `::query()` on such model being an variable. Should I create separate issue? @canvural

PR #1670, opened 2023-06-21
Test user->query

szepeviktor:
From https://github.com/nunomaduro/larastan/issues/1656#issuecomment-1600651965


canvural:
Thanks for showing the error. I'm closing this because it does not have the fix. I'll add it later myself 👍🏽

PR #1673, opened 2023-06-26
Import all global functions

mad-briller:
- [x] Added or updated tests

**Changes**
Adds `use function` for all global functions used in the extension, this reduces the overhead of calls and gives opcache more opportunities to optimize

... just waiting for styleci to send me a patch to format it correctly :sweat_smile:

### DIFF
--- a/src/ApplicationResolver.php
+++ b/src/ApplicationResolver.php
@@ -12,6 +12,11 @@ use Orchestra\Testbench\Foundation\Application as Testbench;
 use Orchestra\Testbench\Foundation\Bootstrap\CreateVendorSymlink;
 use Orchestra\Testbench\Foundation\Config;
 
+use function class_exists;
+use function defined;
+use function file_exists;
+use function getcwd;
+
 /**
  * @internal
  */
--- a/src/Collectors/UsedEmailViewCollector.php
+++ b/src/Collectors/UsedEmailViewCollector.php
@@ -12,6 +12,9 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Collectors\Collector;
 use PHPStan\Type\ObjectType;
 
+use function count;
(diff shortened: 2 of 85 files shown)


szepeviktor (approved):
Very good job!

szepeviktor:
Could you add a rule enforcing it to StyleCI configuration?
https://cs.symfony.com/doc/rules/import/global_namespace_import.html

mad-briller:
the project does not have a local setup for phpcs, and instead uses the `laravel` preset of styleci to enforce coding standards
i'm not at all familiar with styleci, thought i imagine it's possible

szepeviktor:
> and instead uses the `laravel` preset of styleci

I am looking it up for you 🏃🏻 
https://docs.styleci.io/configuration#default-configuration

szepeviktor:
`/.styleci.yml`

```yaml
preset: laravel
enabled:
  - global_namespace_import:
    - import_classes: true
    - import_constants: false
    - import_functions: true
```

Docs says this is not possible.
https://docs.styleci.io/fixers
@GrahamCampbell Could you help us?

sanfair:
Hello, I found that StyleCI has [`native_function_invocation`](https://docs.styleci.io/fixers#native_function_invocation)

So we can update the config like so:
```yml
preset: laravel
risky: true
enabled:
  - native_function_invocation
```

I tried it and it added a leading slash to internal functions in the global namespace (https://github.styleci.io/analyses/ne1YaN)
However, the calls to the user functions have not changed

And it won't work with suggested changes. This way the code will look a little messy in my opinion (https://github.styleci.io/analyses/ma5OVK)

canvural:
If StyleCI is fine with it for now, it's enough. We can find a solution for new code, later.

Thank you!

PR #1674, opened 2023-06-26
Refactor src/Types/GenericEloquentBuilderTypeNodeResolverExtension to be able to fail quicker.

mad-briller:
- [x] Added or updated tests

Profiling larastan on a simple script like this:

```php
<?php

var_dump(collect());
```

i noticed that the `GenericEloquentBuilderTypeNodeResolverExtension` was taking up 8% of the execution time, the majority being spent in the `isSuperTypeOf` checks
and after fiddling with it a bit, the cost is reduced to 4% of time with this small change

it seems counterintuitive but looping twice allows us to fail earlier, which is useful when the majority of times the extension is called with non-matching input, and we know we will only ever iterate twice in each foreach thanks to the if above


canvural:
I'm happy if the tests pass, so we can do 2 loops 😀

Thank you!

PR #1677, opened 2023-06-28
Play with fire 🔥

szepeviktor:
🔥🔥🔥


szepeviktor:
> The 'global_namespace_import' fixer does not exist.

PR #1678, opened 2023-06-28
Make UndefinedArgumentOrOptionRule fail sooner to avoid expensive type system calls

mad-briller:
- [x] Added or updated tests

Reorder the checks in `UndefinedArgumentOrOptionRule` so that cheap checks are first

this reduced the cost of this rule from 1% to 0% of total time on a decent sized class that is not an instance of `Command`.


canvural:
Thank you!

PR #1679, opened 2023-06-28
Check for method name earlier in NoUnnecessaryCollectionCallRule to avoid expensive calls to type system.

mad-briller:
- [x] Added or updated tests

Reorder checks in `NoUnnecessaryCollectionCallRule` so that cheap checks are first

this reduced the cost of this rule on a decent sized class that doesn't use `Collection` at all from 6% to 0% of total time


canvural:
Thank you!

PR #1682, opened 2023-07-03
Uses @template-covariant in collections

Grldk:
- [ ] Added or updated tests

**Changes**

Changes Collection TValue from `@template` to `@template-covariant`

see, https://github.com/laravel/framework/pull/46872

I am not sure if this change needs added tests. All existing tests pass.

### DIFF
--- a/stubs/Collection.stub
+++ b/stubs/Collection.stub
@@ -8,7 +8,8 @@ use Illuminate\Support\Traits\EnumeratesValues;
 
 /**
  * @template TKey of array-key
- * @template TValue
+ * 
+ * @template-covariant TValue
  *
  * @implements \ArrayAccess<TKey, TValue>
  * @implements \Illuminate\Support\Enumerable<TKey, TValue>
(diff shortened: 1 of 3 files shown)


Grldk:
Adding `@template-covariant` in `EloquentBuilder.stub` would solve a number of problems for me as well. Should I create a separate PR for that, or can I add that to this one?

szepeviktor:
It would be nice to fix unit tests here.

Grldk:
Hmm, all tests passed for me locally, so to be honest I didn't really pay any attention to the test results here.. 

I'll try to fix the unit tests soon, hopefully tomorrow or else early next week.

Grldk:
Well, I guess those tests will not pass: https://github.com/phpstan/phpstan/discussions/9590

Not sure what to do with this now. We'll have to wait for call-site variance in phpstan I guess..

mad-briller:
https://github.com/phpstan/phpstan-src/pull/2485#pullrequestreview-1518607915 i believe it's already in the works thanks to this pr

PR #1685, opened 2023-07-10
Fix type of Attribute::make's callable

ludoguenet:
I encountered the following error:
```
Type array of parameter #2 $attributes of passed callable needs to be same or wider than parameter type mixed of accepting callable.
```

However, based on Laravel's documentation and [code](https://github.com/laravel/framework/blob/921e956576aba9655da1f6667fbdbe97394d3cf3/src/Illuminate/Database/Eloquent/Casts/Attribute.php#L48-L55), I believe that this second parameter should be typed as an array<string, mixed> and not fully mixed.

- [ ] Added or updated tests

**Changes**

**Breaking changes**

### DIFF
--- a/stubs/Attribute.stub
+++ b/stubs/Attribute.stub
@@ -27,7 +27,7 @@ class Attribute
      *
      * @template TMakeGet
      * @template TMakeSet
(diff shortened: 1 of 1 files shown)


nunomaduro:

PR #1688, opened 2023-07-17
Implement baseline for e2e tests.

sanfair:
>... implement baseline for e2e tests like PHPStan [has](https://github.com/phpstan/phpstan/blob/1.10.x/e2e/integration/larastan-baseline.neon) ...
>
>https://github.com/nunomaduro/larastan/pull/1560#issuecomment-1445465467

**Changes**

Added phpstan config override and a separate baseline file for each repo in e2e tests.

**Breaking changes**

No changes

**Notes**
I did not implement a fully automated workflow for creating new PR with an updated baseline. Feel free to add this part if needed.
I think you can almost copy-paste it from the [PHPStan repo](https://github.com/phpstan/phpstan/blob/af751d33d0dab6f257e7dc97f75cd371a3466562/.github/workflows/integration-tests.yml#L456-L485). This may also require to update secrets for the repository.

I did not test it in GitHub but run the full workflow locally using [ACT](https://github.com/nektos/act). There was an issue with the PHP extensions configuration but I think it is related to local workflow environment implementation (For some reason PHP setup without GD and PDO_sqlite which prevents koel from installing by composer, I bypassed it with  `--ignore-platform-reqs` flag)

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -15,7 +15,7 @@ on:
 
 jobs:
   integration-tests:
-    name: "Test with ${{ matrix.repo }} on PHP ${{ matrix.php-version }}"
+    name: "Test with ${{ matrix.repository }} on PHP ${{ matrix.php-version }}"
 
     runs-on: "ubuntu-latest"
     timeout-minutes: 30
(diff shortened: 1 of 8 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -82,6 +80,17 @@ jobs:
...
-        run: |
-          cd e2e/
-          composer exec phpstan
+        working-directory: e2e
+        run: composer exec phpstan analyse -- -c "../larastan/e2e/${{ matrix.config }}"
+
+      - name: "Generate baseline"
+        if: ${{ failure() }}
+        working-directory: e2e
+        run: composer exec phpstan analyse -- -c ../larastan/e2e/${{ matrix.config }} -b ../larastan/e2e/${{ matrix.baseline }}
```
canvural, line 89: Wouldn't this make the check always green? 🤔  For example I wouldn't be able to see if a PR has broke something in this analyzed repository.

If that is the case I think we can remove this part (also the uploading). And I can manually update the baseline from time to time.

canvural (changes requested):

canvural:
Awesome! Thank you very much. Just left a small question.

.github/workflows/e2e-tests.yml
sanfair: Let me test it real quick

szepeviktor: > And I can manually update the baseline from time to time.

I hope it will be "remove ignores from baseline"!

szepeviktor:
Please add this new directory to gitattributes to avoid exporting its contents.

.github/workflows/e2e-tests.yml
sanfair: This is how it is configured for the phpstan e2e test so maybe =)
But there has an issue with ignored errors. If an error is ignored in 3rd party repo (e.g. [monicahq/chandler](https://github.com/monicahq/chandler/blob/f57c49400dfa419e78bad240f018bce2949dcac2/phpstan.neon#L41)) and it is fixed by larastan the only way how we can ignore outdated errors is by completely disabling this check for repo (for example in [larastan](https://github.com/nunomaduro/larastan/blob/f8afd8fc55e5fd88b406827c9155c8c6d4f5fb02/e2e/monicahq-chandler.neon#L5 ) and in [phpstan](https://github.com/phpstan/phpstan/blob/af751d33d0dab6f257e7dc97f75cd371a3466562/e2e/integration/drupal.neon#L7))

canvural: Ok, I got it now. Thank you 👍🏽

canvural:
Awesome! Thank you!

PR #1698, opened 2023-07-21
Changed `retry()` helper stub to accept callable as `$sleepMilliseconds`

Plytas:
This PR fixes false negative saying that `Parameter $sleepMilliseconds of function retry expects int, Closure given.`

image

- [x] Added or updated tests

**Changes**

Renamed `$sleep` to `$sleepMilliseconds` and updated stub to accept callable.
[Link to `retry()` helper in laravel](https://github.com/laravel/framework/blob/c59d540ab142d36ee0f62c3ddc749329a968466b/src/Illuminate/Support/helpers.php#L225)

image

**Breaking changes**

None

### DIFF
--- a/tests/Integration/data/helpers.php
+++ b/tests/Integration/data/helpers.php
@@ -6,3 +6,14 @@ function transformAcceptsNonNullableCallable(): void
 {
     transform(User::first(), fn (User $user) => $user->toArray());
 }
+
+function retryAcceptsCallableAsSleepMilliseconds(): void
+{
+    retry(5, function (int $attempt): bool {
+        return false;
+    }, function (int $attempt, Exception $e): int {
+        return 0;
+    }, function (Exception $e): bool {
+        return true;
+    });
+}
--- a/tests/Integration/data/helpers.php
+++ b/tests/Integration/data/helpers.php
@@ -11,9 +11,9 @@ function retryAcceptsCallableAsSleepMilliseconds(): void
 {
     retry(5, function (int $attempt): bool {
         return false;
-    }, function (int $attempt, Exception $e): int {
+    }, function (int $attempt, \Exception $e): int {
         return 0;
-    }, function (Exception $e): bool {
+    }, function (\Exception $e): bool {
         return true;
     });
 }
--- a/stubs/Helpers.stub
+++ b/stubs/Helpers.stub
@@ -100,13 +100,13 @@ function rescue(callable $callback, $rescue = null, $report = true) {}
  * @template TValue
  * @param int|array<int, int> $times
  * @param callable(int): TValue $callback
- * @param int $sleep
+ * @param int|callable(int, \Exception): int $sleepMilliseconds
  * @param null|callable(\Exception): bool $when
  * @phpstan-return TValue
  *
  * @throws \Exception
  */
-function retry($times, callable $callback, $sleep = 0, $when = null) {}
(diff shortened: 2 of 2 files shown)


Plytas:
Couldn't find tests for stubs, only return type assertions for `retry()` helper. If anyone can point me to tests that need to be added, let me know.

sanfair:
Hi, check `retryHelper` test in `tests/Type/data/helpers.php.

Plytas:
I see. So I could add another test like this:
```PHP
assertType('bool', retry(5, function (int $attempt): bool {
    return false;
}, function (int $attempt, Exception $e): int {
    return 0;
}, function (Exception $e): bool {
    return true;
}));
```

And it would pass, because original phpdoc on laravel helper does accept `\Closure`. It's the stub in larastan that's wrong. Initially I wanted to start this PR with just a failing test, but couldn't find a way to test the stub itself.

Should I still add the test I suggested?

sanfair:
Just checked my suggestion. Turns out that even if you provide an invalid type for arguments it does not trigger any error in tests or type tests. 

Check highlighted text.
image

Plytas:
Looks like phpstan is not even running on `/tests`.

szepeviktor:
> Looks like phpstan is not even running on `/tests`.

https://github.com/nunomaduro/larastan/blob/58636034e8fcb8226254fb6f4b56d30e359d6ef3/phpstan.neon.dist#L7-L8

Plytas:
I tried running it for just this `helpers.php` test file
```SHELL
vendor/bin/phpstan analyse tests/Type/data/helpers.php
```

But there's just errors all over the place, it's not using the stubs
image

sanfair:
Found the correct file. `tests/Integration/data/helpers.php`

Plytas:
@sanfair can you elaborate? I only see one unrelated test there.
image

Plytas:
I was able to run phpstan just for that test file and tell it to use `Helpers.stub`.

I can see the error before adjusting the stub:
image

And it's gone after using stub in this PR:
image

But it was a bit hacky to verify this and there's still other tests and stubs unaccounted for.

sanfair:
@Plytas, there is only a test for the `transform` helper yet. 

If you paste the next code there and it will fail tests (because the `$times` type is invalid)
```php

function retryHelper(): void
{
    retry(3, function () {});
    retry('3', function () {});
    retry(new \stdClass(), function () {});
}
```
Add needed modification to test related to your code changes, e.g test that you can pass Closure to `retry` function

Plytas:
@sanfair Thanks so much, I have added a test and verified locally that it fails without stub changes and works with stub changes.

tests/Integration/data/helpers.php
```diff
@@ -6,3 +6,14 @@ function transformAcceptsNonNullableCallable(): void
...
     transform(User::first(), fn (User $user) => $user->toArray());
 }
+
+function retryAcceptsCallableAsSleepMilliseconds(): void
+{
+    retry(5, function (int $attempt): bool {
+        return false;
+    }, function (int $attempt, Exception $e): int {
+        return 0;
+    }, function (Exception $e): bool {
```
villfa, line 16: `Exception` doesn't exist in the current namespace:
```suggestion
    }, function (int $attempt, \Exception $e): int {
        return 0;
    }, function (\Exception $e): bool {
```

canvural:
Thank you!

PR #1706, opened 2023-08-02
feat: add dynamic return type extension for str helper

Daanra:
- [x] Added or updated tests

**Changes**

Adds a return type extension such that code like:
```php
str('foo')->contains('bar')
```
won't fail because PHPStan thinks `str('foo')` resolves to `mixed`

Note that ideally we'd also properly resolve `str()`, right now I'm just leaving it as mixed.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -349,6 +349,11 @@ services:
             - phpstan.broker.dynamicFunctionReturnTypeExtension
 
     -
+        class: NunoMaduro\Larastan\ReturnTypes\Helpers\StrExtension
+        tags:
+            - phpstan.broker.dynamicFunctionReturnTypeExtension
+
+    -
         class: NunoMaduro\Larastan\ReturnTypes\Helpers\TapExtension
         tags:
             - phpstan.broker.dynamicFunctionReturnTypeExtension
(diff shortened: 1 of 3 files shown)


szepeviktor:

I encourage you to write OOP code without global functions.
`Illuminate\Support\Str` does not keep secrets, global functions do.
I am thinker, not a doer. So millions of people will attack me because of this statement.

canvural:
Hi,

I think we can do this in a stub file with conditional return types. No need for an extension. What do you think?

Daanra:
> I think we can do this in a stub file with conditional return types. No need for an extension. What do you think?

That would be great. But can we check for the number of arguments passed with conditional return types? I couldn't find anything on it.

canvural:
I thought we can check for `null` Can't we?

I'm not near a computer so didn't look for the actual function definition. Does it use func_num_args?

Daanra:
> I'm not near a computer so didn't look for the actual function definition. Does it use func_num_args?

Yes it does:
```php
function str($string = null)
{
        if (func_num_args() === 0) {
            return new class
            {
                public function __call($method, $parameters)
                {
                    return Str::$method(...$parameters);
                }

                public function __toString()
                {
                    return '';
                }
            };
        }

        return Str::of($string);
}
```

So `str(null)` actually returns a `Stringable`.

I considered adding a stub like so:
```php
/**
 * @param string|null $value
 * @return ($value is null ? mixed : \Illuminate\Support\Stringable)
 */
function str($value) {}
```
But this would not work well when you're passing a nullable string to `str`.

canvural:
Thank you!

PR #1711, opened 2023-08-07
feat: add test for dropColumn() when an array is passed

EtraEllecon:
- [x] Added or updated tests

**Changes**

This PR adds a test to confirm that an array of column names can be passed to `dropColumn()` and that the columns are indeed dropped by asserting the count/remaining columns.

### DIFF
--- a/tests/Unit/data/migrations_using_drop_column/2023_08_07_000000_create_add_drop_to_users_table.php
+++ b/tests/Unit/data/migrations_using_drop_column/2023_08_07_000000_create_add_drop_to_users_table.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Unit\MigrationsUsingDropColumn;
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class CreateAddDropToUsersTable extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('users', static function (Blueprint $table) {
+            $table->bigIncrements('id');
+            $table->string('name')->nullable();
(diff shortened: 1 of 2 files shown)


tests/Unit/data/migrations_using_drop_column/2023_08_07_000000_create_add_drop_to_users_table.php
```diff
@@ -0,0 +1,34 @@
...
+        Schema::create('users', static function (Blueprint $table) {
+            $table->bigIncrements('id');
+            $table->string('name')->nullable();
+            $table->string('email')->unique();
+            $table->timestamps();
+        });
+
+        Schema::table('users', static function (Blueprint $table) {
+            $table->string('country')->nullable();
+            $table->string('city')->nullable();
```
canvural, line 27: I think it would be better if this was also in the previous block. That way, we can be extra sure `dropColumn` works as expected.

canvural (changes requested):

tests/Unit/data/migrations_using_drop_column/2023_08_07_000000_create_add_drop_to_users_table.php
EtraEllecon: Done.

canvural: Actually 8852c3cbbb074bf9d0f493f5b7d1da6d172b3590 was enough. We don't need the lines 27 to 30 now.

szepeviktor (approved):

tests/Unit/data/migrations_using_drop_column/2023_08_07_000000_create_add_drop_to_users_table.php
EtraEllecon: Roger! Reverted.

PR #1712, opened 2023-08-07
fix: Fixed false detection of columns being nullable

nikolicaleksa:
- [x] Added or updated tests

**Changes**

Fixed a problem that occurred when scanning database migrations. Columns which had `nullable(false)` were incorrectly inferred as nullable.

**Breaking changes**

n/a


canvural (approved):
Perfect. Thank you!

PR #1714, opened 2023-08-11
feat: Support migrations with conditional statements

nikolicaleksa:
- [x] Added or updated tests

Larastan is already doing an awesome job in understanding the table structure and model properties by scanning the migrations, however, a problem arises when there are if conditions inside the migrations. Statements which are under an if condition do not get processed and therefor errors like "Undefined property" for models start to pop up.

One might use if conditions inside the migration files in order to make sure no errors will happen if you accidentally run the same migration multiple times or your database was manually altered (table / columns were deleted / altered by hand).

In my day job, I'm using if conditions in migrations before changing anything as our project is using multi-tenancy and someone is always creating new tenants by manually creating new databases. Because of that, sometimes a situation happens where the new tenant database is up-to-date with the changes, however, that's not reflected in the migrations table. Because of that, our deploys were sometimes failing and we decided to write migrations in a way where errors would not happen and deploy would not fail if some change is already done but not reflected in the migrations table.

**Breaking changes**

n/a

### DIFF
--- a/docs/custom-config-parameters.md
+++ b/docs/custom-config-parameters.md
@@ -20,6 +20,8 @@ parameters:
         - app/Domain/DomainB/migrations
 ```
 
+**Note:** If your migrations are using `if` statements to conditionally alter database structure (ex: create table only if it's not there, add column only if table exists and column does not etc...) Larastan will assume those if statements evaluate to true and will consider everything from the `if` body.
+
 ## `disableMigrationScan`
 **default**: `false`
 
--- a/src/Properties/SchemaAggregator.php
+++ b/src/Properties/SchemaAggregator.php
@@ -550,7 +550,7 @@ final class SchemaAggregator
     }
 
     /**
-     * @param PhpParser\Node\Expr $updateClosure
+     * @param  PhpParser\Node\Expr  $updateClosure
      *
      * @return PhpParser\Node\Stmt\Expression[]
      */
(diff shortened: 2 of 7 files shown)


szepeviktor:

You git client may have a configuration problem (.dev domain)

nikolicaleksa:
Hey szepeviktor, not sure I understand what you mean by Git problem and .dev domain. Can you elaborate further?

szepeviktor:
These three commits are made with an email address not registered here on GitHub.

nikolicaleksa:
Oh, I see. Yeah you are right, thanks for pointing that out.. I haven't used this laptop for Github before creating PRs here so that's the reason. I've registered the email with my Github account now. As I'm not familiar with Github would you say that's enough or do I also need to make it a primary email?

szepeviktor:
> I've registered the email with my Github account

That is okay.

nikolicaleksa:
Hey @szepeviktor, do I need to make some adjustments to the MR? Just checking as there are no updates since I've opened the MR

szepeviktor:
No. Please wait for a review.

canvural:
Hi,

Sorry for the late review.

The code looks good. But just not sure if it is gonna have any side effects. I'll merge it now though, and test it with couple of projects. So, just to let you know if there are gonna be any side effects that people report after the release, I'll rollback the changes 😄 

Thank you!

nikolicaleksa:
@canvural Now that I think about it, one side effect that might happen is the following: If someone was using conditional migrations and in order to get around errors about undefined model properties they used `@phpstan-ignore-line` now Larastan will complain about no error to ignore on that line (unless turned off with `reportUnmatchedIgnoredErrors`). However, I would say this should be the expected behaviour as now there really is no error to ignore on that line (maybe an explanation in release notes would be good so that people don't get confused).

PR #1715, opened 2023-08-12
Initial support for `literal-string` on Illuminate Raw methods

craigfrancis:
- [ TODO ] Added or updated tests
- [ TODO ] Documented user facing changes

Ref the suggestion to Laravel core for [Addressing SQL Injection Vulnerabilities with Raw Methods](https://github.com/laravel/framework/discussions/47257).

[Previously discussed](https://github.com/canvural/larastan-strict-rules/issues/153) with @canvural, where this PR is currently a draft (need to confirm if this is how optional stubs should be added).

I've not really looked at tests yet (first time looking at this project), but I suspect it will be an [Integration test](https://github.com/nunomaduro/larastan/blob/master/tests/Integration/IntegrationTest.php), where I'll probably need to "compare errors".

**Changes**

Adds `checkRawLiteralString`, which ensures the `*Raw()` methods are given a `literal-string`.

The `literal-string` type represents a string that has been defined by the developer, even if it has been concatenated with similar strings. It makes it easy to prove no unsafe user values (e.g. from `$_GET`) have been included in the SQL string.

**Breaking changes**

It should only identify mistakes; but to be cautious, the flag is off by default.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -24,6 +24,7 @@ parameters:
     checkModelProperties: false
     checkPhpDocMissingReturn: false
     checkUnusedViews: false
+    checkRawLiteralString: false
 
 parametersSchema:
     checkOctaneCompatibility: bool()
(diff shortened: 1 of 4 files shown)


nunomaduro:

PR #1723, opened 2023-08-17
Add stub for `FormRequest::validated`

fyrts:
- [X] Added or updated tests

**Changes**

I've added `validated` to the `FormRequest` stub, as I was running into issues with its current `mixed` return type:

```php
$item = Item::create($request->validated());
```

> Parameter $attributes of static method Illuminate\Database\Eloquent\Builder<App\Models\Item>::create() expects array<string, mixed>, mixed given. 

**Breaking changes**


canvural (approved):

canvural:
Thank you!

PR #1729, opened 2023-08-23
Fix Rescue param $report

fotrino:
- [ ] Added or updated tests

**Changes**

Larastan's stub for the `rescue()` helper does not accept a Closure on the `$report` param and give the following error.
```
 ------ -------------------------------------------------------------------
  41     Parameter $report of function rescue expects bool, Closure given.
 ------ -------------------------------------------------------------------
```

This an example from the Laravel Docs, to use a Closure on the mentioned param.

```php
return rescue(function () {
    return $this->method();
}, report: function (Throwable $throwable) {
    return $throwable instanceof InvalidArgumentException;
});
```

**Breaking changes**


nunomaduro:

PR #1734, opened 2023-09-05
route-name-type and stubs for Illuminate\Routing

NiclasvanEyk:
- [x] Added or updated tests

**Changes**

Similar to how view names are checked when using `view`, the functions that generate URLs (`route`, `to_route`, `Url::signedRoute`, etc.) are now checked too. This way a developer gets an error when referring to a route that does not exist.

**Examples**

Created a fresh Laravel application with the following controller:

```php
<?php

namespace App\Http\Controllers;

class ExampleController
{
    public function __invoke(): \Illuminate\Http\RedirectResponse
    {
        return \to_route('non-existing');
    }
}
```

Running PHPStan yields the following:
```
phpstan --no-progress
Note: Using configuration file /private/tmp/laravel/larastan-routes-test/phpstan.neon.dist.
 ------ -----------------------------------------------------------------------------------
  Line   Http/Controllers/ExampleController.php
 ------ -----------------------------------------------------------------------------------
  9      Parameter #1 $route of function to_route expects route-name-string, string given.
         💡 Route 'non-existing' does not exist.
 ------ -----------------------------------------------------------------------------------

 ------ -------------------------------------------------------------------------------
  Line   Http/Middleware/Authenticate.php
 ------ -------------------------------------------------------------------------------
  15     Parameter #1 $name of function route expects route-name-string, string given.
         💡 Route 'login' does not exist.
 ------ -------------------------------------------------------------------------------


 [ERROR] Found 2 errors

```

The newly introduced type also produces helpful error messages in case of typos. If we introduce a route with a similar name:

```php
// routes/web.php

Route::get('/is-existing', function () {})->name('is-existing');
```

The error message changes:
```
phpstan --no-progress
Note: Using configuration file /private/tmp/laravel/larastan-routes-test/phpstan.neon.dist.
 ------ -----------------------------------------------------------------------------------
  Line   Http/Controllers/ExampleController.php
 ------ -----------------------------------------------------------------------------------
  9      Parameter #1 $route of function to_route expects route-name-string, string given.
         💡 Route 'non-existing' does not exist. Did you mean 'is-existing'?
 ------ -----------------------------------------------------------------------------------

 ------ -------------------------------------------------------------------------------
  Line   Http/Middleware/Authenticate.php
 ------ -------------------------------------------------------------------------------
  15     Parameter #1 $name of function route expects route-name-string, string given.
         💡 Route 'login' does not exist.
 ------ -------------------------------------------------------------------------------


 [ERROR] Found 2 errors

```

**Breaking changes**

This could break CI when people upgrade, since it introduces new checks and non-existing routes previously did not lead to errors. However, I think this should not be considered breaking, since the application was in a broken state in the first place. Since it uses the actual router instance of the application, the amount of false-positives should be low as well.


szepeviktor:
Thank you for your PR @NiclasvanEyk!

Your [master branch](https://github.com/NiclasvanEyk/larastan/tree/master) is lagging behind.

NiclasvanEyk:
> Thank you for your PR @NiclasvanEyk!
> 
> Your [master branch](https://github.com/NiclasvanEyk/larastan/tree/master) is lagging behind.

Thanks for the heads-up, I could have sworn I just forked it yesterday, but the 300 commits say otherwise 😅.

-------------

@mad-briller yeah that's what I wanted to add as well, but I did not have time, so I thought I'd just open a draft PR and request reviews when it is actually ready. I'll see if I get to update my fork tomorrow and then tackle the testing part.

szepeviktor:
This route may not be found ...
https://github.com/laravel/fortify/blob/74cd344aea9ab4abb5ec09f7b3349bf3bfc8bbdc/routes/routes.php#L31

NiclasvanEyk:
> This route may not be found ... https://github.com/laravel/fortify/blob/74cd344aea9ab4abb5ec09f7b3349bf3bfc8bbdc/routes/routes.php#L31

Yes, I am also running into a similar problem for the default Laravel skeleton, where the `Authenticate` middleware wants to redirect to `'login'`, however no such route is defined by default.

Do you have any suggestions on how to deal with these kinds of errors? We could just add 'default' ignore statements to the `extensions.neon`.

szepeviktor:
Please add to https://github.com/nunomaduro/larastan/blob/master/phpstan-baseline.neon

NiclasvanEyk:
Does this fix the issue? As I understand it, https://github.com/nunomaduro/larastan/blob/master/phpstan-baseline.neon is used for running `phpstan` on the `larastan` source itself. 

The error arises only when analyzing the user's source, so we'd need to use something like 

```neon
parameters:
	ignoreErrors:
		-
			message: "#^Parameter #1 $name of function route expects route-name-string, string given.$#"
			count: 1
			path: app/Http/Middleware/Authenticate.php
```

to the `extension.neon`, so users use this configuration by default. Or am I misunderstanding something here?

szepeviktor:
I am sorry.
This is where it goes: https://github.com/nunomaduro/larastan/blob/master/tests/laravel-test.sh#L18-L23

tests/laravel-test.sh
```diff
@@ -21,6 +21,12 @@ parameters:
     level: 5
     paths:
         - app/
+    parameters:
+        ignoreErrors:
+            -
+                message: "#^Parameter \\#1 \\$name of function route expects route\\-name\\-string, string given\\.$#"
+                count: 1
+                path: app/Http/Middleware/Authenticate.php
```
szepeviktor, line 29:
```suggestion
    ignoreErrors:
        -
            message: "#^Parameter \\#1 \\$name of function route expects route\\-name\\-string, string given\\.$#"
            count: 1
            path: app/Http/Middleware/Authenticate.php
```

szepeviktor (approved):

NiclasvanEyk:
Just noticed that since the routes are passed as a property, they are cached and not properly re-computed across analysis runs. 

Will push a fix later that re-computes it every run, similar to how the `ViewStringType` does

nunomaduro:

crissi:
Hi @NiclasvanEyk,  seems like a great pr to included in larastan, up for opening it up again?

PR #1757, opened 2023-10-03
Update monicahq-chandler.neon

szepeviktor:
Just testing ...


szepeviktor:
Spotted https://github.com/monicahq/chandler/blob/f57c49400dfa419e78bad240f018bce2949dcac2/phpstan.neon#L40-L41

PR #1758, opened 2023-10-05
Test against php 8.3

sergiy-petrov:
- [x] Added or updated tests

**Changes**

Test against php 8.3

**Breaking changes**

<!-- Are existing use cases affected and require changes when upgrading? 
No


szepeviktor:
We are almost there.
https://www.php.net/archive/2023.php#2023-09-28-2

PR #1760, opened 2023-10-10
Fix collection methods unwrapping outer generic

rovansteen:
- [X] Added or updated tests

Found an issue with working with multiple layers of generics in combination with collections. 

**Changes**

This PR fixes an issue while working with a `Collection<X<Y>>` where it makes the wrong assumption that X is a Collection and transforms it in `Collection<Collection<Y>>`. Adding an explicit check for the type before transforming it fixes it.

**Breaking changes**
—

### DIFF
--- a/src/ReturnTypes/CollectionGenericStaticMethodDynamicMethodReturnTypeExtension.php
+++ b/src/ReturnTypes/CollectionGenericStaticMethodDynamicMethodReturnTypeExtension.php
@@ -123,7 +123,9 @@ class CollectionGenericStaticMethodDynamicMethodReturnTypeExtension implements D
                 if ($type instanceof GenericObjectType && (($innerTypeReflection = $type->getClassReflection()) !== null)) {
                     $genericTypes = $innerTypeReflection->typeMapToList($innerTypeReflection->getActiveTemplateTypeMap());
 
-                    return new GenericObjectType($classReflection->getName(), $genericTypes);
+                    if ($classReflection->isSubclassOf($type->getClassName())) {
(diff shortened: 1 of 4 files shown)


src/ReturnTypes/CollectionGenericStaticMethodDynamicMethodReturnTypeExtension.php
```diff
@@ -123,7 +123,9 @@ private function handleGenericObjectType(ClassReflection $classReflection, Class
                 if ($type instanceof GenericObjectType && (($innerTypeReflection = $type->getClassReflection()) !== null)) {
                     $genericTypes = $innerTypeReflection->typeMapToList($innerTypeReflection->getActiveTemplateTypeMap());
 
-                    return new GenericObjectType($classReflection->getName(), $genericTypes);
+                    if ($classReflection->isSubclassOf($type->getClassName())) {
```
canvural, line 126: What happens if this condition is not met? And in which cases we would hit the `else` here? 🤔

rovansteen: If the `$classReflection` is not a subtype of `Collection` we will just traverse through the types and not modify them. This is what is (correctly) happening in the test cases that I added where it traverses through `Ok` => `Some` and keeps them as they are as they are not a subtype of `Collection`.

robertvansteen:
@canvural any further thoughts on this?

canvural:
Thank you!

PR #1770, opened 2023-10-25
Add tests to reproduce issue #1769

nie7321:
- [x] Added or updated tests
- [ ] Documented user facing changes

**Changes**

Adds a test demonstrating the issue in #1769.

**Breaking changes**

PR #1772, opened 2023-10-26
Fix snake count

### LINKED ISSUE #1769: Eloquent relationship _count property only works when relationship is camelCase
- Larastan Version: 2.6.4
- `--level` used: 5
- Pull request with failing test: https://github.com/nunomaduro/larastan/pull/1770

### Description
Eloquent will automatically provide a `{relationship}_count` property for relationships when `withCount()` is called. Larastan added support for this earlier in the year, but it looks like it is only able to support relationships named in `camelCase`. 

Relationships may be named in `snake_case`. When this is done, Larastan is incorrectly indicating that these properties do not exist.

```php
class Profile extends Model {
    public function upcoming_milestones(): HasMany {
        // . . .
    }
}

$foo = Profile::withCount('upcoming_milestones')->first();

// Access to an undefined property
$foo->upcoming_milestones_count;
```

mad-briller:
- [x] Added or updated tests

Fixes https://github.com/nunomaduro/larastan/issues/1769

Test provided by @nie7321 in https://github.com/nunomaduro/larastan/pull/1770

**Changes**

Fixes `_count` properties when the relation method is snake-cased

### DIFF
--- a/src/Properties/ModelRelationsExtension.php
+++ b/src/Properties/ModelRelationsExtension.php
@@ -52,22 +52,24 @@ final class ModelRelationsExtension implements PropertiesClassReflectionExtensio
         }
 
         if (str_ends_with($propertyName, '_count')) {
-            $propertyName = Str::camel(Str::before($propertyName, '_count'));
+            $propertyName = Str::before($propertyName, '_count');
         }
 
-        $hasNativeMethod = $classReflection->hasNativeMethod($propertyName);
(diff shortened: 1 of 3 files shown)


szepeviktor:
Thank you!

Could we have a test with camelCase too?

szepeviktor (approved):

canvural:
Thank you!

nie7321:
Thanks everyone! I appreciate your work on this!

PR #1779, opened 2023-11-17
Fix Redis facade tests

ondrejmirtes:
- [x] Added or updated tests

Fixing Redis tests after https://github.com/phpstan/phpstan-src/pull/2706


szepeviktor:
Thank you.

PR #1786, opened 2023-11-30
Update README.md

ymigval:
**Changes**

When specifying the package version, unexpected problems may occur, so it has been removed and composer has been left to decide.

**Breaking changes**

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Larastan was created by [Can Vural](https://github.com/canvural) and [Nuno Madur
 **1**: First, you may use [Composer](https://getcomposer.org) to install Larastan as a development dependency into your Laravel project:
 
 ```bash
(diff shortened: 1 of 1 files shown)


szepeviktor:
Thank you for the PR.

> unexpected problems may occur

Could you list one?

ymigval:
> Could you list one?

Good evening,

There may be a conflict in the versions of the project dependencies. Example, the package nunomaduro/larastan version 2.1.0 (which is the one you choose) requires the version of illuminate/console in the range of ^9; but some other package in the project that requires a different version of illuminate/console may conflict. To resolve this conflict you would have to update the dependencies, adjust the dependency versions manually, or revise the project dependencies.

All this is solved by just letting composer install the version of nunomaduro/larastan that suits you best.

canvural:
Hi,

I don't think this is needed. Thank you.

PR #1788, opened 2023-12-05
Add conflict nunomaduro/larastan

lloricode:
This update prevent installing both nunomaduro/larastan and larastan/larastan

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -67,6 +67,9 @@
             ]
         }
     },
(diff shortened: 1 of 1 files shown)


szepeviktor:
Could we have a `*` as version constraint please?

canvural:
Hi,

I don't think this is necessary.

PR #1791, opened 2023-12-05
Update README.md to fix github handle for Nuno

joelclermont:
I'm guessing the Github links for Nuno were replaced accidentally when updating the repo organization.


canvural (approved):
Good catch. Thank you.

PR #1802, opened 2023-12-13
Fix null/void problem (with latest PHPStan release)

Brenneisen:
- [x] Added or updated tests

**Changes**

Fixes broken tests due to a change in PHPStan 1.10.50 (https://github.com/phpstan/phpstan/issues/10291).

**Breaking changes**

None


szepeviktor:
This could be released only with required PHPStan version bump.

Brenneisen:
@nunomaduro Thank you for merging this! Could you also take a look at https://github.com/larastan/larastan/pull/1792? It is really important for us that it can be released shortly.

PR #1808, opened 2023-12-29
feat: Support addColumn in migration helper

nikolicaleksa:
- [x] Added or updated tests

**Changes**

Process `addColumn` method when building tables from migrations


canvural:
Thank you.

PR #1828, opened 2024-02-07
feat: detect usage of env() function outside of config folder

### LINKED ISSUE #1695: Detect usage of `env()` function outside from config folder
Hello, 

In order to be able to use the `config:cache` command, we need to ensure that our project only use the `env()` function inside config files.
See related laravel documentation: https://laravel.com/docs/10.x/configuration#configuration-caching

Unfortunately, it seems that many developers are not aware about this and I can find usage of `env()` function in most of the projects I joined in the last few years.
I am looking for a solution to detect usage of `env()` outside of config files and make our CI to fail if that happens, so we can safely use the `config:cache` command during deployments.

I did not find any rule to detect that in Larastan and I was thinking that it could be a good feature to add.

calebdw:
- [x] Added or updated tests

Closes #1695.

**Changes**

Hello!

This PR adds a rule to report any `env` calls outside the config directory. Currently it just checks for a `config` directory in the file name where the call is located. A more robust solution might be to use the container to grab the configured config path? However this is more complicated and I wasn't sure how exactly to do such a thing.

Thanks!

### DIFF
--- a/src/Rules/NoEnvCallsOutsideOfConfigRule.php
+++ b/src/Rules/NoEnvCallsOutsideOfConfigRule.php
@@ -1,0 +1,60 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Larastan\Larastan\Rules;
+
+use Larastan\Larastan\Concerns\HasContainer;
+use PhpParser\Node;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Name;
+use PHPStan\Analyser\Scope;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleError;
+use PHPStan\Rules\RuleErrorBuilder;
+
+/**
+ * Catches `env()` calls outside of the config directory.
+ *
+ * @implements Rule<FuncCall>
+ */
+class NoEnvCallsOutsideOfConfigRule implements Rule
+{
+    use HasContainer;
+
+    public function getNodeType(): string
+    {
+        return FuncCall::class;
+    }
+
+    /** @return array<int, RuleError> */
+    public function processNode(Node $node, Scope $scope): array
+    {
+        $name = $node->name;
+
+        if (! $name instanceof Name) {
+            return [];
+        }
+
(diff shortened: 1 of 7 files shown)


calebdw:
Looks like tests are currently failing on `2.x` for L11

szepeviktor:
Yes, the scheduled workflow also fails
https://github.com/larastan/larastan/actions/runs/7807784230/job/21296839699

canvural:
Hi,

I think we can use `config_path()` function and do the comparison with that.

Also I do not remember exactly now, but we might need to check that the function is actually coming from Laravel. Or at least that the function is in global namespace and not in `Some\Namespace\env`

Lastly, I'd prefer that it is disabled by default.

calebdw:
Done

canvural (approved):

szepeviktor (changes requested):
Laravel 9 has no `useConfigPath`

canvural:
What's useConfigPath?

szepeviktor:
A method called in this PR: tests/Rules/NoEnvCallsOutsideOfConfigRuleTest.php

src/Rules/NoEnvCallsOutsideOfConfigRule.php
```diff
@@ -0,0 +1,60 @@
...
+    /** @return array<int, RuleError> */
+    public function processNode(Node $node, Scope $scope): array
+    {
+        $name = $node->name;
+
+        if ($name instanceof Expr) {
+            return [];
+        }
+
+        if ($name->toString() !== 'env') {
```
mad-briller: is this check adequate if an `use function Foo\env;` construct is used to import the function rather than fully qualifying it at the call site?

calebdw: good catch! it did not

calebdw:
@szepeviktor should be fixed now, thanks for catching!

szepeviktor (approved):

canvural:
Thank you!

Issue #1774, opened 2023-11-09
Undefined property `pivot`


calebdw:
- Larastan Version: 2.6.4
- `--level` used: 2

### Description
Hello!

PHPStan is complaining about accessing the `pivot` property on a model returned from a `BelongsToMany` relation which has the proper doc type:

```php
    /** @return BelongsToMany<Permission> */
    public function permissions(): BelongsToMany
    {
        return $this->belongsToMany(
            Permission::class,
            'users_permission',
            'user_id',
            'permission_id',
        )->withPivot('level');
    }

```

This is an issue that goes back several years and has not yet been solved (#1308, #515, #883). I understand that I can use `getPropertyValue` but this is a dirty workaround and not an acceptable solution.

Thanks!

juliangums:
This would be great.

rvoortman:
Yes that would be amazing!

EliW:
Just adding some voice here.   I just ran into this while trying to get this codebase to be phpstan:2 compliant.   Yes, there is a workaround ... but the workaround actually doesn't 'do' anything?

Example code:
`                        return $tenant->pivot->type;`

Wherein the pivot has a custom field of 'type' added to it

For me, issuing the getPropertyValue, doesn't work, complaining it's undefined:
`                        return $tenant->getPropertyValue('pivot')->type;`

The getRelationValue works:
`                        return $tenant->getRelationValue('pivot')->type;`

However this seems to miss quite a few potential errors now, because the following should error in phpstan, and doesn't:
`                        return $tenant->getRelationValue('pivot')->not_valid;`

Since it doesn't appear to be now checking what the items returned actually are.    And if you use a different string (typo the 'pivot'), it also should error, and doesn't:
`                        return $tenant->getRelationValue('pivvot')->type;`

Which means that in practice, the workaround is no different than just ignoring the error completely.  It doesn't provide any static analysis.

juliangums:
@EliW that's correct.

sinnbeck:
I was just having this issue, and came up with this solution which seems to work.

First I rename the pivot, define my intermediate class, and load the pivot I need
```php
/** @return BelongsToMany<Customer> */
    public function customers(): BelongsToMany
    {
        return $this->belongsToMany(Customer::class)
            ->using(CustomerFile::class)
            ->as('access')
            ->withPivot('expires_at');
    }
```
And I then add it as a `@property` to the class
```php
/**
 * @property CustomerFile $access
 */
class File extends Model
[...]

Issue #1308, opened 2022-07-04
Pivot with a custom name is not taken into account


pacoorozco:
- Larastan Version: 1.7.13
- `--level` used: 8
- Pull request with failing test: https://github.com/pacoorozco/gamify-laravel/pull/268

### Description

```
 ------ -------------------------------------------------------------------- 
  Line   app/Models/User.php                                                 
 ------ -------------------------------------------------------------------- 
  285    Access to an undefined property Gamify\Models\Question::$response.  
 ------ -------------------------------------------------------------------- 
```

### Laravel code where the issue was found

```php
283: public function getResponseForQuestion(Question $question): ?UserResponse
284:    {
285:        return $this->answeredQuestions()
286:            ->where('question_id', $question->id)
287:            ->first()
288:            ?->response;
289:    }
290:
291: public function answeredQuestions(): BelongsToMany
292:     {
293:        return $this->belongsToMany(
294:            Question::class,
295:            'users_questions',
296:            'user_id',
297:            'question_id'
298:        )
299:            ->as('response')
300:            ->withPivot('points', 'answers')
301:            ->using(UserResponse::class);
302:    }
```

szepeviktor:
I support a project where this can be seen also: `as()` method is not recognized.

pacoorozco:
And have you found any way to fix it? Or Should I ignore these messages?

szepeviktor:
I would love to see this resolved!

PR #1833, opened 2024-02-09
feat: support relation `withDefault`

### LINKED ISSUE #1669: Default Models
- Larastan Version: v10.13.5
- `--level` used: 9

### Description

Use of https://laravel.com/docs/10.x/eloquent-relationships#default-models is supposed to allow removal of conditional / null-safe code. Larastan errors stating that `null` is still returned, when in fact `null` is never returned if `withDefault()` is used:
```
Cannot access property $name on App\Models\Group|null.
```

```php
    /**
     * @return BelongsTo<Group, User>
     */
    public function group(): BelongsTo
    {
        return $this->belongsTo(Group::class)->withDefault(['name' => 'Needs Triage']);
    }

    public function groupName(): string
    {
        return $this->group->name;
    }
```

calebdw:
- [x] Added or updated tests

Closes #1669 

**Changes**

This checks every relation to see if `withDefault` is called, and if so, returns the relation type without the null union.

Note, this is now (potentially) parsing the file twice for each relation: once for the related model name and then again to check for the default method. It might be worth trying to only parse once.

Thanks!

### DIFF
--- a/src/Types/RelationParserHelper.php
+++ b/src/Types/RelationParserHelper.php
@@ -6,6 +6,8 @@ namespace Larastan\Larastan\Types;
 
 use PhpParser\Node;
 use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Stmt\ClassMethod;
+use PhpParser\Node\Stmt\Return_;
 use PhpParser\NodeFinder;
 use PHPStan\Analyser\ScopeContext;
 use PHPStan\Analyser\ScopeFactory;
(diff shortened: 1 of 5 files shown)


src/Types/RelationParserHelper.php
```diff
@@ -112,24 +92,91 @@ public function findRelatedModelInRelationMethod(
         return $this->reflectionProvider->hasClass($returnClass) ? $returnClass : null;
     }
 
-    /**
-     * @param  string  $method
-     * @param  mixed  $statements
-     * @return Node|null
-     */
-    private function findMethod(string $method, $statements): ?Node
+    public function isRelationWithDefault(MethodReflection $methodReflection): bool
```
mad-briller, line 95: i'd be interested to see what the performance impact of this is on a relatively large codebase that is making heavy use of relations, there are ways we can mitigate the impact but i feel like there has to be a simpler way.

for example the `BelongsTo` could have it's template type widened to `?Model` rather than just `Model` and that would allow the developer writing the relationship to be explicit: `BelongsTo<Model>` when `withDefault()` and `BelongsTo<?Model>` without it.

however this requires a breaking change to how relationships are handled right now so i fear it's too late, but i wanted to put the idea forwards regardless

calebdw: After thinking about it more, parsing the file twice is just worst case---it could parse the file twice if the conditions align just right, or once, or it might not parse it at all.

Not sure that it is worth the optimization, but we might could cache the file parse for later use---but this would increase the memory kept around

canvural:
Hi,

While I understand the issue this is solving. I don't think this is the way to go.

We parse the model files, only if the relationship method does not provide any generic return type. This is a special case we take in order to make the developer's life easier. They don't need to provide the generic return type for relation methods. But with this change we'd definitely parse the file for every call now. 

Plus this goes against the philosophy of PHPStan itself. It does not descend down into the method bodies it analyses because of performance reasons.

We might solve this in version 3 though. By adjusting the generics and using similar techniques  that we use in factories.

Hope it makes sense. Thanks!

calebdw:
@canvural, I believe that the vast majority of developers do not add the generic type hints to their relationships (there was an issue last week where someone was strongly opposed to just adding the php return type, much less the generic!)---so in reality I believe that the models are already being parsed several times over.

I have been thinking about this, I think we need to implement caching here so that every method / model is parsed at most once (similar to parsing the migrations). This would also allow us to support other things in a single pass (`withDefault`, `as`, `withPivot`, etc)---I think this is also integral to solving #1308 & #1774.

calebdw:
@canvural, you've also mentioned `v3` a few times. Can you create a `3.x` branch (or `master`) so we can start making PRs against it? It would also be helpful to come up with a roadmap to v3 that we can start working towards.

canvural:
> I believe that the vast majority of developers do not add the generic type hints to their relationships (there was an issue last week where someone was strongly opposed to just adding the php return type, much less the generic!)

Then they should not even use a tool that's whole purpose is checking types 🙂

>  you've also mentioned v3 a few times. Can you create a 3.x branch (or master) so we can start making PRs against it? It would also be helpful to come up with a roadmap to v3 that we can start working towards.

I'd rather make an issue first where we can gather the ideas for v3. I'll try to open it this week.

PR #1834, opened 2024-02-09
feat: return array shape from model only method

### LINKED ISSUE #1645: Return a shape from `Model->only`
- Larastan Version: 2.6.0
- `--level` used: 9

### Description

Laravel types the return of `Model->only` as a plain array. In cases when the method is passed in a constant array, an array shape can be returned instead.

I've created a return type extension that I use in my projects, I'm just unsure when I'll have the time to create a test case for it to submit a PR. Sharing it here in case someone else wants to take it and run with it, otherwise I'll get to submitting a PR at some point.

```php
use PhpParser\Node\Expr\Array_;
use PhpParser\Node\Scalar\String_;
use PHPStan\Analyser\Scope;
use PHPStan\Reflection\MethodReflection;
use PHPStan\Type\ArrayType;
use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
use PHPStan\Type\Constant\ConstantStringType;
use PHPStan\Type\DynamicMethodReturnTypeExtension;
use PHPStan\Type\ErrorType;
use PHPStan\Type\MixedType;
use PHPStan\Type\Type;

class ModelOnlyDynamicReturnTypeExtension implements DynamicMethodReturnTypeExtension {
    public function getClass(): string {
        return Model::class;
    }

    public function isMethodSupported(MethodReflection $methodReflection): bool {
        return $methodReflection->getName() === 'only';
    }

    public function getTypeFromMethodCall(
        MethodReflection $methodReflection,
        MethodCall $methodCall,
        Scope $scope
    ): Type {
        $fields = $methodCall->getArgs();
        if (\count($fields) < 1) {
            return new ErrorType();
        }

        if ($fields[0]->value instanceof Array_) {
            $fields = $fields[0]->value->items;
        }
        $fields = \array_map(static fn ($f) => $f->value, \array_filter($fields));

        $model = $scope->getType($methodCall->var);
        $array = ConstantArrayTypeBuilder::createEmpty();
        foreach ($fields as $field) {
            if (! $field instanceof String_) {
                return new ArrayType(new MixedType(), new MixedType());
            }

            $name = $field->value;
            if (! $model->hasProperty($name)->yes()) {
                return new ArrayType(new MixedType(), new MixedType());
            }

            $array->setOffsetValueType(
                new ConstantStringType($name),
                $model->getProperty($name, $scope)->getReadableType(),
            );
        }

        return $array->getArray();
    }
}
```

calebdw:
- [x] Added or updated tests

Closes #1645 

**Changes**

Returns the array shape provided to `$model->only()`, credit goes to @simon-tma for writing the extension.

Thanks!

### DIFF
--- a/src/ReturnTypes/ModelOnlyDynamicMethodReturnTypeExtension.php
+++ b/src/ReturnTypes/ModelOnlyDynamicMethodReturnTypeExtension.php
@@ -7,21 +7,17 @@ namespace Larastan\Larastan\ReturnTypes;
 use Illuminate\Database\Eloquent\Model;
 use PHPStan\Type\NullType;
 use PHPStan\Type\StringType;
-use PhpParser\Node\Expr\Array_;
 use PhpParser\Node\Expr\MethodCall;
-use PhpParser\Node\Scalar\String_;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\MethodReflection;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
-use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
 use PHPStan\Type\ErrorType;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\Type;
 
-use function array_map;
-use function array_filter;
+use function array_reduce;
 use function count;
 
 final class ModelOnlyDynamicMethodReturnTypeExtension implements DynamicMethodReturnTypeExtension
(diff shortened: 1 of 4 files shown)


src/ReturnTypes/ModelOnlyDynamicMethodReturnTypeExtension.php
```diff
@@ -0,0 +1,76 @@
...
+
+        foreach ($fields as $field) {
+            if (! $field instanceof String_) {
+                return new ArrayType(new MixedType(), new MixedType());
+            }
+
+            $name = $field->value;
+
+            if (! $model->hasProperty($name)->yes()) {
+                return new ArrayType(new MixedType(), new MixedType());
```
mad-briller: this seems a bit drastic to bail entirely if a single property was not found, as opposed to setting the value as `NullType`

src/ReturnTypes/ModelOnlyDynamicMethodReturnTypeExtension.php
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        $fields = array_map(static fn ($f) => $f->value, array_filter($fields));
+
+        $model = $scope->getType($methodCall->var);
+        $array = ConstantArrayTypeBuilder::createEmpty();
+
+        foreach ($fields as $field) {
+            if (! $field instanceof String_) {
+                return new ArrayType(new MixedType(), new MixedType());
```
mad-briller: same with this, would i not be better to just `continue` and ignore the field if it's not a string?

mad-briller: so i just tested what laravel will do if given a non-string type to `only` and this is the behaviour:

```php
        dd($model->only(['test', 1]));
```

```
array:2 [
  "test" => null
  1 => null
]
```

src/ReturnTypes/ModelOnlyDynamicMethodReturnTypeExtension.php
calebdw: Yeah, I agree---I'll update it to return null

src/ReturnTypes/ModelOnlyDynamicMethodReturnTypeExtension.php
calebdw: It's not just if it's not a string, what about variables or expressions?

E.g., 

```php
$model->only([
    ...$columns,
    $var,
]);
```

calebdw: I updated it to return `array<string, mixed>` if we can't determine all of the columns---I don't like skipping over fields we can't parse and returning an array shape that is incorrect.

mad-briller: hmm that's an interesting point yeah;

you could use `$scope->getType($field)->getConstantStrings()` for the variadic or `$scope->getType($fields[0])->getItemType()->getConstantStrings()` for an array to determine if there are any known constant strings in the expression that is passed to only to try to be a bit more versatile, rather than relying on the AST node directly

that would allow you to properly detect unions of strings or other weird edge cases

calebdw: Done

tests/Type/data/model-methods.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace ModelRelations;
```
canvural:
```suggestion
namespace ModelMethods;
```

canvural (changes requested):
Looks good! Just left one remark. And if you can rebase on top of latest changes CI should run.

tests/Type/data/model-methods.php
calebdw: Ah, good catch!

canvural (approved):

canvural:
Thank you!

PR #1835, opened 2024-02-09
fix: check relation parent instead of string in name for many/morph

calebdw:
**Changes**

Hello! 

Please see https://github.com/staudenmeir/laravel-adjacency-list/issues/216 for context. Some packages don't follow the convention of using `Many` or `MorphTo` in the relation class name and so the string check breaks. This PR checks for the parent relation to be more robust.

Thanks!

### DIFF
--- a/src/Properties/ModelRelationsExtension.php
+++ b/src/Properties/ModelRelationsExtension.php
@@ -5,6 +5,12 @@ declare(strict_types=1);
 namespace Larastan\Larastan\Properties;
 
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use Illuminate\Database\Eloquent\Relations\HasMany;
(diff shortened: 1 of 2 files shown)


canvural:
Can we add a test with a custom relation class?

Also if you can rebase on top of latest changes CI should run.

calebdw:
Test added!

canvural (approved):
Thank you!

canvural:
@calebdw This broke handling of `HasOneThrough` Larastan now thinks it returns a collection, because that relation extends `HasManyThrough` Just writing in case you have time to fix it, if not I'll do it tomorrow 👍🏽

calebdw:
Interesting, most of the relations extends a HasOneOrMany type relation. I'll take a look and add a test so it doesn't happen

calebdw:
@canvural, see PR: https://github.com/larastan/larastan/pull/1841 for fix

PR #1837, opened 2024-02-09
fix: Carbon v3 macro reflection

calebdw:
Hello!

Carbon v3 switched from a static property to an instance property: https://github.com/briannesbitt/Carbon/commit/a12231089585d602c01bb4efa7c25f10bdbad0e2

This PR updates the macro extension to handle this change and should fix the pipeline.

Thanks!

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -2,6 +2,7 @@
 
 declare(strict_types=1);
 
+use Composer\InstalledVersions;
 use Illuminate\Contracts\Console\Kernel;
 use Illuminate\Contracts\Foundation\Application;
 use Laravel\Lumen\Application as LumenApplication;
@@ -30,4 +31,8 @@ if ($app instanceof Application) {
 
 if (! defined('LARAVEL_VERSION')) {
     define('LARAVEL_VERSION', $app->version());
+}
+
+if (! defined('CARBON_VERSION')) {
+    define('CARBON_VERSION', InstalledVersions::getVersion('nesbot/carbon'));
 }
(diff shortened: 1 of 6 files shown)


calebdw:
Why can the global `CARBON_VERSION` constant not be found? 

PHPStan v1.7.0 claims that nothing special is needed, but it's still defined in the bootstrap file anyways...
https://phpstan.org/user-guide/discovering-symbols#global-constants

calebdw:
Another thing to consider is that we might not have to support Carbon macros because Carbon publishes an `extension.neon` for their macros which is automatically installed when using the [phpstan/extension-installer](https://github.com/phpstan/extension-installer)

CanVuralStudocu:
Oops, I did not see this and fixed the same issue here: https://github.com/larastan/larastan/commit/6fb6803c039cdd163a657fad6758257b45f187de

> Another thing to consider is that we might not have to support Carbon macros because Carbon publishes an extension.neon for their macros which is automatically installed when using the [phpstan/extension-installer](https://github.com/phpstan/extension-installer)
That's something to consider for the future. Not sure if it does the same things exactly.

canvural:
Oops, I did not see this and fixed the same issue here: https://github.com/larastan/larastan/commit/6fb6803c039cdd163a657fad6758257b45f187de

> Another thing to consider is that we might not have to support Carbon macros because Carbon publishes an extension.neon for their macros which is automatically installed when using the [phpstan/extension-installer](https://github.com/phpstan/extension-installer)

That's something to consider for the future. Not sure if it does the same things exactly.

PR #1838, opened 2024-02-10
feat: create model appends rule

### LINKED ISSUE #1831: [Feature] Detect appended accessors
### Description
See https://github.com/laravel/framework/pull/50003 for context. Essentially only **computed properties** should be appended as accessors are already called when casting a model to an array. We should be able to detect and report this case.

On a side note, could also report entries in `$appends` that don't exist on the model.

calebdw:
Closes #1831 

- [x] Added or updated tests

**Changes**

Hello!

This PR creates a new ModelAppends rule that validates model appends are both:
- valid properties
- computed properties

As noted in the issue, appending an Accessor (attribute that modifies a database field) causes a very subtle bug where the value in the model array/json is always null.

Thanks!

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -400,6 +400,11 @@ services:
         class: Larastan\Larastan\Rules\ModelProperties\ModelPropertyRule
 
     -
+        class: Larastan\Larastan\Rules\ModelAppendsRule
+        tags:
+            - phpstan.rules.rule
+
+    -
         class: Larastan\Larastan\Rules\ModelProperties\ModelPropertyStaticCallRule
 
     -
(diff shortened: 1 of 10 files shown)


calebdw:
Note, I could remove the existence check if using `model-property` will check existence instead:

```php
    /** @var array<int, model-property<self>> */
    protected $appends = [
        'computed_property',
        'non_existent',
        'email', // This is a database column, should not be appended
        'name', // This is an accessor, should not be appended
    ];
```

calebdw:
So it turns out that the `model-property` type can only be used on method calls and not property definitions...

canvural:
Hi,

The rule makes sense. Nice catch with the bug!

Thanks.

canvural:
Hi,

I made this rule optional for now. 4834d0f 

I got many complaints about it. Again, we can enable it by default for v3. Also note for the future, we should not add any new rules in a minor/patch version.

PR #1839, opened 2024-02-11
fix: infinite mixin recursion

### LINKED ISSUE #1830: Infinite recursion until phpstan runs out of memory
- Larastan Version: 2.8.1
- `--level` used: 1
- Pull request with failing test: -

### Description

I ran into an instance where running phpstan (with larastan/larastan) installed took long to execute and then always failing, because it was running out of memory.

I have dug down and reproduced the problem in this repo: https://github.com/emil-nasso/larastan-recursion-reproduction

The problem can be reproduced by running: `./vendor/bin/phpstan analyze --memory-limit=512M app/Http/Controllers/TestController.php`

### The reproduction

We have a very simple controller that accesses a property of a model:

```php
class TestController extends Controller
{
    public function __invoke(): string
    {
        $model = MyModel::make();
        return $model->testAttribute;
    }
}
```

The model is very simple

```php
class MyModel extends BaseModel
{
}
```

but it does extend a `BaseModel`

```php
/**
 * @mixin MyModel
 */
abstract class BaseModel extends Model
{
}

```

that uses a mixing annotation.

### What happens?

When checking the controller, phpstan tries to check if `testAttribute` exists on `MyModel` using the `ModelPropertyExtension` where it uses the `ReflectionHelper` to fetch properties: https://github.com/larastan/larastan/blob/2.x/src/Properties/ModelPropertyExtension.php#L58

The `ReflectionHelper` then uses the `MixinPropertiesClassReflectionExtension` here: https://github.com/larastan/larastan/blob/2.x/src/Reflection/ReflectionHelper.php#L30

In `MixinPropertiesClassReflectionExtension` it looks at `MyModel` again, because of the `@mixin` annotation, and we have a loop that will go on for ever.

### How can this be fixed?

When `MixinPropertiesClassReflectionExtension` is instantiated ([here](https://github.com/larastan/larastan/blob/2.x/src/Reflection/ReflectionHelper.php#L30)), a list of classes to be ignored can be passed to the constructor like this:

```diff
- return (new MixinPropertiesClassReflectionExtension([]))->hasProperty($classReflection, $propertyName); // @phpstan-ignore-line
+ return (new MixinPropertiesClassReflectionExtension([$classReflection->getName()]))->hasProperty($classReflection, $propertyName); // @phpstan-ignore-line
```

This tells the class to lookup the property, based on the mixin but tells it "by the way, I'm class X, I already know I don't have the property, so ignore any mixings mentioning me".

There might also be an opportunity to fix this directly in [MixinPropertiesClassReflectionExtension](https://github.com/phpstan/phpstan-src/blob/1.11.x/src/Reflection/Mixin/MixinPropertiesClassReflectionExtension.php) but I'm not 100% certain on how, of that would have any side effects.

Let me know if you want me to create a PR with tests for the fix in this package and I can see what I can do there.

calebdw:
Closes #1830

Thanks!

### DIFF
--- a/tests/Type/GeneralTypeTest.php
+++ b/tests/Type/GeneralTypeTest.php
@@ -13,67 +13,67 @@ class GeneralTypeTest extends TypeInferenceTestCase
      */
     public static function dataFileAsserts(): iterable
     {
-        yield from self::gatherAssertTypes(__DIR__.'/data/request-object.php');
-        yield from self::gatherAssertTypes(__DIR__.'/data/eloquent-builder.php');
-        yield from self::gatherAssertTypes(__DIR__.'/data/query-builder.php');
-        yield from self::gatherAssertTypes(__DIR__.'/data/paginator-extension.php');
-        yield from self::gatherAssertTypes(__DIR__.'/data/model-properties.php');
-        yield from self::gatherAssertTypes(__DIR__.'/data/model-properties-relations.php');
-        yield from self::gatherAssertTypes(__DIR__.'/data/route.php');
-        yield from self::gatherAssertTypes(__DIR__.'/data/conditionable.php');
-        yield from self::gatherAssertTypes(__DIR__.'/data/tappable.php');
-        yield from self::gatherAssertTypes(__DIR__.'/data/translate.php');
-        yield from self::gatherAssertTypes(__DIR__.'/data/model-factories.php');
-        yield from self::gatherAssertTypes(__DIR__.'/data/environment-helper.php');
-        yield from self::gatherAssertTypes(__DIR__.'/data/view.php');
-        yield from self::gatherAssertTypes(__DIR__.'/data/bug-1346.php');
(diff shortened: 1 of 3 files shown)


tests/Type/GeneralTypeTest.php
canvural, line 1: Is it re-arranged alphabetically? I don't think we need this. This should not be a concern to anyone. And right now it makes the diff unreadble.

calebdw: Yes, I executed `:sort` on the entries and added the `bug-1830` file. I think it's better for a list of items to be sorted instead of unsorted: a) it's easier to know where to put a new file, and b) it's easier to find entries in a sorted list.

I split the sorting into its own commit so the changes are easier to follow. However, I can pop that commit off if you feel that strongly in favor of an unsorted list.

Alternatively, we could restructure the test so it automatically grabs all the files in the directory so we don't have to manually list each one.

canvural:
Thank you!

PR #1702, opened 2023-07-30
Add stub for Str::replace

besanek:
- [ ] Added or updated tests
- [ ] Documented user facing changes

**Changes**

In Laravel 10.16 the signature of the `Str::replace` method has changed, where the return type is now `string|string[]` instead of `string`. This MR will improve this behavior, according to the $subject parameter sets a specific return type.

**Breaking changes**

No breaking changes.

PR #1840, opened 2024-02-11
Add stub for Str::replace

besanek:
- [x] Added or updated tests

Reopen of #1702 with rebase to 2.x branch, Traversable support and added tests.

**Changes**

In Laravel 10.16 the signature of the Str::replace method has changed, where the return type is now string|string[] instead of string. This MR will improve this behavior, according to the $subject parameter sets a specific return type.

**Breaking changes**

No breaking changes.

### DIFF
--- a/stubs/common/Support/Str.stub
+++ b/stubs/common/Support/Str.stub
@@ -1,0 +1,19 @@
+<?php
+
+namespace Illuminate\Support;
+
+use Traversable;
+
+class Str
+{
+    /**
+     * @template TKey of scalar
+     * @template TSubject of string|iterable<string>|Traversable<TKey,string>
+     * @param string|iterable<string> $search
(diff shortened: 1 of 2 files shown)


canvural:
Hi,

We still support Laravel 9.x So you need to put your stub into `10.0.0` folder. And for the test, you need to conditionally include it. There are already existing examples you can look at.

besanek:
Well, it's a little tricky. Laravel 9 also returns the same types as Laravel 10, but has incorrect annotation. So maybe we should leave it in common to fix the return type there as well?

szepeviktor:
... or have two copies

canvural:
Aah ok. It's a bug fix. I thought returning an array was added new in 10.16

So, this is fine 👍🏽  Thank you.

PR #1841, opened 2024-02-11
fix: HasOneThrough relation being recognized as a collection

calebdw:
Fixes (and adds a test for) the HasOneThrough regression caused by #1835. I also decided to readd the string checks at the end in case there are third party extensions that extend Relation for some reason.

Thanks!


canvural:
Thank you!

PR #1843, opened 2024-02-12
feat: infer nullable type information by filled/blank function calls

### LINKED ISSUE #1290: blank helper should track nulls from variable
- Larastan Version:v2.1.11
- `--level` used: 8

### Description
blank helper should convert nullable to non-nullable variable 

```php
 protected function parsePhonesToFilemaker(string $phoneType, ?string $number): array
    {
        if (blank($number)) {
            return [];
        }

        $phoneNumberUtil = PhoneNumberUtil::getInstance();
        $phoneNumberObject = $phoneNumberUtil->parse($number);

        return [
            $phoneType . '_country_code' => $phoneNumberObject->getCountryCode(),
            $phoneType . ''              => $phoneNumberObject->getNationalNumber(),
            $phoneType . '_type'         => Str::of($phoneType)->ucfirst()->toString(),
            $phoneType . '_primary'      => 'Primary',
        ];
    }
```

Error:
`Parameter #1 $numberToParse of method libphonenumber\PhoneNumberUtil::parse() expects  string, string|null given.`

tpetry:
The result value of the `blank` and `filled` helpers imply some type information that is currently not supported by Larastan. To be precise, the `blank` helper will return `true` on null while the `filled` will return `false`. This enhanced type inference is especially usefull when working with nullable Eloquent attributes:

```php
if (filled($model->value)) {
  executeSomething($model->value);
}
```

Currently, PHPStan will report errors as `null` can't be passed to `executeSomething`. But the value can never be null because of a `true` result of the `filled` check. The new stubs help PHPStan to narrow the types of values passed to `filled` and `blank` functions.

Resolves #1290 

**Changes**

Added `blank` and `filled` helpers to the helper stubs

**Breaking changes**

None

### DIFF
--- a/stubs/common/Helpers.stub
+++ b/stubs/common/Helpers.stub
@@ -169,3 +169,23 @@ function optional($value = null, callable $callback = null) {}
  * @return ($value is empty ? TDefault : TReturn)
  */
 function transform($value, callable $callback, $default = null): mixed {}
+
+/**
+ * Determine if a value is "filled".
+ *
+ * @phpstan-assert-if-true !=null $value
+ *
+ * @param  mixed  $value
+ * @return bool
+ */
+function filled($value) {}
+
+/**
+ * Determine if the given value is "blank".
+ *
+ * @phpstan-assert-if-false !=null $value
+ *
+ * @param  mixed  $value
+ * @return bool
+ */
+function blank($value) {}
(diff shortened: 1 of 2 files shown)


canvural:
Thank you! Finally we can merge this 😄 Hope to make a release this week 👍🏽

PR #1844, opened 2024-02-12
chore: add tests for SupportCollection firstWhere

### LINKED ISSUE #1832: Support firstWhere on a collection
- Larastan Version: 2.8.1
- `--level` used: 5

### Description
Actually, LaraStan doesn't support the method `firstWhere()` on a collection, it return this kind of error:
> Match arm comparison between array|null and 'nps' is always false.

```php
$kpi = $options->firstWhere('name', '=', 'kpi')?->value; // kpi is not an array.

$benchmarking = match ($kpi) {
    'nps' => new Nps(), // Match arm comparison between array|null and 'nps' is always false.
    ...
};
```

calebdw:
**Changes**

This adds tests for the SupportCollection `firstWhere` method and subsequently closes #1832 as the method is working as expected.

Thanks!

### DIFF
--- a/tests/Type/data/collection-stubs.php
+++ b/tests/Type/data/collection-stubs.php
@@ -74,6 +74,9 @@ assertType('App\User|bool', $collection->first(function (User $user) {
 assertType('App\User|null', $collection->firstWhere('blocked'));
 assertType('App\User|null', $collection->firstWhere('blocked', true));
 assertType('App\User|null', $collection->firstWhere('blocked', '=', true));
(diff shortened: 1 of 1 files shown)


canvural:
We already have the tests for it. Just above where you added. Plus we don't have any special code for `firstWhere` So there is nothing to test actually 😄 

Thank you though.

calebdw:
Well, technically those were tests on the `EloquentCollection` so I added these for the `SupportCollection` just to be thorough.

canvural:
Ah, didn't catch that. In any case Laravel already has it's own tests for that. So we should be good 👍🏽

PR #1845, opened 2024-02-12
fix: model `first` not checking model properties

### LINKED ISSUE #1824: `ModelPropertyRule` rule do not support methods such as: select(), find(), first()
- Larastan Version: 2.8.1
- `--level` used: 5

### Description

I have observed that the `ModelPropertyRule` rule does not support methods that involve selecting columns, such as `select()`, `find()`, and `first()`.

### Example

```php
MyModel::query()
    ->where('xxx1')
    ->first(['xxx2', 'xxx3']);
```

In this example, an error will be triggered for `->where('xxx1')`, but not for `->first(['xxx2', 'xxx3'])`.

Is this behavior a bug, or does the `ModelPropertyRule` rule intentionally not support these methods?

calebdw:
- [x] Added or updated tests

**Changes**
Closes #1824 

This corrects the eloquent builder stubs for the `first` method.

Also, I noticed that the `select` method wasn't working---it looks like this is only defined on the QueryBuilder which doesn't have the model generics

Thanks!

### DIFF
--- a/stubs/10.0.0/EloquentBuilder.stub
+++ b/stubs/10.0.0/EloquentBuilder.stub
@@ -101,7 +101,7 @@ class Builder
     /**
      * Execute the query and get the first result.
      *
-     * @param  array<model-property<TModelClass>|int, mixed>|string  $columns
+     * @param  array<int, (model-property<TModelClass>|'*')>|model-property<TModelClass>|'*'  $columns
      * @return TModelClass|null
      */
     public function first($columns = ['*']);
--- a/stubs/common/EloquentBuilder.stub
+++ b/stubs/common/EloquentBuilder.stub
@@ -101,7 +101,7 @@ class Builder
     /**
      * Execute the query and get the first result.
      *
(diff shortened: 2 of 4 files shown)


canvural:
Hi,

We can actually just keep [`d948f42` (#1845)](https://github.com/larastan/larastan/pull/1845/commits/d948f42634f54e4596beb9ea2e1eedfc0714980a) the other commits are not needed.

> Also, I noticed that the select method wasn't working---it looks like this is only defined on the QueryBuilder which doesn't have the model generics

Yeah, so we cannot fix it with the `model-property` type. It'd need another extension (but debatable how useful it is 🤷🏽 )

Also please don't include any code style fixes in the PRs. I appreciate it, but CS can be handled by tools and not anything to be concerned manually. I'll set it in the CI this week 👍🏽

calebdw:
done

canvural:
Thank you!

clementbirkle:
Thank you @calebdw for this PR.
This PR should fix "select" method too: https://github.com/larastan/larastan/pull/1854

PR #1846, opened 2024-02-12
chore: add test for nullable collection cast

### LINKED ISSUE #1820: Attributes cast as Collections incorrectly interpreted as non-nullable
- Larastan Version: 1.10.56
- `--level` used: 5

### Description

I'm probably doing something wrong here, but I've searched the repo and can't find anything...

I have an Eloquent model which has a fillable property called `input_values`. This is set as a `collection` in `$casts = [...]`. At the database-level, this is stored in a nullable JSON column.

There are no docblocks indicating this is guaranteed as a `Collection`, and even so, I have `treatPhpDocTypesAsCertain: false` in my config. But Larastan reports:

```
------ ------------------------------------------------------------------------------------------------
  Line   Example/ExampleModel.php
 ------ ------------------------------------------------------------------------------------------------
  251    Using nullsafe method call on non-nullable type Illuminate\Support\Collection. Use -> instead.
         ✏️  Example/ExampleModel.php
 ------ ------------------------------------------------------------------------------------------------
```

The affected line is shown below.

```php
$fallbackValue = $this->input_values?->get($field->id, null) ?? $field->default;
```

I believe this is a false-positive, because `input_values` is only cast to a `Collection` when a non-null value is returned from the DB (as per Laravel docs).

calebdw:
**Changes**

This adds a test for the nullable collection cast and subsequently closes #1820 as it's working as expected.

Thanks!

### DIFF
--- a/tests/Type/data/model-properties.php
+++ b/tests/Type/data/model-properties.php
@@ -42,6 +42,7 @@ function foo(User $user, Account $account, Role $role, Group $group, Team $team,
     assertType('array', $user->array);
     assertType('array', $user->json);
     assertType(Collection::class, $user->collection);
+    assertType(Collection::class.'|null', $user->nullable_collection);
     assertType(Carbon::class, $user->date);
     assertType(Carbon::class, $user->datetime);
     assertType(CarbonImmutable::class, $user->immutable_date);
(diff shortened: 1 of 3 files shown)


szepeviktor (approved):

canvural:
Thank you!

PR #1848, opened 2024-02-13
feat: return all auth models when no guard is specified

calebdw:
- [x] Added or updated tests

**Changes**

Hello!

This PR revives #1232 by @mirucon which was closed when the `master` branch was deleted.

The motivation for this PR is that the "default" guard in a Laravel application is **not** constant but is rather used by Laravel to store the currently *chosen* guard.

Calling `Auth::shouldUse` / `Auth::setDefaultDriver` overwrites the default given in the config file:
https://github.com/laravel/framework/blob/0c574d42e9c7d82a0bd304a5cb2ce031a914bfa6/src/Illuminate/Auth/AuthManager.php#L201-L225

For example, I work on a large codebase with multiple guards and we use a middleware to set the "default" guard depending on the route. The result is that calling `Auth::user()` in different locations in the codebase results in *a different auth model returned*.

As this config change occurs at runtime, the best thing to do is to just return all possible auth models when there is no guard specified.

#### Note

As with the original author, I'm not too familiar with how the `Pipes\Auths` is used. Currently, I break after the first found model but perhaps I should continue to loop through all of the models? There don't seem to be any tests for the pipe classes.

**Breaking changes**

There might be a slight BC at level 7 when you can't call method on union types that don't exist on both types.

Thanks!

### DIFF
--- a/src/Concerns/LoadsAuthModel.php
+++ b/src/Concerns/LoadsAuthModel.php
@@ -8,17 +8,31 @@ use Illuminate\Contracts\Config\Repository as ConfigRepository;
 
 trait LoadsAuthModel
 {
-    /** @phpstan-return class-string|null */
-    private function getAuthModel(ConfigRepository $config, ?string $guard = null): ?string
+    /** @phpstan-return list<class-string> */
+    private function getAuthModels(ConfigRepository $config, ?string $guard = null): array
     {
-        if (
-            ($guard === null && ! ($guard = $config->get('auth.defaults.guard'))) ||
-            ! ($provider = $config->get('auth.guards.'.$guard.'.provider')) ||
-            ! ($authModel = $config->get('auth.providers.'.$provider.'.model'))
-        ) {
-            return null;
+        $guards = $config->get('auth.guards');
+        $providers = $config->get('auth.providers');
+
+        if (! is_array($guards) || ! is_array($providers)) {
+            return [];
         }
 
-        return $authModel;
+        return array_reduce(
+            is_null($guard) ? array_keys($guards) : [$guard],
+            function ($carry, $guardName) use ($guards, $providers) {
+                $provider = $guards[$guardName]['provider'] ?? null;
+                $authModel = $providers[$provider]['model'] ?? null;
+
+                if (! $authModel || in_array($authModel, $carry, strict: true)) {
+                    return $carry;
+                }
+
+                $carry[] = $authModel;
+
+                return $carry;
+            },
+            initial: [],
+        );
     }
 }
--- a/src/Methods/Pipes/Auths.php
+++ b/src/Methods/Pipes/Auths.php
@@ -44,10 +44,12 @@ final class Auths implements PipeContract
         $config = $this->resolve('config');
 
         if ($config !== null && in_array($classReflectionName, $this->classes, true)) {
-            $authModel = $this->getAuthModel($config);
+            $authModels = $this->getAuthModels($config);
 
-            if ($authModel !== null) {
-                $found = $passable->sendToPipeline($authModel);
+            foreach ($authModels as $authModel) {
+                if ($found = $passable->sendToPipeline($authModel)) {
+                    break;
+                }
             }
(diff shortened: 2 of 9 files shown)


canvural:
Thank you!

lloricode:
Hi @calebdw!

In my project got almost 32+ errors

```
 Call to method can() on an unknown class App\Models\User.                                                    
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols       
```

I'm trying to find new solution, but no documentation provided
can you help me provide how to apply fixes without touching production code?

thanks in advance

calebdw:
I take it the App\Models\User does not exist (should it?)?

It sounds like you might have a guard configured that you're not using and references an unknown model (so just clean up the config).

Some examples of the code and configuration would be helpful. What laravel version are you on?

lloricode:
> I take it the App\Models\User does not exist (should it?)?
> 
> It sounds like you might have a guard configured that you're not using and references an unknown model (so just clean up the config).
> 
> Some examples of the code and configuration would be helpful. What laravel version are you on?

since last version of laravel 10, currently I'm on laravel 11, and yes we use Filament v3 for handling auth guard, so I'm not sure how to fix this in larastan

calebdw:
Sounds like it might be an issue with how laravel 11 is merging the configuration.

What do you get when you dump the guards from the configs?

lloricode:
Screenshot from 2024-03-13 10-07-03

lloricode:
> Screenshot from 2024-03-13 10-07-03

we use `admin` for filament and `api` for api/customers, other is loaded by laravel

calebdw:
I need to see what all the values in the guards array is (`auth.guards`).

In addition, I need to see what is actually in your config file

I suspect that the `web` guard points to the App\Models\User model which does not exist in your codebase

lloricode:
Screenshot from 2024-03-13 10-40-49
Screenshot from 2024-03-13 10-40-06

lloricode:
@calebdw after this new config
Screenshot from 2024-03-13 10-59-06

all is working with no errors, but can we override this in larastan level?

lloricode:
can we set up different guard on `phpstan.neon`? hope this possible, even for laravel 10

lloricode:
Screenshot from 2024-03-13 11-09-40
 I update `web=>null` to make sure it can not use on production code, it work on this update, but seems tricky or dangerous also, my only goal is this can not use on impleementation

calebdw:
No you cannot setup a different guard in the `phpstan.neon`, Larastan just pulls from your configured guards---you really shouldn't have guards/providers configured that you're not using.

This primarily seems to be an issue with how Laravel is merging the configs together on Laravel 11---you should raise an issue there. If you don't specify a `web` guard then it shouldn't be in the resulting config.

I was just going to suggest setting `'web' => null` to override the framework, there's nothing wrong with that if you're not using that guard (it's not tricky or dangerous).

Alternatively, you could just tell phpstan to [ignore that particular error](https://phpstan.org/user-guide/ignoring-errors#ignoring-in-configuration-file), or you could just copy the `App\Models\User` model from Laravel into your application even if you don't use it.

lloricode:
> No you cannot setup a different guard in the `phpstan.neon`, Larastan just pulls from your configured guards---you really shouldn't have guards/providers configured that you're not using.
> 
> This primarily seems to be an issue with how Laravel is merging the configs together on Laravel 11---you should raise an issue there. If you don't specify a `web` guard then it shouldn't be in the resulting config.
> 
> I was just going to suggest setting `'web' => null` to override the framework, there's nothing wrong with that if you're not using that guard (it's not tricky or dangerous).
> 
> Alternatively, you could just tell phpstan to [ignore that particular error](https://phpstan.org/user-guide/ignoring-errors#ignoring-in-configuration-file), or you could just copy the `App\Models\User` model from Laravel into your application even if you don't use it.

it seems ok to me using `'web' => null` to make obvious in the future, and will raise this on laravel, thanks for the clarification on what going on, 

and again thanks for the improvement of larastan, it make our project better :100:

PR #1851, opened 2024-02-13
chore: add tests for LazyCollection flatmap

### LINKED ISSUE #1704: LazyCollection flatMap returns wrong type
- Larastan Version: 2.6.4
- `--level` used: 3
- Pull request with failing test: [this repository](https://github.com/JosephSilber/larastan-regression/blob/c755e37db3596f2884390c0fd408337f077a5d77/app/Http/Controllers/Controller.php#L14-L19)

### Description

When using the `flatMap` method on a `LazyCollection`, it doesn't pick up the static type from `EnumeratesValues`. It returns `Illuminate\Support\Traits\EnumeratesValues<int, mixed>` instead of a `LazyCollection`.

image

```php
use Illuminate\Support\LazyCollection;

function test(): LazyCollection
{
    return LazyCollection::times(5)->flatMap(function ($number) {
        return [$number];
    });
}
```

calebdw:
Hello!

Closes #1704 as it's working as expected.

Thanks!

### DIFF
--- a/tests/Type/data/collection-generic-static-methods.php
+++ b/tests/Type/data/collection-generic-static-methods.php
@@ -5,6 +5,7 @@ use App\TransactionCollection;
 use App\User;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Illuminate\Support\Collection as SupportCollection;
+use Illuminate\Support\LazyCollection;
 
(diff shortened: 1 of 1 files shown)


canvural:
Thank you

szepeviktor:
We are moving so fast.

PR #1852, opened 2024-02-13
fix: LazyCollection generics

### LINKED ISSUE #1704: LazyCollection flatMap returns wrong type

calebdw:
So #1851 added a test, but I wasn't aware that the particular test file didn't run on Laravel > 9.48. Adding the same test to this file causes the test to fail with the error reported in #1704 (sorry!).

This updates / renames the Collection dynamic return type extensions to work with the Enumerable interface and support LazyCollections.

Closes #1704 (for reals)


calebdw:
@canvural,

I don't think the first test file actually gets tested because the minimum version is L9.52:
https://github.com/larastan/larastan/blob/52b51412b28755e2fb759e02de0e3107b741f599/tests/Type/CollectionDynamicReturnTypeExtensionsTest.php#L18-L22

Also, why do we have two similarly named files?
- tests/Type/CollectionDynamicReturnTypeExtensionsTest.php
- tests/Type/CollectionDynamicReturnTypeExtensionTest.php

canvural:
Good catch! I re-opened the issue.

I think we can drop the condition and just go with `collection-generic-static-methods-l948.php` since like you said we support minimum 9.52

> Also, why do we have two similarly named files?
>
>    tests/Type/CollectionDynamicReturnTypeExtensionsTest.php
>    tests/Type/CollectionDynamicReturnTypeExtensionTest.php

`CollectionDynamicReturnTypeExtensionTest` is testing the return type extensions we wrote. The other is for the adjustments we made with the stubs or the generic static shenanigans.

calebdw:
@canvural, all tests are passing---I essentially updated the `CollectionGenericStaticMethodDynamicMethodReturnTypeExtension` to look at the Enumerable class instead of the SupportCollection interface.


Not sure if this is the best way to solve this or not---there are methods on a Collection that are not on the Enumerable class

calebdw:
Also, there's a similar class that might need to be updated as well?

- CollectionGenericStaticMethodDynamicStaticMethodReturnTypeExtension

calebdw:
@canvural, are the dynamic return type extensions still required or can we achieve the same effect with conditional return types (which should already exist)?

image

calebdw:
marking as draft until I can take a better look at this next week

canvural:
Hi,

Tests are passing that's good 👍🏽 

Thanks! I'll test this further before release.

Issue #1523, opened 2023-01-18
`withWhereHas()` method is not understand


utsavsomaiya:
- Larastan Version: 2.4.0
- `--level` used: 8
- Pull request with failing test:

https://user-images.githubusercontent.com/96036522/213170360-216c05ef-40f4-4338-8cba-887adcb00db1.mp4

### Laravel code where the issue was found

When I use the with method the larastan understand. It will not give me an error

### Before

```php
$videos = Video::query()
      ->with('liveStreamVideo:id,video_id,status')
      ->whereHas('liveStreamVideo', function ($query): void {
          $query->where('status', LivestreamVideoStatus::Scheduled)
              ->whereNotNull('video_sdk_room_id')
              ->where('start_at', '<=', now());
      })
      ->get();

$videos->each(function ($video): void {
        $video->liveStreamVideo?->update([
            'status' => LivestreamVideoStatus::Active,
        ]);
});
```

### After

```php
$videos = Video::query()
        ->withWhereHas('liveStreamVideo', function ($query): void {
            $query->where('status', LivestreamVideoStatus::Scheduled)
                ->whereNotNull('video_sdk_room_id')
                ->where('start_at', '<=', now());
        })
        ->get();

  $videos->each(function ($video): void {
      $video->liveStreamVideo?->update([
          'status' => LivestreamVideoStatus::Active,
      ]);
  });
```

szepeviktor:
Hello! 👋🏻 

It seems `withWhereHas` is not supported.
It is very easy to add some code to Larastan in a PR :)

[off] I would use the "Before" version.

takigawaakinori:
@canvural 
This issue has been solved. Please close this if there are no problems.

PR #1853, opened 2024-02-14
chore: add `withWhereHas` method to supported Eloquent Builder methods

takigawaakinori:
…(#1523)

- [x] Added or updated tests

**Changes**

add `withWhereHas` method to supported Eloquent Builder methods.

**Breaking changes**


tests/Type/data/model.php
```diff
@@ -343,6 +343,16 @@ function testRelationMethods(): void
         //assertType('Illuminate\Database\Eloquent\Builder<App\User>', $query);
     });
 
+    User::withWhereHas('accounts', function (Builder $query) {
```
calebdw, line 353: I've found that it's a bad idea to type hint the query object as that causes phpstan to drop the generic information:

```suggestion
    User::withWhereHas('accounts', function ($query) {
```

takigawaakinori: Thank you. I will check it.

takigawaakinori: I understood that testRelationMethods is for asserting the type of $query. In fact, the test on line 457 gives PostBuilder type hint. So without type hints, $query would simply be mixed and tests would fail.
Instead, I suggest adding the following test:

```php
function testWithWhereHas()
{
    assertType('Illuminate\Database\Eloquent\Builder<App\User>', (new User)->withWhereHas('accounts', function ($query) {
        return $query->where('active', true);
    }));
}
```

This will assert that the return value is of type Illuminate\Database\Eloquent\Builder<App\User> for the return value of withWhereHas. And this test worked in my environment.

calebdw: I was talking about dropping the generic information inside the closure

calebdw: But perhaps that was a bug that has been fixed

takigawaakinori:
Workflow seems to have failed by accident.

szepeviktor:
> Workflow seems to have failed by accident.

Rerunning ...

tests/Type/data/model.php
takigawaakinori: Thank you for information. Did you mention this issue?
https://github.com/phpstan/phpstan/issues/2934

tests/Rules/RelationExistenceRuleTest.php
```diff
@@ -53,8 +53,8 @@ public function testRule(): void
                 12,
             ],
             [
-                'Relation \'foo\' is not found in App\Group model.',
-                14,
+                'Relation \'foo\' is not found in App\User model.',
```
canvural, line 56: I don't understand why these are changed to `User` It should be an error for the `Group` model.

tests/Type/data/model.php
canvural: This is exactly why we have the commented assert lines everywhere here. So this is fine here.

canvural (approved):

canvural:
Thank you!

PR #1854, opened 2024-02-14
fix: model select not checking model properties

clementbirkle:
This PR fix this issue with "select" method: https://github.com/larastan/larastan/issues/1824

### DIFF
--- a/stubs/10.0.0/QueryBuilder.stub
+++ b/stubs/10.0.0/QueryBuilder.stub
@@ -242,6 +242,14 @@ class Builder
     protected function setAggregate($function, $columns);
 
     /**
+     * Set the columns to be selected.
+     *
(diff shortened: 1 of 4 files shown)


calebdw:
I already tried this and it didn't work because the query builder doesn't know about the model generics (you haven't passed any model to `model-property`, like `model-property<static>`).

I suspect that if you added a test that should pass that it would fail:

```php
\App\User::query()->select('email');
```

clementbirkle:
I just tested with your example and the test passed.

You're probably right...

calebdw:
Interesting---if it work it works, but I'm not sure how :upside_down_face: 

Can probably also do the same for `addSelect`

clementbirkle:
I tested it in the real world, and yes, you are correct; it doesn't work. I'm canceling this PR.

PR #1856, opened 2024-02-15
Allow running laravel-test.sh locally

spawnia:
- [x] Added or updated tests

**Changes**

This allows running the whole-application validation step that is used in GitHub actions locally.
I changed the directory it uses to create the application in order to not reach out of the project directory.

**Breaking changes**

None.


tests/laravel-test.sh
```diff
@@ -2,31 +2,32 @@
...
 
 cat >phpstan.neon <<"EOF"
 includes:
-    - ./vendor/larastan/larastan/extension.neon
+    - vendor/larastan/larastan/extension.neon
 parameters:
     level: 5
     paths:
-        - app/
+        - app
```
szepeviktor: This a very good PR!
The slash here signifies that app is a directory. Please revert it.

szepeviktor:
```suggestion
        - app/
```

szepeviktor (approved):

canvural:
Hi,

What is the purpose of running this check locally? I believe 99% of the cases unit tests and static analysis in enough to run locally. If you want to get more feedback you can open a PR and then even the E2E tests can run.

spawnia:
The purpose is the same as running any other test locally: making sure that changes made to the code are correct. I don't really see a downside to enabling it.

calebdw:
@canvural, I would agree that it is useful to be able to run e2e tests locally without having to keep pushing up to Github---I'm trying to debug an issue with monica right now on one of my PRs and being able to run this script locally would be very helpful.

canvural:
I don't think this is useful to the average contributor. Tests are the most important thing when it comes to checking if a PR works correctly or not. `laravel-test.sh` or E2E tests are there to just check if we accidentally break anything. Which we can see in the PR itself.

Thanks for understanding.

calebdw:
@canvural, this may not be useful to the average contributor---but this *is useful* when an e2e test does fail and you need to debug it. One of my PRs (https://github.com/larastan/larastan/pull/1879) is causing the Monica e2e test to fail and I have no idea why---I **need** to run this script locally to debug / test things out without having to keep pushing dummy changes up to github to test. My only option is to manually update the script locally which is a pain.

> laravel-test.sh or E2E tests are there to just check if we accidentally break anything. Which we can see in the PR itself.

This is kind of like saying that users should never run tests locally because they'll see if the tests break in the PR. Some times you don't need to run the tests locally, but we all know how invaluable it is to be able to locally run tests when working on / debugging something for the fast feedback loop---the same thing applies to the e2e tests.

I help maintain the [tree-sitter-php](https://github.com/tree-sitter/tree-sitter-php) grammar and we [have a similar e2e script](https://github.com/tree-sitter/tree-sitter-php/blob/master/script/parse-examples) that is used in ci. However, users can certainly execute the script locally if they need to.

I urge you to reconsider---there's no downside to this. On the contrary, it is generally bad for a script to reach outside the project directory. You'll notice in the tree-sitter-php script that everything is contained inside the project directory.

composer.json
```diff
@@ -75,9 +75,11 @@
         "test:cs": "phpcs",
         "test:types": "phpstan analyse --ansi --memory-limit 256M",
         "test:unit": "phpunit --colors=always -d memory_limit=1408M",
+        "test:application": "tests/laravel-test.sh",
         "test": [
             "@test:types",
-            "@test:unit"
+            "@test:unit",
+            "@test:application"
```
spawnia, line 82: Would you reconsider merging this pull request if I remove this line? That way, the end-to-end tests are not run by default but are possible to run if needed.

PR #1858, opened 2024-02-15
Improve type inference surrounding Arr::get and Arr::pull.

mad-briller:
**Changes**

Improves type inference surrounding `Arr::get` and `Arr::pull`, such as return types and type-specifying the `Arr::pull` `$array` argument. 

Also provides a rule to alert when the `$key` is not present on the `$array` for both `Arr::get` and `Arr::pull`, closing: https://github.com/larastan/larastan/issues/1727

### DIFF
--- a/src/Rules/ArrGetPullRule.php
+++ b/src/Rules/ArrGetPullRule.php
@@ -1,0 +1,85 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Larastan\Larastan\Rules;
+
+use Illuminate\Support\Arr;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\Type\VerbosityLevel;
+use PHPStan\Type\MixedType;
+use PhpParser\Node;
+use PhpParser\Node\Expr\StaticCall;
+
+use function in_array;
+
+/** @implements Rule<StaticCall> */
+class ArrGetPullRule implements Rule
+{
+    public function __construct(
+        private ReflectionProvider $reflectionProvider,
+    ) {
+    }
+
+    public function getNodeType(): string
+    {
+        return StaticCall::class;
+    }
+
+    public function processNode(Node $node, Scope $scope): array
+    {
+        if (! $node->name instanceof Node\Identifier) {
+            return [];
+        }
+
+        if (! $node->class instanceof Node\Name) {
+            return [];
+        }
+
+        $methodName = strtolower($node->name->name);
+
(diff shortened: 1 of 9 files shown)


tests/Type/GeneralTypeTest.php
```diff
@@ -73,6 +73,8 @@ public static function dataFileAsserts(): iterable
             yield from self::gatherAssertTypes(__DIR__.'/data/query-builder-l10-24.php');
         }
 
+        yield from self::gatherAssertTypes(__DIR__.'/data/arr-get.php');
```
calebdw: I would put this in the list above right after `application-make`---no need to list it after all the conditionals

mad-briller:
The static analysis actions failure is because i'm checking `$type->isObject()->yes()` before using `->isInstanceOf()` [as phpstan suggests](https://github.com/phpstan/phpstan-src/blob/1.11.x/src/Rules/Api/ApiInstanceofTypeRule.php#L73) instead of `instanceof ObjectType` but it doesn't seem to correctly infer that means it's an `ObjectType`?

szepeviktor (approved):

tests/Type/GeneralTypeTest.php
mad-briller: sure! also wanted to say great work on the prs recently mate

src/Rules/ArrGetPullRule.php
```diff
@@ -0,0 +1,85 @@
...
+    {
+        if (! $node->name instanceof Node\Identifier) {
+            return [];
+        }
+
+        if (! $node->class instanceof Node\Name) {
+            return [];
+        }
+
+        $methodName = $node->name->name;
```
canvural:
```suggestion
        $methodName = strtolower($node->name->name);
```

tests/Rules/data/arr-get-pull.php
```diff
@@ -0,0 +1,25 @@
...
+ * @param array<string, mixed> $mixedArray
+ * @param array<'foo', mixed> $specificKeys
+ */
+function test(array $mixedArray, array $specificKeys, array $unspecified, mixed $key): void
+{
+    Arr::get(['foo' => 1], 'bar');
+    Arr::get(['foo' => 1], 'foo');
+    Arr::pull(['foo' => 1], 'foo');
+    Arr::pull(['foo' => 1], 'bar');
+
```
canvural, line 17: Can we add a test with regular arrays? Like `Arr::get([1,2,3], 4)`

canvural: Also maybe with a array shape from PHPDocs.

canvural (changes requested):

canvural:

PR #1859, opened 2024-02-16
Add support for restoreOrCreate and createOrRestore SoftDelete macros.

### LINKED ISSUE #1819: SoftDeletes "createOrRestore" macro not recognised
- Larastan Version: 2.8.1
- `--level` used: 2

### Description

Calling the `createOrRestore` macro on an Eloquent Builder for a `MorphOne` relationship (whose child model implements `SoftDeletes`) throws errors for each model with the relationship. For example:

```
------ ------------------------------------------------------------------------------------------------------------------------
  Line   Models/Traits/HasIdentifier.php (in context of class App\Auth\User)
 ------ ------------------------------------------------------------------------------------------------------------------------
  92     Call to an undefined method Illuminate\Database\Eloquent\Relations\MorphOne<App\Models\Identifier>::createOrRestore().
         ✏️  Models/Traits/HasIdentifier.php
 ------ ------------------------------------------------------------------------------------------------------------------------
```

Other `SoftDeletes` macros like `withTrashed()` etc. work as expected.

Example `MorphOne` relationship, which correctly returns an `Identifier` model `withTrashed()`:

```php
/** The Identifier relationship for this Model. */
public function identifier(): MorphOne
{
    return $this->morphOne(Identifier::class, 'identifiable')->withTrashed();
}
```

Calling the `createOrRestore` macro on the identifier relationship works as expected in the codebase:

```php
$this->identifier()->createOrRestore([
    'identifiable_type' => $this->getMorphClass(),
    'identifiable_id'   => $this->getKey(),
]);
```

But throws the error above in Larastan.

mad-briller:
- [x] Added or updated tests

closes https://github.com/larastan/larastan/issues/1819

**Changes**
Added support for 2 missing macros on the SoftDeletes trait; `restoreOrCreate` and `createOrRestore`. `createOrRestore` was added in Laravel 10.44.0 whereas `restoreOrCreate` has existed since `9.x`.

### DIFF
--- a/src/Methods/EloquentBuilderForwardsCallsExtension.php
+++ b/src/Methods/EloquentBuilderForwardsCallsExtension.php
@@ -105,7 +105,7 @@ final class EloquentBuilderForwardsCallsExtension implements MethodsClassReflect
         if ($ref === null) {
             // Special case for `SoftDeletes` trait
             if (
-                in_array($methodName, ['withTrashed', 'onlyTrashed', 'withoutTrashed', 'restore'], true) &&
+                in_array($methodName, ['withTrashed', 'onlyTrashed', 'withoutTrashed', 'restore', 'createOrRestore', 'restoreOrCreate'], true) &&
                 array_key_exists(SoftDeletes::class, $modelReflection->getTraits(true))
             ) {
                 $ref = $this->reflectionProvider->getClass(SoftDeletes::class)->getMethod($methodName, new OutOfClassScope());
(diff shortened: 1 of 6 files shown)


canvural:
We need to find a better way to handle this soft deleted. It's causing problems for a long time 😅  But hey, if it works it works for now 😄 


Thank you!

mad-briller:
Short of reimplementing the macro logic in a method reflection extension i'm not sure what else could be done, it's unfortunate that this approach requires a lot of manual maintenance but this allows us to kick the can down the road just a little bit longer :sweat_smile:

PR #1860, opened 2024-02-19
Improve return type inference of Collection::groupBy.

### LINKED ISSUE #1813: Eloquent Collection `groupBy` has incorrect return type when grouping by multiple fields
- Larastan Version: 2.8.1
- `--level` used: 3

### Description

`Collection->groupBy` accepts a list of fields to group by. If more than one field is passed, it uses the higher-order `->map->groupBy` to recurse. This causes the return to change from being `static` to always being `Illuminate\Support\Collection`.

Also, there are extra levels of nesting for each additional field that is grouped on.

```php
use App\Models\User;

/** @return Collection<int|string, Collection<int|string, Collection<int, User>>> */
function example(): Collection {
  return User::all()->groupBy(['id', 'name']);
}
```

### Potential Fix

I've created this extension for my project, which appears to generate the correct return type for `groupBy` with multiple arguments.

```php
use PHPStan\Analyser\Scope;
use PHPStan\Reflection\MethodReflection;
use PHPStan\Type\Constant\ConstantIntegerType;
use PHPStan\Type\DynamicMethodReturnTypeExtension;
use PHPStan\Type\Generic\GenericObjectType;
use PHPStan\Type\IntegerRangeType;
use PHPStan\Type\IntegerType;
use PHPStan\Type\StringType;
use PHPStan\Type\Type;
use PHPStan\Type\UnionType;

class CollectionGroupByReturnExtension implements DynamicMethodReturnTypeExtension {
    public function getClass(): string {
        return Collection::class;
    }

    public function isMethodSupported(MethodReflection $methodReflection): bool {
        return $methodReflection->getName() === 'groupBy';
    }

    public function getTypeFromMethodCall(
        MethodReflection $methodReflection,
        MethodCall $methodCall,
        Scope $scope
    ): ?Type {
        $fields = $methodCall->getArgs();
        if (\count($fields) < 1) {
            return null;
        }

        $groupBy = $fields[0]->value;
        $groupByType = $scope->getType($groupBy);

        $collectionType = $scope->getType($methodCall->var);

        if (
            // @phpstan-ignore-next-line
            ! $collectionType instanceof GenericObjectType ||
            ! $groupByType->isIterable()->yes()
        ) {
            return null;
        }

        $limit = $groupByType->getArraySize();
        if ($limit instanceof ConstantIntegerType) {
            $limit = $limit->getValue();
        } elseif ($limit instanceof IntegerRangeType) {
            $limit = $limit->getMin();
        }

        if (! \is_int($limit) || $limit < 2) {
            return null;
        }

        $inner = new GenericObjectType(
            Collection::class,
            $collectionType->getTypes(),
        );

        while ($limit > 0) {
            $inner = new GenericObjectType(
                Collection::class,
                [
                    new UnionType([new IntegerType(), new StringType()]),
                    $inner,
                ],
            );

            $limit -= 1;
        }

        return $inner;
    }
}
```

mad-briller:
- [x] Added or updated tests

closes https://github.com/larastan/larastan/issues/1813

**Changes**

Adds a return type extension for the `Enumerable::groupBy()` method that provides more precise type information

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -290,6 +290,10 @@ services:
         tags:
             - phpstan.broker.dynamicMethodReturnTypeExtension
     -
+        class: Larastan\Larastan\ReturnTypes\EnumerableGroupByReturnTypeExtension
+        tags:
+            - phpstan.broker.dynamicMethodReturnTypeExtension
+    -
         class: Larastan\Larastan\ReturnTypes\CollectionWhereNotNullDynamicReturnTypeExtension
         tags:
             - phpstan.broker.dynamicMethodReturnTypeExtension
--- a/src/ReturnTypes/EnumerableGroupByReturnTypeExtension.php
+++ b/src/ReturnTypes/EnumerableGroupByReturnTypeExtension.php
@@ -1,0 +1,148 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Larastan\Larastan\ReturnTypes;
+
+use Illuminate\Support\Enumerable;
+use PHPStan\Analyser\OutOfClassScope;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Type\BenevolentUnionType;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\DynamicMethodReturnTypeExtension;
+use PHPStan\Type\Generic\GenericObjectType;
+use PHPStan\Type\IntegerType;
+use PHPStan\Type\MixedType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+use PhpParser\Node\Expr\MethodCall;
+
+use function array_reverse;
+use function count;
+
+class EnumerableGroupByReturnTypeExtension implements DynamicMethodReturnTypeExtension
+{
+    public function getClass(): string
+    {
+        return Enumerable::class;
+    }
+
+    public function isMethodSupported(MethodReflection $methodReflection): bool
+    {
+        return $methodReflection->getName() === 'groupBy';
+    }
+
+    public function getTypeFromMethodCall(
+        MethodReflection $methodReflection,
+        MethodCall $methodCall,
+        Scope $scope
+    ): ?Type {
+        if (count($methodCall->getArgs()) < 1) {
+            return null;
+        }
+
+        $calledOnType = $scope->getType($methodCall->var);
+        $objectClassReflections = $calledOnType->getObjectClassReflections();
+
+        if (!isset($objectClassReflections[0])) {
+            return null;
+        }
+
+        $groupByType = $scope->getType($methodCall->getArgs()[0]->value);
+        $propertyTypes = [];
+
+        $collectionName = $objectClassReflections[0]->getName();
+
+        $valueType = new MixedType();
+
+        if ($objectClassReflections[0]->isGeneric()) {
+            $tValueType = $methodReflection->getDeclaringClass()->getActiveTemplateTypeMap()->getType('TValue');
+
+            if ($tValueType !== null) {
+                $valueType = $tValueType;
+            }
+        }
+
+        if ($groupByType->isString()->yes()) {
+            $propertyTypes[] = $groupByType;
+        } elseif ($groupByType->isConstantArray()->yes()) {
+            if ($groupByType->isIterableAtLeastOnce()->no()) {
+                return $this->unknownKeysType($collectionName, $valueType);
+            }
+
+            $valuesArray = $groupByType->getConstantArrays()[0]->getValueTypes();
+
+            foreach ($valuesArray as $valuesType) {
+                $propertyTypes[] = $valuesType;
+            }
+        } else {
+            return $this->unknownKeysType($collectionName, new MixedType());
+        }
+
+
+        $innerKeyType = new IntegerType();
+
+        if (count($methodCall->getArgs()) >= 2) {
+            $preserveType = $scope->getType($methodCall->getArgs()[1]->value);
+            $tKeyType = $methodReflection->getDeclaringClass()->getActiveTemplateTypeMap()->getType('TKey');
+
+            if ($tKeyType === null) {
+                $innerKeyType = new BenevolentUnionType([new IntegerType(), new StringType()]);
+            } elseif ($preserveType->isTrue()->yes()) {
+                $innerKeyType = $tKeyType;
+            } elseif ($preserveType->isTrue()->maybe()) {
+                $innerKeyType = TypeCombinator::union($tKeyType, new IntegerType());
+            }
+        }
+
+
+        $inner = new GenericObjectType($collectionName, [$innerKeyType, $valueType]);
+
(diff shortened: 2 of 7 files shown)


tests/Type/data/collection-generic-static-methods-l948.php
```diff
@@ -44,8 +44,8 @@
 assertType('Illuminate\Support\Collection<App\User, int>', $collection->flip());
 assertType('Illuminate\Support\Collection<int, string>', $items->flip());
 
-assertType('Illuminate\Database\Eloquent\Collection<(int|string), Illuminate\Database\Eloquent\Collection<(int|string), App\User>>', $collection->groupBy('id'));
-assertType('Illuminate\Support\Collection<(int|string), Illuminate\Support\Collection<(int|string), int>>', $items->groupBy('id'));
+assertType('Illuminate\Database\Eloquent\Collection<int, Illuminate\Database\Eloquent\Collection<int, App\User>>', $collection->groupBy('id'));
+assertType("Illuminate\Support\Collection<'', Illuminate\Support\Collection<int, int>>", $items->groupBy('id'));
```
mad-briller, line 48: If it can't read a property off the types you have in your collection it helpfully groups them all into an empty string group...

tests/Type/data/collection-group-by.php
```diff
@@ -0,0 +1,74 @@
...
+        'Illuminate\Support\Collection<int, Illuminate\Support\Collection<string, Illuminate\Support\Collection<int|string, App\User>>>',
+        $stringUsers->groupBy(['id', 'name'], $preserve)
+    );
+    assertType(
+        "Illuminate\Support\Collection<'', Illuminate\Support\Collection<int, string>>",
+        $stringStrings->groupBy(['id'])
+    );
+    assertType(
+        'Illuminate\Support\Collection<(int|string), Illuminate\Support\Collection<(int|string), App\User>>',
+        $users->groupBy($keys)
```
mad-briller, line 73: if the `$keys` are unknown we defer to the phpdoc rather than trying to be smart, unfortunately leading to incorrect typing in many cases

the other option would be to return `Collection<(int|string), mixed>` but that also has downsides although is closer to the truth

szepeviktor:
One of the `getName()` calls fails.

mad-briller:
> One of the `getName()` calls fails.

yeah already on it :)

tests/Type/data/collection-generic-static-methods-l948.php
calebdw: I get why you do this, although not sure I really care for it...

extension.neon
```diff
@@ -289,6 +289,10 @@ services:
         class: Larastan\Larastan\ReturnTypes\CollectionFilterDynamicReturnTypeExtension
         tags:
             - phpstan.broker.dynamicMethodReturnTypeExtension
+    -
+        class: Larastan\Larastan\ReturnTypes\CollectionGroupByReturnTypeExtension
```
calebdw: Is there a way to fix this without having to introduce a new class? I don't think we really want to make a return type extension for every collection method...

Also https://github.com/larastan/larastan/pull/1852 moves the target class from `Collection` to `Enumerable` as this will not work for LazyCollections

tests/Type/data/collection-group-by.php
calebdw: I would rather have less specific types than incorrect types---I think in this case we would be able to return `Collection<(int|string), Collection<(int|string), mixed>>` as `groupBy` always returns a collection of collections right?

src/ReturnTypes/CollectionGroupByReturnTypeExtension.php
```diff
@@ -0,0 +1,128 @@
...
+use PHPStan\Type\MixedType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+use PhpParser\Node\Expr\MethodCall;
+
+use function array_reverse;
+use function count;
+
+class CollectionGroupByReturnTypeExtension implements DynamicMethodReturnTypeExtension
```
calebdw: Name of this should probably be `EnumerableGroupByReturnTypeExtension`

extension.neon
mad-briller: each collection method has it's own unique workings though which would cause a lot of complexity to manifest in a single place, rather than having a single extension for a single job.

personally i would prefer to see an extension for each method rather than trying to abstract everything into a single extension, it keeps things straightforwards and to the point

src/ReturnTypes/CollectionGroupByReturnTypeExtension.php
mad-briller: yeah good shout

tests/Type/data/collection-group-by.php
mad-briller: i'd have to check but yeah potentially, unless empty array causes it to return an empty top level collection? don't have anymore time for today will check tomorrow

tests/Type/data/collection-generic-static-methods-l948.php
mad-briller: i don't care for it either but it accurately models what laravel is doing with it's types.. i could make it be `new StringType()` but that would give a false sense of security aswell

extension.neon
calebdw: That's fair, if the logic is different. I guess I would be opposed to two different classes with the same logic

tests/Type/data/collection-generic-static-methods-l948.php
calebdw: If it can't read the type then why not just default to  `(int|string)` like the previous diff had?

mad-briller: because phpstan will report an error when trying to access any other key, alerting the developer to the mistake they have made:

https://phpstan.org/r/45d8567e-0831-4ad5-a35e-fcfe78f2d09c

tests/Type/data/collection-group-by.php
mad-briller: yeah even in the empty array case it returns that structure so i've refined that now, aswell as adding a test for the empty array case

src/ReturnTypes/EnumerableGroupByReturnTypeExtension.php
```diff
@@ -0,0 +1,148 @@
...
+                        $types[] = $valueType->getProperty($constantString->getValue(), new OutOfClassScope())->getReadableType();
+                    } elseif ($valueType->hasOffsetValueType($constantString)->yes()) {
+                        $types[] = $valueType->getOffsetValueType($constantString);
+                    }
+                }
+
+                if (count($types) === 0) {
+                    $keyType = new ConstantStringType('');
+                } else {
+                    $keyType = TypeCombinator::union(...$types);
```
canvural, line 117: Shouldn't we check if the types are either `int` or `string` ? What I mean is that for example objects can't be array keys. Or if any of the types were nullable we would have for example `int|null` as collection key. Which does not make sense.

Maybe we can call `toArrayKey` on this type?

canvural (changes requested):

canvural:

PR #1861, opened 2024-02-19
Fix not respecting return type of overridden Request::user() method.

### LINKED ISSUE #1718: Wrong return type when overwriting Request::user()
- Larastan Version: 2.6.4
- `--level` used: 8

### Description
Larastan will overwrite the return type of requests that narrow the return of `user()`.
```
  5     Cannot access property $id on App\User|null.
```

```php
<?php

namespace App\Http\Requests;

resolve(AuthedRequest::class)->user()->id
```

```php
<?php

namespace App\Http\Requests;

use Illuminate\Foundation\Http\FormRequest;
use App\User;

class AuthedRequest extends FormRequest
{
    public function authorize(): bool
    {
        return parent::user() instanceof User;
    }

    public function user($guard = null): User
    {
        $user = parent::user($guard);
        if (!$user instanceof User) {
            abort(403);
        }

        return $user;
    }
}
```

mad-briller:
- [x] Added or updated tests

closes https://github.com/larastan/larastan/issues/1718

**Changes**

Change RequestUserExtension to use the specific type of an overridden `Request::user()` method

### DIFF
--- a/src/ReturnTypes/RequestUserExtension.php
+++ b/src/ReturnTypes/RequestUserExtension.php
@@ -10,7 +10,6 @@ use Larastan\Larastan\Concerns\LoadsAuthModel;
 use PhpParser\Node\Expr\MethodCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\MethodReflection;
-use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
@@ -39,7 +38,11 @@ final class RequestUserExtension implements DynamicMethodReturnTypeExtension
         MethodReflection $methodReflection,
         MethodCall $methodCall,
         Scope $scope,
-    ): Type {
+    ): Type|null {
+        if ($methodReflection->getDeclaringClass()->getName() !== Request::class) {
+            return null;
+        }
+
         $config     = $this->getContainer()->get('config');
         $authModels = [];
 
@@ -49,7 +52,7 @@ final class RequestUserExtension implements DynamicMethodReturnTypeExtension
         }
 
         if (count($authModels) === 0) {
-            return ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+            return null;
         }
 
         return TypeCombinator::addNull(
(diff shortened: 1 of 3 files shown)


tests/Type/GeneralTypeTest.php
```diff
@@ -59,6 +59,7 @@ public static function dataFileAsserts(): iterable
         yield from self::gatherAssertTypes(__DIR__.'/data/view-exists.php');
         yield from self::gatherAssertTypes(__DIR__.'/data/view.php');
         yield from self::gatherAssertTypes(__DIR__.'/data/where-relation.php');
+        yield from self::gatherAssertTypes(__DIR__.'/data/bug-1718.php');
```
calebdw: I would put this with the other bugs

calebdw:
You beat me to this!

src/ReturnTypes/RequestUserExtension.php
```diff
@@ -48,6 +48,10 @@ public function getTypeFromMethodCall(
         MethodCall $methodCall,
         Scope $scope
     ): Type {
+        if ($methodReflection->getDeclaringClass()->getName() !== Request::class) {
+            return ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
```
canvural: Can't we return `null` here?

canvural (changes requested):
Looks good. Thanks.

Just left a one comment. Also the branch has merge conflicts now. That's why it is a good idea to put any test case in `GeneralTypeTest` to at the end of the file. The order of the tests do not matter at all 😊

calebdw:
> Also the branch has merge conflicts now. That's why it is a good idea to put any test case in GeneralTypeTest to at the end of the file. The order of the tests do not matter at all 😊

I get that order does not matter (other than for aesthetics), but you would still have merge conflicts if two separate PRs placed files in the same spot at the end of the file.

mad-briller:
i'll get to this proper when i have some time after work

HDVinnie:
Whats the status on this @mad-briller @calebdw @canvural

szepeviktor:
This PR needs to be fixed.

mad-briller:
not sure what's going on with CI, appears to be some stuff broken outside the scope of this pr, and that are failing on other recent prs too
but this pr should be ready now, sorry for the delay it slipped my mind

calebdw:
Awesome thanks! Laravel 11 tests are broken until #1990 is merged and the other tests are failing due to PHPstan 1.12 which will be fixed with #2028

canvural:
Thank you!

PR #1862, opened 2024-02-19
Fix UnusedViewsRule not providing an existing file path to RuleErrorBuilder

### LINKED ISSUE #1725: UnusedViewsRule does not pass an existing file path to RuleErrorBuilder
- Larastan Version: b49b6d3ec144c675fa86bbe453ef11d8c42c2016

### Description

The next version of PHPStan is going to require an existing file path to be passed to `RuleErrorBuilder::file()`. This prevents various issues with the baseline and result cache.

When I try this, Larastan test suite fails with:

```
1) Rules\UnusedViewsRuleTest::testRule
PHPStan\ShouldNotHappenException: File unused.blade.php does not exist.

phar:///Users/ondrej/Development/larastan/vendor/phpstan/phpstan/phpstan.phar/src/Rules/RuleErrorBuilder.php:56
/Users/ondrej/Development/larastan/src/Rules/UnusedViewsRule.php:75
phar:///Users/ondrej/Development/larastan/vendor/phpstan/phpstan/phpstan.phar/src/Testing/RuleTestCase.php:126
phar:///Users/ondrej/Development/larastan/vendor/phpstan/phpstan/phpstan.phar/src/Testing/RuleTestCase.php:86
/Users/ondrej/Development/larastan/tests/Rules/UnusedViewsRuleTest.php:50
```

That's because `$unusedViews` in UnusedViewsRule contains strings like `unused` and not paths to existing files.

mad-briller:
- [x] Added or updated tests
 
closes https://github.com/larastan/larastan/issues/1725

**Changes**

The `UnusedViewsRule` now uses the `ViewFinderInterface` to specify the full path to the unused view, rather than the short name. This is to resolve a constraint with a future version of PHPStan that will require all `RuleErrorBuilder->file()` calls to be absolute paths to existing files.

**Breaking changes**

It's likely that any past baselines that were generated including an `UnusedViewsRule` violation will need regenerating as the reported path will change. Not alot we can do about that though as it's a necessary change for a future version of phpstan.

### DIFF
--- a/src/Rules/UnusedViewsRule.php
+++ b/src/Rules/UnusedViewsRule.php
@@ -70,8 +70,10 @@ final class UnusedViewsRule implements Rule
 
         $errors = [];
         foreach ($unusedViews as $file) {
+            $path = $view->getFinder()->find($file);
+
             $errors[] = RuleErrorBuilder::message('This view is not used in the project.')
-                ->file($file.'.blade.php')
+                ->file($path)
                 ->line(0)
                 ->build();
         }
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,7 @@
 # Upgrade Guide
 
+- The UnusedViewsRule has been changed to specify the absolute path of the unused view, rather than the view name. This may mean that baselines will need regenerating to account for this change.
+
 ## Upgrading to `2.7.0` from `2.6.5`
 
 ### Organization Change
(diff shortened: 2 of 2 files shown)


src/Rules/UnusedViewsRule.php
```diff
@@ -5,6 +5,7 @@
 namespace Larastan\Larastan\Rules;
 
 use Illuminate\View\Factory;
+use Illuminate\View\ViewFinderInterface;
```
canvural: I don't see where this interface is used 🤔

mad-briller: my mistake, a holdover from an original approach before i learned that `->getFinder()` existed

canvural:
Thank you!

PR #1863, opened 2024-02-19
Fixed issue where collections with invalid templates configured would cause a fatal error.

### LINKED ISSUE #1565: Null given to ObjectType::isSuperTypeOf()
- Larastan Version: 2.4.1
- `--level` used: 8

### Description

The use of `random()` results in:
```
Internal error: PHPStan\Type\ObjectType::isSuperTypeOf(): Argument #1 ($type) must be of type PHPStan\Type\Type, null given, called in                                          
     /vendor/nunomaduro/larastan/src/ReturnTypes/CollectionGenericStaticMethodDynamicMethodReturnTypeExtension.php on line 116
```

```php
/** @var Brand $brand */
$brand = Brand::findOrFail(1);
$department = $brand->groups->random();
```
```php
/** @return BelongsToMany<UserGroup> */
public function groups(): BelongsToMany
{
  return $this->belongsToMany(
    UserGroup::class,
    BrandOperatorGroupMembership::class,
    'brand_id',
    'group_id'
  );
}
```

mad-briller:
- [x] Added or updated tests

closes https://github.com/larastan/larastan/issues/1565

**Changes**

Fixes a long standing issue where custom collections that don't specify `@extends` cause crashes when calling functions on them that are handled by `CollectionGenericStaticMethodDynamicMethodReturnTypeExtension`.

Comments in the report suggest adding a rule to enforce that collections always have `@extends` but this seems superfluous as phpstan already reports this at a sufficiently high level.

### DIFF
--- a/tests/Type/GeneralTypeTest.php
+++ b/tests/Type/GeneralTypeTest.php
@@ -20,6 +20,7 @@ class GeneralTypeTest extends TypeInferenceTestCase
         yield from self::gatherAssertTypes(__DIR__.'/data/auth.php');
         yield from self::gatherAssertTypes(__DIR__.'/data/benchmark.php');
         yield from self::gatherAssertTypes(__DIR__.'/data/bug-1346.php');
+        yield from self::gatherAssertTypes(__DIR__.'/data/bug-1565.php');
         yield from self::gatherAssertTypes(__DIR__.'/data/bug-1760.php');
         yield from self::gatherAssertTypes(__DIR__.'/data/bug-1830.php');
         yield from self::gatherAssertTypes(__DIR__.'/data/carbon.php');
--- a/tests/Type/data/bug-1565.php
+++ b/tests/Type/data/bug-1565.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Bug1565;
+
+use Illuminate\Support\Collection;
+
+use function PHPStan\Testing\assertType;
+
+class TestCollection extends Collection
+{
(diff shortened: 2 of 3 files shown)


szepeviktor (approved):

tests/Type/GeneralTypeTest.php
```diff
@@ -59,6 +59,7 @@ public static function dataFileAsserts(): iterable
         yield from self::gatherAssertTypes(__DIR__.'/data/view-exists.php');
         yield from self::gatherAssertTypes(__DIR__.'/data/view.php');
         yield from self::gatherAssertTypes(__DIR__.'/data/where-relation.php');
+        yield from self::gatherAssertTypes(__DIR__.'/data/bug-1565.php');
```
calebdw: I would move this up with the other bugs

tests/Type/data/bug-1565.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Bug1565;
+
+use Illuminate\Support\Collection;
+
+use function PHPStan\Testing\assertType;
+
+class TestCollection extends Collection
```
calebdw, line 9: I would probably add a comment that the generics are specifically excluded.

calebdw: eh, nevermind---I missed that this was defined directly in the test file

canvural:
Thank you!

PR #1864, opened 2024-02-25
fix: ci failing when testing outside working dir

calebdw:
Hello!

CI is currently failing when testing Laravel outside of the working directory. The tests fully pass when testing inside the Laravel application so this points to a bootstrapping and/or a path/stub resolution issue (as best as I can tell).

I don't really see the need / point of trying to test outside the working dir so just opted to remove the test altogether.

Thanks!

### DIFF
--- a/stubs/common/EloquentCollection.stub
+++ b/stubs/common/EloquentCollection.stub
@@ -25,25 +25,25 @@ class Collection extends BaseCollection implements QueueableCollection
     public function find($key, $default = null);
 
     /**
-         * Run a map over each of the items.
-         *
-         * @template TMapValue
-         *
-         * @param  callable(TModel, TKey): TMapValue  $callback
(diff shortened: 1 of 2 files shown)


szepeviktor:
> CI is currently failing when testing Laravel outside of the working directory.

Thank you for your report.
Could you show me an example?

calebdw:
https://github.com/larastan/larastan/actions/runs/8034035538/job/21945136111

szepeviktor:
This failure is caused by a420e73b2c40ebef27637a3b482e0ebd187e5178

canvural:
Hi,

The issue is not related to that. As @szepeviktor pointed out it is caused by the latest commit. I'm trying to fix it actually. But it's taking some time. Maybe I can revert the last commit in the meanwhile.

calebdw:
I understand which commit caused the failure. However, I believe the generics are correct and there's nothing in the commit that would be responsible for the error.

On top of that, the tests **do pass**, when phpstan is executed inside the project indicating that there's nothing wrong with the commit

szepeviktor:
This is good news!
Now we have something to investigate.

PR #1867, opened 2024-02-26
chore: setup codestyle checks

canvural:
WIP

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
         "phpstan/phpstan": "^1.10.50"
     },
     "require-dev": {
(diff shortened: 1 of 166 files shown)


composer.json
```diff
@@ -68,6 +72,7 @@
         }
     },
     "scripts": {
+        "test:cs": "phpcs",
```
calebdw, line 75: Why not use Laravel Pint or StyleCI (it looks like StyleCI was used in the past)?

canvural: They both use PHPCSFixer. Personally I like the PHP CodeSniffer more.

Issue #1870, opened 2024-03-07
Parent class not being accepted in place of template type, static, or this


calebdw:
- Larastan Version: 2.9.2
- `--level` used: 3

### Description
Hello!

Not sure if this is an issue with Larastan or PHPStan, or I could just be doing something wrong.

I have many errors in my codebase where the parent class is not accepted in place of a template type, static, or this---and these usually happen on the parent class where I'm trying to define the proper generics for the children. 

The errors follow the form:

```
should return TFoo of Foo but returns Foo

should return static(Foo) but returns Foo

should return $this(Foo) but returns Foo
```

### Laravel code where the issue was found

```php
class CacheableService
{
    public function __construct(
        protected Repository $cache,
    ) {
    }

    /**
     * @template TModel of Model
     * @param class-string<TModel> $model
     * @return EloquentCollection<int, TModel>
     */
    public function put(string $model): EloquentCollection
    {
        $records = $model::query()
            ->orderBy('id')
            ->get();

        $this->cache->forever($this->resolveKey($model), $records);

        // Method put() should return Collection<int, TModel of Model> but returns Collection<int, Model>.
        return $records;
    }
}
```

```php
/** @template TRecord of Model */
class Token extends Model
{
    /** @return BelongsTo<TRecord,static> */
    public function record(): BelongsTo
    {
        // no errors here though
        return $this->belongsTo(
            static::getTokenType()->record_type,
            'record_id',
        );
    }

    /** @return BelongsTo<TokenType,static> */
    public function type(): BelongsTo
    {
         // Method Token::type() should return BelongsTo<TokenType, static(Token<TRecord of Model>)> but returns BelongsTo<TokenType, Token>.
        return $this->belongsTo(TokenType::class, 'type_id');
    }

    /** @param TRecord $record */
    public static function generate(Model $record): static
    {
        $token = static::create(['record_id' => $record->getKey()]);

        return $token; // Method Token::generate() should return static(Token<TRecord of Model>) but returns Token.
    }
}
```

Note that the Token model fails even when the model is not generic so that's not the issue here:

> Method Token::generate() should return static(Token) but returns Token.

Granted, the Token class might could use some structural changes, but I'm just trying to work within the confines of the existing architecture (#legacy) while having types

CC: @ondrejmirtes 

Thanks!

canvural:
If you can replicate it on https://phpstan.org/try , that might indicate it is a PHPstan issue. No need to tag Ondrej every time 😊

[...]

PR #1871, opened 2024-03-08
fix: BelongsTo TChildModel generics

calebdw:
- [x] Added or updated tests

This is related to #1870 but does not fully solve it.

**Changes**

This PR fixes the relationship and model relation dynamic return type extensions as well as the relation forwarding extension to properly use `$this` as the BelongsTo TChildModel generic instead of creating a new object class.

This allows inheritance to work properly:

```php
class Baz extends Model
{
    /** @return BelongsTo<Foo, $this> */
    public function foo(): BelongsTo
    {
        return $this->belongsTo(Foo::class);
    }
}

class ChildBaz extends Baz
{
}

// used to be `BelongsTo<Foo, Baz>`!
assertType('BelongsTo<Foo, ChildBaz>', $childBaz->foo());
```

Also fixed type mismatch:

```php
class Baz extends Model
{
    /** @return BelongsTo<Foo, $this> */
    public function foo(): BelongsTo
    {
        // this failed phpstan because the relation method returned
        // `BelongsTo<Foo, Baz>` instead of `BelongsTo<Foo, $this(Baz)>`
        return $this->belongsTo(Foo::class);
    }
}
```

With these two changes, relation forwarding (inside the relation method) started failing with a nesting issue which I fixed while also simplifying the logic:

```php
class Baz extends Model
{
    /** @return BelongsTo<Foo, $this> */
    public function foo(): BelongsTo
    {
        // this was returning `BelongsTo<Foo, BelongsTo<Foo, $this(Baz)>>`
        return $this->belongsTo(Foo::class)->orderBy('id');
    }
}
```

**Breaking changes**

If users have specified the BelongsTo generics with a class or `static` they will need to update to `$this`  (depending on their phpstan level).:
```diff
- /** @return BelongsTo<Foo, Baz> */
- /** @return BelongsTo<Foo, static> */
+ /** @return BelongsTo<Foo, $this> */
```

Thanks!

### DIFF
--- a/src/Methods/RelationForwardsCallsExtension.php
+++ b/src/Methods/RelationForwardsCallsExtension.php
@@ -6,7 +6,6 @@ namespace Larastan\Larastan\Methods;
 
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
-use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use Larastan\Larastan\Reflection\EloquentBuilderMethodReflection;
 use PHPStan\Reflection\ClassReflection;
@@ -16,8 +15,8 @@ use PHPStan\Reflection\MissingMethodFromReflectionException;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\ShouldNotHappenException;
-use PHPStan\Type\Generic\GenericObjectType;
 use PHPStan\Type\ObjectType;
+use PHPStan\Type\ThisType;
 
 use function array_key_exists;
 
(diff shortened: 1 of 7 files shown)


canvural:
Hi,

I explained in the issue with more detail. Unfortunately this is not the right fix.

Thanks for understanding!

calebdw:
@canvural, this is a valid issue and is the correct way to handle BelongsTo generics. The BelongsTo relation *does return `$this`*:

https://github.com/laravel/framework/blob/6a8b5051553321a4c18c2eadf17d192c92f3e42e/src/Illuminate/Database/Eloquent/Relations/BelongsTo.php#L212

As I pointed out in the PR description, inheritance is currently broken:

```php
class Baz extends Model
{
    /** @return BelongsTo<Foo, $this> */
    public function foo(): BelongsTo
    {
        return $this->belongsTo(Foo::class);
    }
}

class ChildBaz extends Baz
{
}

// used to be `BelongsTo<Foo, Baz>`!
assertType('BelongsTo<Foo, ChildBaz>', $childBaz->foo());
```

This PR is required for my codebase to reach Level 3

canvural:
It might be a valid issue. But the fix is not right. We cannot force to return `ThisType`

calebdw:
Why not? `ThisType` is exactly what should be returned---the BelongsTo returns the `$this` that the relation was called on

canvural:
BelongsTo does not return `$this`. It returns the child object.

calebdw:
For this class:

```php
class Baz extends Model
{
    /** @return BelongsTo<Foo, $this> */
    public function foo(): BelongsTo
    {
        return $this->belongsTo(Foo::class);
    }
}
```

Calling a method **does return $baz**--- the instance that the relationship was defined on, not a new a new Baz or static instance.
```php
$baz = Baz::first();
// this returns $baz 
$baz->foo()->associate($foo)
```

In this case `$this` when defined on a model is the most correct type

canvural:
> the instance that the relationship was defined on, not a new a new Baz 

In the eyes of static analysis there is no difference there. 

I agree there might be an issue with the inheritance. But returning `ThisType` is not correct fix. We need to find other ways.

PR #1872, opened 2024-03-08
fix: float not assignable to decimal cast

### LINKED ISSUE #1869: Model property with decimal cast (numeric-string) does not accept float or int
- Larastan Version: 2.9.2
- `--level` used: 3

### Description
Hello!

First off, I'm not sure if this is a Larastan issue or a PHPStan issue, but here goes:

On PHPStan level 3, types assigned to properties are checked. In my work codebase, we use the `decimal` model cast somewhat occasionally (mainly on prices and other things that need a guaranteed precision). However, I'm now running into an error where I cannot assign any numerical value to the property:

> Property LineItem::$agreed_price (numeric-string|null) does not accept float.

Type `int` and literals (`0`, `0.`) are not accepted either.

The only way to fix this that I've found is to cast it to a string but that's not really something I'd like to do all over.

Thoughts?

```php
class LineItem extends Model
{
    protected $casts = ['agreed_price' => 'decimal:2'];
}
// ...

// error
$lineItem->agreed_price = $quantity * $discountPrice;
// works, but ugly
$lineItem->agreed_price = (string) $quantity * $discountPrice;
// works, but verbose and I'd rather not have to repeat the precision around the codebase
$lineItem->agreed_price = number_format($quantity * $discountPrice, 2);
```

P.S. I'm making the push to bring my work codebase up to PHPStan level 3...so there will likely be several issues as I run into things that aren't easily fixable.

CC: @ondrejmirtes 

Thanks!

calebdw:
- [x] Added or updated tests

Closes #1869

**Changes**
Hello!

This PR updates the Eloquent `decimal` writable type to accept `float`.

I also tried to cleanup the file a bit by using the `::class` constant instead of the entire raw class string, but feel free to pop that off if you don't like it.

Thanks!

### DIFF
--- a/src/Properties/ModelCastHelper.php
+++ b/src/Properties/ModelCastHelper.php
@@ -4,12 +4,22 @@ declare(strict_types=1);
 
 namespace Larastan\Larastan\Properties;
 
+use Carbon\Carbon;
+use Carbon\CarbonImmutable;
 use Illuminate\Contracts\Database\Eloquent\Castable;
 use Illuminate\Contracts\Database\Eloquent\CastsAttributes;
 use Illuminate\Contracts\Database\Eloquent\CastsInboundAttributes;
+use Illuminate\Database\Eloquent\Casts\ArrayObject;
+use Illuminate\Database\Eloquent\Casts\AsArrayObject;
+use Illuminate\Database\Eloquent\Casts\AsCollection;
+use Illuminate\Database\Eloquent\Casts\AsEncryptedArrayObject;
+use Illuminate\Database\Eloquent\Casts\AsEncryptedCollection;
+use Illuminate\Database\Eloquent\Casts\AsStringable;
 use Illuminate\Support\Arr;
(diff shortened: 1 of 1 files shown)


src/Properties/ModelCastHelper.php
```diff
@@ -100,7 +100,7 @@ public function getWriteableType(string $cast, Type $originalType): Type
         $attributeType = match ($cast) {
             'int', 'integer', 'timestamp' => new IntegerType(),
             'real', 'float', 'double' => new FloatType(),
-            'decimal' => TypeCombinator::intersect(new StringType(), new AccessoryNumericStringType()),
+            'decimal' => TypeCombinator::intersect(new StringType(), new AccessoryNumericStringType(), new FloatType()),
```
szepeviktor, line 115: This is the real change.

canvural:
Adding just the float is right here. Thanks!

PR #1874, opened 2024-03-09
fix: return $this from builder methods

### LINKED ISSUE #1538: Inconsistent return types on model scopes with certain whereX functions when using static
- Larastan Version: 2.4.0
- `--level` used: 6
- ~~Pull request with failing test~~ Laravel example with failing result: https://github.com/jnoordsij/laravel-laravel/tree/model-scope-examples

### Description
On my abstract base model, I have implemented a few scopes that are shared between models that have columns with identical definitions. However, where for some of these scopes I can list the return type to be `Builder<static>` without issues, some of them will yield an error with some weird kind of nested `Builder`, e.g.
```
Method Sandbox\Model::scopeValid() should return
Illuminate\Database\Eloquent\Builder<static(Sandbox\Model)>
but returns
Illuminate\Database\Eloquent\Builder<Illuminate\Database\Eloquent\Builder<Illuminate\Database\Eloquent\Builder<Illuminate\Database\Eloquent\Builder<static(Sandbox\Model)>>>>.
```

When I use `self` instead of `static` the problem disappears, but this causes problems elsewhere, as I can not chain it with specific scopes or where clauses on the models implementing the abstract base model.

This might be partially solved by changing this
https://github.com/nunomaduro/larastan/blob/8982d3142c4b0de94c326f8f56212663b3e25b61/stubs/EloquentBuilder.stub#L235
to `@return static` (this is done on most other methods in this stub as well; I'm not entirely sure for what purpose).

```php
<?php

namespace Sandbox;

use Illuminate\Database\Eloquent\Model as BaseModel;
use Illuminate\Support\Carbon;

abstract class Model extends BaseModel
{
    /**
     * Scope for valid entries.
     *
     * @param  Builder<static>  $query
     * @return Builder<static>
     */
    public function scopeValid(Builder $query): Builder
    {
        return $query->whereDate('ValidUntil', '>=', Carbon::today())
                     ->orWhereNull('ValidUntil');
    }

    /**
     * Regular where scope.
     *
     * @param  Builder<static>  $query
     * @return Builder<static>
     */
    public function scopeTest(Builder $query): Builder
    {
        return $query->where('Column', true);
    }

    /**
     * Or null where scope.
     *
     * @param  Builder<static>  $query
     * @return Builder<static>
     */
    public function scopeTestOrNull(Builder $query): Builder
    {
        return $query->test()->orWhereNull('Column');
    }

    /**
     * Or null where scope.
     *
     * @param  Builder<static>  $query
     * @return Builder<static>
     */
    public function scopeTestWhereNull(Builder $query): Builder
    {
        return $query->whereNull('Column');
    }
}
```

See also https://github.com/jnoordsij/laravel-laravel/tree/model-scope-examples (run `composer install && ./vendor/bin/phpstan`).

### LINKED ISSUE #1676: Method Builder::whereX() should return static(Builder<Model>) but returns Builder<Model>
- Larastan Version: 2.6.3
- `--level` used: 5
- Pull request with failing test: https://github.com/sanfair/larastan-playground/pull/1

### Description

When defining a new `whereX` method in custom builder with return type annotation phpstan shows errors.

In PR you can find both possible definitions that correctly execute at runtime, but trigger errors on static analysis.

Real application example: https://github.com/koel/koel/blob/master/app/Builders/AlbumBuilder.php#L10-L13
Note: in this case, an error can be resolved by specifying the return type as `self`, but it is not always correct.

Possible related issues from phpstan: https://github.com/phpstan/phpstan/issues/5132 https://github.com/phpstan/phpstan/issues/4648
Note: solution with `self` used from these issues

### Code where the issue was found

```php
use Illuminate\Database\Eloquent\Builder as BaseBuilder;

/**
 * @template TModelClass of \Illuminate\Database\Eloquent\Model
 *
 * @extends BaseBuilder<TModelClass>
 */
abstract class Builder extends BaseBuilder
{
    public function whereSomething($column, $values, $boolean = 'and'): static
    {
        return $this->whereNotIn(...func_get_args());
    }

    public function whereSomethingElse($column, $values, $boolean = 'and'): self
    {
        return $this->whereNotIn(...func_get_args());
    }
}
```

----

I managed to find the exact place where the type is defined. I tried to replace `\PHPStan\Type\Generic\GenericObjectType` with `\PHPStan\Type\StaticType` but then it fails in other places.

https://github.com/nunomaduro/larastan/blob/706784ec6b1bad487692730dc20aeee0ff1b19fa/src/Methods/EloquentBuilderForwardsCallsExtension.php#L150-L154

calebdw:
- [x] Added or updated tests

Related to #1871 and #1870 (but does not fully solve). Closes #1676. Should close #1538 (but will want to confirm).

**Changes**
Hello!

This PR updates the Builder methods to properly return `$this` (previously there was a mix of $this, static, and self) instead of creating new object classes.

I also included some slight refactors to simplify and improve the readability of the code---this includes things like removing duplicated code, returning early, reducing the number of indentation levels, etc. However, I've carefully segregated the changes into their own commits to make reviewing the changes easier---if the whole diff is a little hard to follow the individual commits should help to walk through the changes made.

Breaking changes

If users have specified the return type to be `static` or `self` they might need to update to `$this` (depending on their phpstan level).

Thanks!

### DIFF
--- a/src/Methods/EloquentBuilderForwardsCallsExtension.php
+++ b/src/Methods/EloquentBuilderForwardsCallsExtension.php
@@ -16,10 +16,9 @@ use PHPStan\Reflection\MissingMethodFromReflectionException;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\ShouldNotHappenException;
-use PHPStan\Type\Generic\GenericObjectType;
 use PHPStan\Type\Generic\TemplateObjectType;
 use PHPStan\Type\IntegerType;
-use PHPStan\Type\ObjectType;
+use PHPStan\Type\ThisType;
 
 use function array_key_exists;
 use function in_array;
@@ -113,7 +112,7 @@ final class EloquentBuilderForwardsCallsExtension implements MethodsClassReflect
             } elseif ($methodName === 'restoreOrCreate' || $methodName === 'createOrRestore') {
                 $returnType = $modelType;
             } else {
-                $returnType = new GenericObjectType($classReflection->getName(), [$modelType]);
+                $returnType = new ThisType($classReflection);
             }
 
             return new EloquentBuilderMethodReflection(
@@ -134,10 +133,8 @@ final class EloquentBuilderForwardsCallsExtension implements MethodsClassReflect
 
         if (in_array($methodName, $this->builderHelper->passthru, true)) {
             $returnType = $parametersAcceptor->getReturnType();
-        } elseif ($classReflection->isGeneric()) {
-            $returnType = new GenericObjectType($classReflection->getName(), [$modelType]);
         } else {
-            $returnType = new ObjectType($classReflection->getName());
+            $returnType = new ThisType($classReflection);
         }
 
         return new EloquentBuilderMethodReflection(
--- a/src/Methods/EloquentBuilderForwardsCallsExtension.php
+++ b/src/Methods/EloquentBuilderForwardsCallsExtension.php
@@ -29,6 +29,9 @@ final class EloquentBuilderForwardsCallsExtension implements MethodsClassReflect
     /** @var array<string, MethodReflection> */
     private array $cache = [];
 
+    /** @var string[] */
+    private array $softDeletesMethods = ['withTrashed', 'onlyTrashed', 'withoutTrashed', 'restore', 'createOrRestore', 'restoreOrCreate'];
+
     public function __construct(private BuilderHelper $builderHelper, private ReflectionProvider $reflectionProvider)
     {
     }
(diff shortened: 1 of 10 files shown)


jnoordsij:
As mentioned in https://github.com/larastan/larastan/issues/1538#issuecomment-1987873622, can confirm this solves the issues reported on #1538.

canvural:
Hi,

Like I explained in other PR, this is not the way to go. Errors like these are always expected when you are analyzing the scopes, relations etc. itself. Because PHPStan does not have the right information. Most important thing is to look if there are errors or not when analyzing the code that calls these scopes/relations. If not then everything is fine 👍🏽 

I know this is causing issues, but I think this is a much bigger issue which might need to be solved in PHPStan.

Thanks for understanding.

calebdw:
@canvural, this is the correct fix---**all builder methods do return `$this`** as opposed to `self` (a new instance of the declaring class) or `static` (a new instance of a child class). The same instance is returned and this PR correctly captures that.

> Errors like these are always expected when you are analyzing the scopes, relations etc. itself. Because PHPStan does not have the right information. Most important thing is to look if there are errors or not when analyzing the code that calls these scopes/relations. If not then everything is fine 

I would disagree here because this PR *does understand* what's going on and it fixed the linked issues. I'm not a fan of ignoring errors, especially when there is a solution that does not throw errors in the first place.

nxu:
I completely agree with @calebdw, this is the correct solution. Even Laravel itself uses `@return $this` in the builder method docblocks, the builder methods always return the object itself, never a new instance.

As I see it, this PR does exactly what Larastan is supposed to do: it provides a Laravel specific solution to a Laravel specific PHPStan issue.

jnoordsij:
(preface: I'm not exactly sure about the specifications given I haven't checked this in a while, but let me try to join in based on the things I see and do remember)

> Because PHPStan does not have the right information.

Isn't that exactly the originating problem and thus the whole point of this fix? Providing PHPStan with more accurate information, so it can check in a more accurate manner?

I do agree it would be much neater if there was some way where PHPStan could inherit that itself in some way, but it needs some form of input to do so, right? If this PR does exactly that (and I haven't heard yet of anything contradicting that), what exactly is the issue with proceeding in this?

Furthermore, if this would require some change on the PHPStan side, then probably it would be better to at least create some viable MWE to show where PHPStan could do better, report that at their end and then list that issue here as a blocker that should provide an alternate solution.

canvural:
Again, this is not the correct solution.
 > the builder methods always return the object itself, never a new instance.
 
 I don't understand this. There is no difference between those two in the eyes of static analysis.

> Isn't that exactly the originating problem and thus the whole point of this fix? Providing PHPStan with more accurate information, so it can check in a more accurate manner?

PHPStan and Larastan already understands the code good enough when analyzing the builder calling code. So there is no problem there. The issue arrives when you analyze the builder class itself. But the issues there are low priority. And happens with even just PHPStan.

calebdw:
> Again, this is not the correct solution.

@canvural, you're so adamant that this is not the correct solution, however, you've yet to provide the "correct" solution. I understand you haven't had time to fully work through the solution, but in your mind, what form would the "correct" solution take? Simply ignoring the errors is not a solution...

> I don't understand this. There is no difference between those two in the eyes of static analysis.

This is simply not true---PHP itself only has `self` and `static` return types. If there was no difference between `static` and `$this` "in the eyes of static analysis", then PHPStan would not have added the `$this` return type. I've found most often that the difference manifests itself in the retention of generic information (or lack thereof) when chaining method calls---particularly when these methods exist on a parent class/trait.

Here is what the ChatGPT has to say:
<details>
Prompt:

> What is the difference between the `static` and the `$this` return type in PHPStan? Is there any difference in the eyes of static analysis? How does one over the other affect generic types where a certain where a certain class has the form `Foo<Bar>`?

In PHPStan, and more broadly in the field of static analysis for PHP, the return types static and $this are used to provide more specific type information, especially in a context where method chaining and inheritance are involved. Understanding the difference between these two is crucial for correct type hinting and can significantly affect how static analysis tools interpret the code, especially when dealing with generic types.

### 1. Understanding static and $this Return Types
- **`$this` Return Type**: When a method's return type is hinted as `$this`, it indicates that the method returns the same instance of the class on which it was called. This is most useful in fluent interfaces where method chaining is expected. PHPStan understands that the returned type is exactly the same object as `$this`, with the same class and type parameters.
- **`static` Return Type**: The `static` return type is similar but subtly different. It indicates that the method returns an instance of the same class where the method is defined, or possibly a subclass. This is particularly important in inheritance scenarios. `static` is a way to denote late static binding which means the return type will be the class of the object upon which the method is called, not necessarily the class where the method is defined.
### 2. Differences in Static Analysis
In the realm of static analysis, these differences are important:

- **Using `$this`**: When `$this` is used as a return type, PHPStan (or any static analyzer) expects that the method will return exactly the same type of object, including all generic parameters. For instance, if a class is generic like `Foo<Bar>` and a method returns 
- '$this`, PHPStan understands that the returned object is also of type `Foo<Bar>`.
- **Using `static`**: The use of `static` allows for a more flexible return type, which can vary depending on the subclass from which it is called. If `Foo<Bar>` is a base class and `Baz` is a subclass, a method in `Foo` with a return type of `static` can return an instance of `Baz`. This is particularly useful in factory methods or when implementing the prototype pattern. However, this can introduce complexities in type inference, especially with generics, as `static` could refer to any subclass with potentially different type parameters.
### 3. Impact on Generic Types
When dealing with generic types like `Foo<Bar>`, the choice between `$this` and `static` can significantly affect type safety and correctness:

- **Methods Returning `$this`**: These are straightforward in their implications. PHPStan will enforce that the returned object is not only of the same class but also maintains the same generic parameters. This is safer in terms of type correctness.
- **Methods Returning `static`**: This allows for more flexibility, as subclasses can maintain or change the generic parameter. However, this can potentially lead to type safety issues if not properly handled. For instance, if a subclass does not correctly override generic parameters, it might lead to situations where the static analysis tool cannot accurately predict the type or enforces an incorrect one.
### 4. Conclusion
In summary, choosing between `$this` and `static` depends largely on the specific requirements of your codebase, especially regarding inheritance and method chaining. For maintaining strict type safety with generics, `$this` is more appropriate, whereas `static` offers more flexibility at the potential cost of type safety. Static analysis tools like PHPStan will treat these two differently, and understanding these nuances is key to leveraging these tools effectively for cleaner, more reliable code.

</details>

> The issue arrives when you analyze the builder class itself. But the issues there are low priority.

I understand and respect that it's a low priority for Larastan, however, I would like Larastan to be able to analyze my builders without errors ( I'm not a fan of ignoring errors)---this is why I implemented the fix, because it wouldn't have been fixed otherwise (in the near term at least).

I've spent a lot of time on this, and I believe returning `$this` from the builders to be the most correct solution (although granted my implementation could probably be improved :sweat_smile:). Again, returning `$this` **is exactly what the actual builder classes return**, and this **fixes the issues when analyzing the builder classes** so I fail to see what about this implementation is so "incorrect" that it could not be merged (all the tests pass as well indicating nothing broke).

### Path Forward
So then, **what is the path forward and how do we fix the issue at hand**? Going back and forth like this is not likely to change anyone's mind (although I hope I've presented the case well enough to change your mind :upside_down_face:). *I want to work with you to fix the issue, not argue against you and nothing gets done.* I would appreciate some direction instead just being shut down with "this is not the correct solution" and no indication of what the correct solution might be.

Thank you for your time and all the hard work you've put into this awesome package that we all love and want to make even better.

PR #1876, opened 2024-03-10
fix: factory trashed method

### LINKED ISSUE #1730: Call to an undefined method Database\Factories\<MyModel>::trashed()
- Larastan Version: 2.6.4
- `--level` used: 6

### Description
I got warning about using the `trashed` state method, that is defined in `Illuminate\Database\Eloquent\Factories\Factory`

```php
$user = \App\Models\User::factory()->trashed()->create();
```

I even tried to annotate the method in my `UserFactory` but this does not solve the issue. The only way I figurate out to solve the issue (without ignoring the line) is to define a custom state that calls parent trashed, but of course is not that cool

calebdw:
- [x] Added or updated tests

Closes #1730

**Changes**

This updates the factory method extension to account for the virtual `trashed` method.

Thanks!

### DIFF
--- a/tests/Type/data/model-factories.php
+++ b/tests/Type/data/model-factories.php
@@ -7,8 +7,7 @@ use App\User;
 
 use function PHPStan\Testing\assertType;
 
-function doFoo(): void
-{
+function dummy(?int $foo): void {
     assertType('Database\Factories\UserFactory', User::factory());
     assertType('Database\Factories\UserFactory', User::factory()->new());
     assertType('App\User', User::factory()->createOne());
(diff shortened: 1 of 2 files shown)


szepeviktor (approved):

tests/Type/data/model-factories.php
```diff
@@ -7,81 +7,70 @@
 
 use function PHPStan\Testing\assertType;
 
-function doFoo(): void
```
canvural, line 10: `assertType` calls should always be in a dummy function. Please revert.

calebdw: @canvural, why is that? We have several test files in this project that don't follow that convention:

- tests/Type/data/environment-helper.php
- tests/Type/data/translator.php
- tests/Type/data/request-user.php
- tests/Type/data/collection-stubs.php
- ...

**All** of the type test files in the Laravel framework don't follow that:
- https://github.com/laravel/framework/blob/48246da2320c95a17bfae922d36264105a917906/types/Support/Helpers.php#L5

And the same thing with PHPStan tests:
- https://github.com/phpstan/phpstan-src/blob/f9361865944f2da686049fcdeef173065052d82b/tests/PHPStan/Analyser/data/round-php8.php#L13


I can revert if you want, but I feel that the dummy function is unnecessary, adds an extra indentation level, and the function name doesn't even describe the tests being performed.

canvural: I don't remember the exact reason, but it was something related to performance. It's related to PHPStan's internals.

And yes, all the examples you gave, I'd say they are wrong. I've been trying to move away and wrap them in dummy functions where possible.

calebdw: I'd rather get clarification from Ondre before we just call his own tests wrong...that performance issue could have been fixed

canvural: The example file you gave is 2 years old. It's normal it's not updated.

calebdw: I wrapped the tests in a closure

tests/Type/data/model-factories.php
```diff
@@ -7,8 +7,7 @@
...
-{
+function (?int $foo) {
```
canvural: Isn't this a syntax error? 🤔 Doesn't it need a name? Even if it is valid, I'd rather have a `dummy` name.

calebdw: No syntax error, it's just an anonymous closure. PHPStan does this with their files:

https://github.com/phpstan/phpstan-src/blob/6026869f3eec4623a0be812ef69e1dc27a430d4e/tests/PHPStan/Analyser/data/bug-4339.php#L7

Personally I'd rather just have an anonymous closure instead of a named function like `doFoo` or `dummy`. You can also run into issues with namespace clashes with that approach.

canvural: That's why we have unique namespace for each data file already. And for readability I'd rather have a dummy name.

calebdw: done

canvural:
Thank you.

PR #1878, opened 2024-03-11
feat: support L11 casts method

### LINKED ISSUE #1877: Support Laravel 11 new method of model casts
- Larastan Version: 11.x
- `--level` used: 5

### Description
Using new L11 casts function doesnt work, as shown in the print below it cant understand the property as an array:

Capturar

calebdw:
- [x] Added or updated tests

Closes #1877 

Thanks!

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,11 +1,6 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Constant LARAVEL_VERSION not found\\.$#"
-			count: 1
-			path: src/LarastanStubFilesExtension.php
-
-		-
 			message: "#^PHPDoc tag @var assumes the expression with type PHPStan\\\\Type\\\\Type\\|null is always PHPStan\\\\Type\\\\Constant\\\\ConstantStringType but it's error\\-prone and dangerous\\.$#"
 			count: 1
 			path: src/Methods/HigherOrderCollectionProxyExtension.php
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -6,3 +6,5 @@ parameters:
     phpVersion: 80200
     paths:
         - src
+    ignoreErrors:
+        - "#^Constant LARAVEL_VERSION not found\\.$#"
--- a/src/Properties/ModelPropertyHelper.php
+++ b/src/Properties/ModelPropertyHelper.php
@@ -18,12 +18,14 @@ use PHPStan\Type\ObjectType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\TypeCombinator;
 use ReflectionException;
+use ReflectionMethod;
 
 use function array_key_exists;
 use function count;
 use function implode;
 use function in_array;
 use function method_exists;
+use function version_compare;
 
 class ModelPropertyHelper
 {
@@ -86,6 +88,12 @@ class ModelPropertyHelper
             $modelInstance = $classReflection->getNativeReflection()->newInstanceWithoutConstructor();
         } catch (ReflectionException) {
             throw new ShouldNotHappenException();
+        }
+
+        if (version_compare(LARAVEL_VERSION, '11.0.0', '>=')) {
+            // Needed to merge the model casts method into the property
+            $reflection = new ReflectionMethod($modelInstance, 'initializeHasAttributes');
+            $reflection->invoke($modelInstance);
         }
 
         $tableName = $modelInstance->getTable();
--- a/tests/Type/GeneralTypeTest.php
+++ b/tests/Type/GeneralTypeTest.php
@@ -78,6 +78,10 @@ class GeneralTypeTest extends TypeInferenceTestCase
             yield from self::gatherAssertTypes(__DIR__ . '/data/bug-1819.php');
         }
 
+        if (version_compare(LARAVEL_VERSION, '11.0.0', '>=')) {
+            yield from self::gatherAssertTypes(__DIR__ . '/data/model-casts-l11.php');
+        }
+
         //##############################################################################################################
 
         // Console Commands
--- a/tests/Type/data/model-casts-l11.php
+++ b/tests/Type/data/model-casts-l11.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace ModelProperties;
+
+use function PHPStan\Testing\assertType;
+
+/** @var \App\User $user */
+assertType('array', $user->l11_cast);
--- a/tests/application/app/User.php
+++ b/tests/application/app/User.php
@@ -214,6 +214,13 @@ class User extends Authenticatable
         $this->active = 1;
     }
 
+    protected function casts(): array
+    {
+        return [
+            'l11_cast' => 'array',
+        ];
+    }
+
     protected function newStyleAttribute(): Attribute
     {
         return Attribute::make(
--- a/tests/application/database/migrations/2020_01_30_000000_create_users_table.php
+++ b/tests/application/database/migrations/2020_01_30_000000_create_users_table.php
@@ -56,6 +56,7 @@ class CreateUsersTable extends Migration
             $table->date('immutable_date');
             $table->dateTime('immutable_datetime');
             $table->timestamp('timestamp');
+            $table->text('l11_cast');
 
             $table->timestamps();
             $table->softDeletes();


tests/Integration/IntegrationTest.php
```diff
@@ -28,6 +28,12 @@ public static function dataIntegrationTests(): iterable
         }
 
         yield [__DIR__ . '/data/eloquent-builder-l10.php'];
+
+        if (! version_compare(LARAVEL_VERSION, '11.0.0', '>=')) {
+            return;
+        }
+
+        yield [__DIR__ . '/data/model-casts-l11.php'];
```
szepeviktor: For Laravel 12

```suggestion
        if (version_compare(LARAVEL_VERSION, '11.0.0', '>=')) {
            yield [__DIR__ . '/data/model-casts-l11.php'];
        }
```

szepeviktor (approved):

tests/Integration/IntegrationTest.php
calebdw: I could do that, but this was the existing pattern there which is what I followed:

https://github.com/larastan/larastan/blob/2ac60334eb7e5d693df9a9150bd6798be680f99c/tests/Integration/IntegrationTest.php#L26-L30

szepeviktor: All right.

zepfietje:
Are there any plans to merge this PR and tag a new release, @szepeviktor?
Laravel 11 has been out for almost two weeks now, and it's really annoying to get PHPStan errors in new applications that use the updated casting approach out of the box. 🙂

szepeviktor:
@zepfietje I was told not to make releases.

zepfietje:
> @zepfietje I was told not to make releases.

Could you clarify,  @szepeviktor? 😄

szepeviktor:
Only @canvural knows what the plans are for Larastan.

zepfietje:
I've reached out to @canvural on X, since I believe this PR is quite important as it affects all new Laravel projects.

szepeviktor:
The "data" file is in the wrong directory: `tests/Type/data/model-casts-l11.php`

szepeviktor (approved):

studnitz:
If you don't want to wait until this is merged and released, you can alias the fix intoyour project with these composer commands:

```sh
composer config repositories.larastan-fix vcs https://github.com/calebdw/larastan  
composer require --dev --with-all-dependencies "larastan/larastan:dev-l11_casts as 2.9.2-dev"
```

src/Properties/ModelPropertyExtension.php
```diff
@@ -132,6 +134,13 @@ public function getProperty(
             throw new ShouldNotHappenException();
         }
 
+        if (version_compare(LARAVEL_VERSION, '11.0.0', '>=')) {
+            // Needed to merge the model casts method into the property
+            // @phpstan-ignore-next-line
+            $reflection = new ReflectionMethod($modelInstance, 'initializeHasAttributes');
+            $reflection->invoke($modelInstance);
```
canvural: This is not ideal. We should not add new code that actually executes the analyzed code. I know there are codes like that already, but the least we could do is to avoid adding new ones.

We might check if the `casts` method has return type with array shape maybe. If yes we can use that directly! If not we need to parse the file and analyze it's contents to understand the shape it returns. But I hope most of the developers will write the array shape return type

canvural (changes requested):

src/Properties/ModelPropertyExtension.php
calebdw: I know it's not ideal, but the whole reason that Laravel included the `casts()` method was so that you could have more complicated expressions (like function calls) in the cast definition:

```php
/**
 * Get the attributes that should be cast.
 *
 * @return array<string, string>
 */
protected function casts(): array
{
    return [
        'options' => AsCollection::using(OptionCollection::class),
    ];
}
```

instead of trying to string concatenate a class name with argument together:

```php
protected $casts = [
        'options' => AsCollection::class.':'.OptionCollection::class,
];
```

Behind the scenes, everything gets compiled down to `array<string, string>` so that it can be merged into the `$casts` property (that Larastan understands), but the function needs to be called to do that. I'm not sure how well Larastan will be able parse more complicated expressions.

I don't see this as any different than using the container when needed (which executes code). It's also a very simple solution (only two lines, not including the if condition) whereas trying to parse the file, analyze the contents, handle edge conditions, etc. would be a much more complex undertaking. This code even calls the `getCasts()` method so we don't have to try and parse the model `$casts` property ourselves.

Thoughts?

canvural: > whole reason that Laravel included the casts() method was so that you could have more complicated expressions

Exactly! What if this method does some side effects or does HTTP calls (exaggerating but you get the point) ? Reading the array shape return type would be much safer.

> I don't see this as any different than using the container when needed (which executes code). It's also a very simple solution (only two lines, not including the if condition) whereas trying to parse the file, analyze the contents, handle edge conditions, etc. would be a much more complex undertaking. This code even calls the getCasts() method so we don't have to try and parse the model $casts property ourselves.

Yep, existing code is also wrong. Should not have been there. But some stuff is hard to remove, so we rolled with it. At least we can do is not do add new codes like that.

Executing code caused so much trouble in the past. And we slowly moved away from it. The existence of the code that executes the analyzed code does not mean it is the correct way.

canvural: Also, the more I think about this I come to the conclusion that developers should provide the array shape and we read that. We don't even need to parse the file.

calebdw: > What if this method does some side effects or does HTTP calls (exaggerating but you get the point) 

Then I would say don't do stupid things inside the casts method? :upside_down_face: This method is automatically called by the framework everytime the model is booted so you'll be bitten real fast if there's side effects in there. Given this, I don't think it's unreasonable / unsafe for us to call it

The casts work as-is, how about we release this to support L11 and then we can revisit how we parse the casts in v3? If you're opposed to calling this method then we would also need to not call `getCasts`---although that could break things if someone overrides that method.

> Also, the more I think about this I come to the conclusion that developers should provide the array shape and we read that. We don't even need to parse the file.

What exactly do you mean by this? Are you saying they would have to duplicate the array shape with `@returns`? Something like:

```php
/**
 * Get the attributes that should be cast.
 *
 * @return array{options: OptionCollection}
 */
protected function casts(): array
{
    return [
        'options' => AsCollection::using(OptionCollection::class),
    ];
}
```

However, this would fail phpstan because `array{option: string}` is actually returned...which is not helpful for us to use

canvural: > Then I would say don't do stupid things inside the casts method? 🙃

Exactly. But people will do that and they will come here to complain about it. Happened in past.

> However, this would fail phpstan because array{option: string} is actually returned...which is not helpful for us to use

Yeah, I don't have a solution for that now.. But it's only an issue with the `using` method

calebdw: > But it's only an issue with the using method

I don't think so---it would be an issue for all of the casts if that's the road you're trying to go down:

```php
/**
 * Get the attributes that should be cast.
 *
 * @return array{
 *     options: OptionCollection,
 *     started_at: Carbon,
 *     is_complete: bool,
 *     status: int,
 *     total: float,
 * }
 */
protected function casts(): array
{
    return [
        'options' => AsCollection::using(OptionCollection::class),
        'started_at' => 'datetime',
        'is_complete' => 'bool',
        'status' => 'int',
        'float' => 'float',
    ];
}
```

All of these would fail because the type is actually `string` for all of the array keys:

```php
 * @return array{
 *     options: string,
 *     started_at: string,
 *     is_complete: string,
 *     status: string,
 *     total: string,
 * }

canvural:
```php
* @return array{
 *     options: 'OptionCollection',
 *     started_at: 'Carbon',
 *     is_complete: 'bool',
 *     status: 'int',
 *     total: 'float',
 * }
 ```

this is what I meant. Just like `$casts` property. And we handle that currently by mapping these string to PHPStan type.

calebdw: I'm still not sure how viable that is given that the `casts` method was created due to the headaches of trying to define more complicated types solely by strings / string concatenation.

Not sure many people would care for the duplication either---if I can describe the cast by a simple string there's no reason for me to create the `casts` method with the `@return` doc when I can just use the `$casts` property

canvural: > if I can describe the cast by a simple string there's no reason for me to create the casts method

Great. So you can keep using `$casts` property. It's not deprecated.

mdpoulter: While the `$casts` property might not be deprecated, it's no longer documented in the latest documentation (https://laravel.com/docs/11.x/eloquent-mutators#attribute-casting). It seems somewhat paradoxical to prioritize safeguarding developers from potential pitfalls while simultaneously expecting them to be familiar with previous framework versions.

gabrielrbarbosa: I agree with "how about we release this to support L11" since it was released almost 2 months ago and this is breaking compatibility for all users...

marco-introini: I cannot agree more: all my CIs are red for 2 months for this specific problem, and I believe that many people are in my situation

canvural: These errors are not a blocker. You can ignore them and make your build green. Larastan supports L11. It is just missing a feature. It's not urgent or anything.

And like I said, I cannot merge this implementation.

jakelehner:
Has this issue been closed and branch deleted **without** it being merged?

What's going on here? This is embarrassing for this package.

@canvural 
* No, Larastan does not currently support L11 due to not supporting the documented `casts` method. 
* No, we cannot simply "ignore them and make [our] build green. I tried that. The number of ignore rules this requires is unrealistic in our code base and defeats the purpose of having Larastan run.
* Laravel Shift converts all cast arrays to the new cast method. So anyone using Shift to upgrade and running Larastan will run in to this issue. Our only options at this stage is to rollback the change, or insert a ton of hacky ignore rules as noted. Neither of these are acceptable solutions.
* At this stage, Larastan is **the only blocker** I have to upgrading to L11. Yes, it is a blocker. Your suggestions are not acceptable for our production environment.

Guys, seriously this is embarrassing for this package. 

Honestly if this isn't resolved soon I may just very well have to remove the package from our repo and go without code analysis until it's resolved.

canvural:
> Has this issue been closed and branch deleted without it being merged?

Yes. Good observation. Now work on your reading skills and you'd understand why.

> No, Larastan does not currently support L11 due to not supporting the documented casts method.

Last time I checked, you could install Larastan to a project that uses Laravel 11. That means it supports it. Missing features does not mean it doesn't support it.

> No, we cannot simply "ignore them and make [our] build green. I tried that. The number of ignore rules this requires is unrealistic in our code base and defeats the purpose of having Larastan run.

Use a baseline then.

> Laravel Shift converts all cast arrays to the new cast method. So anyone using Shift to upgrade and running Larastan will run in to this issue. Our only options at this stage is to rollback the change, or insert a ton of hacky ignore rules as noted. Neither of these are acceptable solutions.

Sounds like a you problem.

> At this stage, Larastan is the only blocker I have to upgrading to L11. Yes, it is a blocker. Your suggestions are not acceptable for our production environment.

Again, you are the one refusing to work with it. Nothing I can do. Larastan supports Laravel 11.

> Guys, seriously this is embarrassing for this package.

Why do you act like I owe you something? This is an open source project. Deal with it.

> Honestly if this isn't resolved soon I may just very well have to remove the package from our repo and go without code analysis until it's resolved.

Oh, please yes! I hope you'll stop using it forever honestly.

adamgyongyosi:
@canvural thx

https://github.com/larastan/larastan/commit/66db15f82209d971e9bcca99ea3b2cb031be9d52

a-bashtannik:
@jakelehner I totally understand your concerns and condemn these toxic responses: 
> Oh, please yes! I hope you'll stop using it forever honestly.

But it's an open-source world, always a fight between contributors and users. I think the requirement of array shape tag is an unacceptable solution if you have many lines of code. We have models with dozens and hundreds of fields - imagine how our models should look now. 

I would like to share a compromise solution we found: we used barryvdh/laravel-ide-helper package which can generate and cast attributes to `@property` phpdoc tags on the model class. It also creates some trash, like magic methods (e.g. `whereIsToxic`) but it is quite easy to cleanup useless data using some regexp. For complex casts you still need to define `@properties` manually but even in this case, the solution can save you a lot of time. 

Good luck!

canvural:
> @jakelehner I totally understand your concerns and condemn these toxic responses: 
> > Oh, please yes! I hope you'll stop using it forever honestly.
> 
> But it's an open-source world, always a fight between contributors and users. I think the requirement of array shape tag is an unacceptable solution if you have many lines of code. We have models with dozens and hundreds of fields - imagine how our models should look now. 
> 
> I would like to share a compromise solution we found: we used barryvdh/laravel-ide-helper package which can generate and cast attributes to `@property` phpdoc tags on the model class. It also creates some trash, like magic methods (e.g. `whereIsToxic`) but it is quite easy to cleanup useless data using some regexp. For complex casts you still need to define `@properties` manually but even in this case, the solution can save you a lot of time. 
> 
> Good luck! 
> 

So you think my response was toxic but not the original one?

PR #1880, opened 2024-03-13
test: add test for plucking computed properties

### LINKED ISSUE #1188: collection->pluck() for custom attributes shouldn't raise warning
- Larastan Version: 2.1.2
- `--level` used: 6
- Pull request with failing test: don't know how

### Description

If your model has a custom attribute that **isn't** just modified by an accessor but creates a new one from other values, then you can't use pluck() on the Builder. You can only do it on the collection, once the models with at least the relevant attributes. As such, the warning should not be appearing when either of the parameters of the pluck() call is such an attribute.

Assume you have a User model with first_name and last_name database columns. You then have a custom attribute:

```php
public function getFullNameAttribute(): string
{
  return $this->first_name . ' ' . $this->last_name;
}
```
(Equivalently, in Laravel 9 Attribute format:)
```php
public function fullName(): Attribute
{
  return Attribute::get(fn() => $this->first_name . ' ' . $this->last_name);
}
```

Then, having the line `User::get(['first_name', 'last_name'])->pluck('full_name')` succeeds but raises a Larastan warning "Called 'pluck' on Laravel collection, but could have been retrieved as a query.  ". But `User::pluck('full_name')` fails with 
`Illuminate\Database\QueryException with message 'SQLSTATE[42S22]: Column not found: 1054 Unknown column 'full_name' in 'field list' (SQL: select full_name from users)'`

calebdw:
Hello!

This PR adds a test that closes #1188.

Thanks!

### DIFF
--- a/tests/Integration/data/eloquent-builder-l10.php
+++ b/tests/Integration/data/eloquent-builder-l10.php
@@ -11,3 +11,5 @@ User::query()->where(DB::raw('1'), 1)->get();
 /** @see https://github.com/larastan/larastan/issues/1806 */
 User::query()->orderBy(Post::query()->select('id')->whereColumn('user_id', 'users.id'));
 User::query()->orderByDesc(Post::query()->select('id')->whereColumn('user_id', 'users.id'));
+
+User::query()->get()->pluck('computed');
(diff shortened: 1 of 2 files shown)


szepeviktor:
Thank you.

PR #1882, opened 2024-03-15
fix: check php version before accessing isEnum

calebdw:
Not sure why ci started failing, but this should hopefully fix it...

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -53,7 +53,7 @@ jobs:
 
           -
             repository: filamentphp/filament
-            ref: 60af69e66b0289c0a9cffd7c5930028ef6b715c5
+            ref: fa626b4b82e2439fcfc3d4941e1310a64b2c009f
             php-version: "8.2"
             config: filamentphp-filament.neon
             baseline: filamentphp-filament.baseline.neon
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -65,8 +65,8 @@ jobs:
           tools: "composer:v2"
           coverage: "none"
 
-      - name: "Check Composer configuration"
-        run: "composer validate --strict"
+      # - name: "Check Composer configuration"
+      #   run: "composer validate --strict"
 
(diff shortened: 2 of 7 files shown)


calebdw:
So PHP 8.0.6 is installed in the ci, but it's still passing the version check?

szepeviktor:
> So PHP 8.0.6 is installed in the ci, but it's still passing the version check?

This is a custom `ReflectionProvider`
https://github.com/phpstan/phpstan-src/blob/1.11.x/src/Reflection/ReflectionProvider.php

calebdw:
The error message is: `Error: Call to undefined method ReflectionClass::isEnum()`, if it were the custom reflection provider then it would be `Error: Call to undefined method PHPStan\Reflection\ClassReflection::isEnum()`.

However, that still doesn't explain:
1. why it suddenly started failing in the first place
2. the `isEnum()` call is still behind a version check so shouldn't be called
3. the `isEnum()` call works on PHP 8.1 tests

calebdw:
@ondrejmirtes, something was introduced in PHPStan 1.10.61 that is causing this run to fail:

https://github.com/larastan/larastan/actions/runs/8397529980/job/23000977060?pr=1882

I pinned phpstan at 1.10.60 and the job does not fail (however, job fails up to the latest version 1.10.64).

calebdw:
This commit might fix the job?

https://github.com/phpstan/phpstan-src/commit/4c167d7ff11dee9122a298ab0eb941ac61bb2ff1

ondrejmirtes:
This is a bug that's purely in PHPStan, you don't need to do anything in Larastan to work around that, I'm going to fix it.

ondrejmirtes:
The fix:

* https://github.com/ondrejmirtes/BetterReflection/commit/94ca1336c58332604a74012d20286a240db376ea

It's going to be released in PHPStan 1.10.65 soon.

ondrejmirtes:
Let's try this: https://github.com/phpstan/phpstan/releases/tag/1.10.65

calebdw:
Thanks Ondre!

PR #1879, opened 2024-03-12
[3.x] feat: support multiple database connections

calebdw:
- [x] Added or updated tests
- [x] Documented user facing changes

Closes #1577, closes #1642, closes #2002

**Changes**

This PR introduces support for multiple database connections as well as Postgres Schema (sub databases).

**Breaking changes**
None that should affect the user, but the existing code allowed for parsing multiple schema files for a single connection (i.e., the only connection). However, Laravel saves the schema squash files as `<connection>-schema.sql`. So to able to understand which schema file belongs to what connection I rely on this naming convention to determine the connection name from the filename. In practice this means there should only be one file per connection.

For example, the following files would both be parsed under the `mysql` connection:
- `mysql-schema.{sql,dump}`
- `mysql.{sql,dump}`

This has been released on my [fork](https://github.com/calebdw/larastan) if anyone would like to take advantage of this immediately.

PR #1884, opened 2024-03-15
Support schema variable

adriaanzon:
In my project I'm interacting with Laravel Telescope's [EntryModel](https://github.com/laravel/telescope/blob/5.x/src/Storage/EntryModel.php). Because its [migration](https://github.com/laravel/telescope/blob/5.x/database/migrations/2018_08_08_100000_create_telescope_entries_table.php) first assigns the schema builder to a variable and then calls the `create()` method on that variable, Larastan isn't currently able to understand this migration file when adding it to `databaseMigrationsPath`. I get the following error:

```
Access to an undefined property Laravel\Telescope\Storage\EntryModel::$content.
```

I modified one of the migrations to use a variable, to demonstrate that the tests fail, and I provided a possible solution with this pull request. However, I'm not very familiar with this codebase and php-parser, and I'm open to any suggestions/changes. I'm also aware that there is an open PR that touches the same code (#1879), so after that has been merged, this needs to change anyway.

### DIFF
--- a/src/Properties/SchemaAggregator.php
+++ b/src/Properties/SchemaAggregator.php
@@ -83,6 +83,28 @@ final class SchemaAggregator
                 && ($stmt->expr->class->toCodeString() === '\Schema' || (new ObjectType('Illuminate\Support\Facades\Schema'))->isSuperTypeOf(new ObjectType($stmt->expr->class->toCodeString()))->yes())
             ) {
                 $statement = $stmt->expr;
+            } elseif (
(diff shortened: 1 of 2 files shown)


calebdw:
Nice!

Though I'm not sure that Larastan would be able to understand which connection that migration should go under since it's hidden behind a method call and then a config call---even in #1879 it would just put the schema into the default connection

canvural:

PR #1888, opened 2024-03-22
feat: add laravel and php version to issue template

calebdw:
Hello!

This just adds the Laravel and PHP version to the issue template as that can provide valuable context.

Thanks!


szepeviktor:
I would gladly ask people for Laravel version on a case-by-case basis.

calebdw:
Is there any reason to not just ask for it in the template?

szepeviktor:
I don't make decisions here.
I think filling in fields annoys people.

canvural:
Hi,

I updated the issue templates to the new version and added the `Laravel version` there.
Thanks!

PR #1899, opened 2024-04-02
test: regression test for mapping custom simple collections

erikgaal:
- [x] Added or updated tests

Since `2.9.1` a change was introduced to Collections where calling `map` or `mapWithKeys` on a custom `\Illuminate\Support\Collection` would make PHPStan report returning a basic `Collection`, even though the custom Collection was actually returned.

On `a79b46b9` running this testcase results in the following failure:

```
1) Tests\Type\CollectionDynamicReturnTypeExtensionsTest::testFileAsserts with data set "/Users/erikgaal/Dev/larastan/tests/Type/data/collection-generic-static-methods-l948.php:204" ('type', '/Users/erikgaal/Dev/larastan...48.php', 'App\FavoritesCollection', 'Illuminate\Support\Collection...rites>', 204)
Expected type App\FavoritesCollection, got type Illuminate\Support\Collection<(int|string), App\ValueObjects\Favorites> in /Users/erik.gaal/Dev/larastan/tests/Type/data/collection-generic-static-methods-l948.php on line 204.
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'App\FavoritesCollection'
+'Illuminate\Support\Collection<(int|string), App\ValueObjects\Favorites>'
```

**Changes**

This has since been fixed in trunk, but no tests exist for this scenario. This PR adds regression tests for these two scenario's.

**Breaking changes**

None.

### DIFF
--- a/tests/Type/data/collection-generic-static-methods-l948.php
+++ b/tests/Type/data/collection-generic-static-methods-l948.php
@@ -4,6 +4,7 @@ use App\Group;
 use App\Transaction;
 use App\TransactionCollection;
 use App\User;
+use App\ValueObjects\Favorites;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Illuminate\Support\Collection as SupportCollection;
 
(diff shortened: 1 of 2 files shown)


canvural:
Hi,

If you can update the PR with the latest changes we might be able to see if the checks are passing.

PR #1906, opened 2024-04-06
fix: redis get return type

calebdw:
Hello! 

This should fix the ci, see https://github.com/phpstan/phpstan-src/pull/2990 for reference.

Thanks!


szepeviktor:
Thank you Caleb.

Do you know the reason for this change?

calebdw:
It looks like the redis get could return other data types besides just a string. That pr solved another issues on the phpstan side

PR #1907, opened 2024-04-11
Make Scope interface generic.

func0der:
- [x] Added or updated tests

**Changes**

Make the `Scope` eloquent class generic, so the `apply` method has proper type hinting and extension support.

**Breaking changes**

Not really. I could not find documentation about what you consider "breaking". Are new phpstan errors considered "breaking"? 

----

I need a little guidance or help/contribution as to how to write tests for this change. :)

### DIFF
--- a/stubs/common/EloquentBuilder.stub
+++ b/stubs/common/EloquentBuilder.stub
@@ -20,7 +20,7 @@ class Builder
      * Register a new global scope.
      *
      * @param  string  $identifier
-     * @param  \Illuminate\Database\Eloquent\Scope|\Closure  $scope
+     * @param  \Illuminate\Database\Eloquent\Scope<static, TModelClass>|\Closure  $scope
      * @return static
      */
     public function withGlobalScope($identifier, $scope);
--- a/stubs/common/EloquentBuilder.stub
+++ b/stubs/common/EloquentBuilder.stub
@@ -545,20 +545,6 @@ class Builder
 }
 
 /**
- * @template TModelClass of Model
- */
-class Scope {
-    /**
-     * Apply the scope to a given Eloquent query builder.
-     *
-     * @param  \Illuminate\Database\Eloquent\Builder<TModelClass>  $builder
-     * @param  TModelClass  $model
-     * @return void
-     */
-    public function apply(Builder $builder, Model $model);
-}
-
-/**
  * @method static \Illuminate\Database\Eloquent\Builder<static> withTrashed(bool $withTrashed = true)
  * @method static \Illuminate\Database\Eloquent\Builder<static> onlyTrashed()
  * @method static \Illuminate\Database\Eloquent\Builder<static> withoutTrashed()
--- a/stubs/common/EloquentBuilder.stub
+++ b/stubs/common/EloquentBuilder.stub
@@ -28,7 +28,7 @@ class Builder
     /**
      * Remove a registered global scope.
      *
-     * @param  \Illuminate\Database\Eloquent\Scope|string  $scope
+     * @param  \Illuminate\Database\Eloquent\Scope<static, TModelClass>|string  $scope
      * @return static
      */
     public function withoutGlobalScope($scope);
--- a/stubs/common/EloquentBuilder.stub
+++ b/stubs/common/EloquentBuilder.stub
@@ -544,7 +544,19 @@ class Builder
     public function chunk($count, $callback);
 }
 
-class Scope {}
+/**
+ * @template TModelClass of Model
+ */
+class Scope {
+    /**
+     * Apply the scope to a given Eloquent query builder.
+     *
+     * @param  \Illuminate\Database\Eloquent\Builder<TModelClass>  $builder
+     * @param  TModelClass  $model
+     * @return void
+     */
+    public function apply(Builder $builder, Model $model);
+}
 
 /**
  * @method static \Illuminate\Database\Eloquent\Builder<static> withTrashed(bool $withTrashed = true)
--- a/stubs/10.0.0/EloquentBuilder.stub
+++ b/stubs/10.0.0/EloquentBuilder.stub
@@ -20,7 +20,7 @@ class Builder
      * Register a new global scope.
      *
      * @param  string  $identifier
-     * @param  \Illuminate\Database\Eloquent\Scope|\Closure  $scope
+     * @param  \Illuminate\Database\Eloquent\Scope<static, TModelClass>|\Closure  $scope
      * @return static
      */
     public function withGlobalScope($identifier, $scope);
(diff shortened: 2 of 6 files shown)


calebdw:
Interesting, but what use-case are you trying to fix / solve here? Is there an error that you where getting that led to this change?

func0der:
The *Laravel*  `Scope::apply()` method uses `\Illuminate\Database\Eloquent\Builder`  which is a generic class as defined in this *Larastan*  stub here: https://github.com/larastan/larastan/blob/2.x/stubs/common/EloquentBuilder.stub#L6

With our implementation:

```
class AnnouncementTypeScope implements Scope
{
    /**
     * @param Builder<Update|Announcement> $builder
     */
    public function apply(Builder $builder, Model $model): void
    {
        ....
    }
}
```
 
We got this error:

```
------ ---------------------------------------------------------------------------------------------------------------------------------------------
Line app/Scopes/AnnouncementTypeScope.php
------ ---------------------------------------------------------------------------------------------------------------------------------------------
22 Parameter #1 $builder (Illuminate\Database\Eloquent\Builder<App\Models\Announcement\Announcement|App\Models\Announcement\Update>) of method
App\Scopes\AnnouncementTypeScope::apply() should be contravariant with parameter $builder (Illuminate\Database\Eloquent\Builder) of method
Illuminate\Database\Eloquent\Scope::apply()
------ ---------------------------------------------------------------------------------------------------------------------------------------------
```

calebdw:
Okay---to test you would need to create a custom scope that implements the generics and then assert the types.

Also note that this should work with custom builders that are not generic themselves.

stubs/common/EloquentBuilder.stub
```diff
@@ -544,7 +544,19 @@ class Builder
...
+class Scope {
+    /**
+     * Apply the scope to a given Eloquent query builder.
+     *
+     * @param  \Illuminate\Database\Eloquent\Builder<TModelClass>  $builder
+     * @param  TModelClass  $model
+     * @return void
+     */
+    public function apply(Builder $builder, Model $model);
+}
```
calebdw: This looks brittle to me:

```suggestion
/**
 * @template TBuilder of Builder
 * @template TModelClass of Model
 */
class Scope {
    /**
     * Apply the scope to a given Eloquent query builder.
     *
     * @param  TBuilder  $builder
     * @param  TModelClass  $model
     * @return void
     */
    public function apply(Builder $builder, Model $model);
}
```

Then to use:
```php
/** @implements Scope<Builder<Update|Announcement>, Update|Annoucement> */
class AnnouncementTypeScope implements Scope
{
    public function apply(Builder $builder, Model $model): void
    {
        // ...
    }
}
```

which will also work with custom builders:
```php
/** @implements Scope<AnnouncementBuilder, Annoucement> */
class AnnouncementScope implements Scope
{
    public function apply(Builder $builder, Model $model): void
    {
        // ...
    }
}
```



I'd also recommend moving to its own stub now that you're adding generics to it

func0der:
> Also note that this should work with custom builders that are not generic themselves.

Since the interface of the `Scope` requires an Eloquent-Builder to be given to `apply()` as the first parameter, Custom builders would **require** to extend the Eloquent-Builder which makes them generic. 
So there should be no need to make the $builder non generic, is there?

> Okay---to test you would need to create a custom scope that implements the generics and then assert the types.

Which folder would they go in? COuld you point me towards an example test, that I can adapt?

calebdw:
A custom Builder of the form:

```php
/** @extends Builder<Announcement> */
class AnnouncementBuilder extends Builder
{
    // ...
}
```

is not generic and would break your code.

A custom Builder of the form:

```php
/**
 * @template TModelClass of Model
 * @extends Builder<TModelClass>
 */
class PostBuilder extends Builder
{
    // ...
}
```

is generic and would still break your code.

For tests there's no external api so you're just going to have to define the custom scope and then assert the types inside the `apply` method

calebdw:
This test should be good enough:

tests/Type/data/custom-eloquent-scope.php
```php
<?php

use Illuminate\Database\Eloquent\Scope;

/** @implements Scope<Illuminate\Database\Eloquent\Builder<App\User>,App\User> */
class UserScope implements Scope
{
    public function apply($builder, $model): void
    {
        assertType('Illuminate\Database\Eloquent\Builder<App\User>', $builder);
        assertType('App\User', $model);
    }
}

/** @implements Scope<App\PostBuilder<App\Post>,App\Post> */
class PostScope implements Scope
{
    public function apply($builder, $model): void
    {
        assertType('App\PostBuilder<App\Post>', $builder);
        assertType('App\Post', $model);
    }
}
```

And then you'll have to add the file in `tests/Type/GeneralTypeTest.php` next to the other custom files:

```diff
        yield from self::gatherAssertTypes(__DIR__ . '/data/custom-eloquent-builder.php');
        yield from self::gatherAssertTypes(__DIR__ . '/data/custom-eloquent-collection.php');
+       yield from self::gatherAssertTypes(__DIR__ . '/data/custom-eloquent-scope.php');
        yield from self::gatherAssertTypes(__DIR__ . '/data/database-transaction.php');
```

func0der:
@calebdw Thanks for all the support. 
I think I managed to put something together.
Also added a test case for the custom non generic builder case you mentioned.

func0der:
I hope i got the tests down now ^^

canvural:
Hi,

I don't understand the need for this change. You can already do:
```php
class AnnouncementTypeScope implements Scope
{
    /**
     * @param Builder<Update|Announcement> $builder
     * @param Update|Announcement $model
     */
    public function apply(Builder $builder, Model $model): void
    {
        ....
    }
}

```

Writing `@implements` and getting the same types from there seems useless..

I tested this, and could not get the issue that you were getting. Can you revert all the changes and try to make a test case that fails, instead?

func0der:
@canvural You are right. My bad. 
This error comes only in combination with `phpstan/phpstan-strict-rules`, which are common for phpstan users though.

With the generic interface solution this problem is resolved. I can not get it to work without it.

I am not sure if and how I would write a test case for the combination of the two. I can prepare you are repo where you can see the same error. But I can not write a test that does fail in the `larastan` repository alone, AFAIK.

Either way. Making `Scope` generic apart from fixing this issue, also prepares us for up and coming changes to that interface from Laravel, where more methods might become generic.
Also the `Scopes` interface itself is a very generic interface that is always used in some model context anyway, AFAIK. 

So while this change might not seem per sé _required_, it is still _useful_.

func0der:
So no merge for this?

canvural:
> With the generic interface solution this problem is resolved.

This is enough information for me. I do not see any other issues we could solve by making the interface generic.

Closing it for now.

PR #1910, opened 2024-04-11
feat: add model stubs

### LINKED ISSUE #1908: `Model::$with` is not stubbed, but `Model::with()` is
- Larastan Version: 2.9.2
- `--level` used: 6

### Description

Currently the method `Model::with()` is document or better stubbed here: https://github.com/larastan/larastan/blob/2.x/stubs/common/Model.stub#L45

Yet the property `Model::$with` is not. Is there are reason for that?

As far as I see, it should be type something like this:

```php
 /**
  * @var array<array-key, string|string[]>
  */
protected $with = []
```

Is there any reason it has not yet been stubbed?

### LINKED ISSUE #1909: `Model::update()` is stubbed, but `Model::save()`  is not
- Larastan Version: 2.9.2
- `--level` used: 6

### Description

Method `Model::update()` has been stubbed: https://github.com/larastan/larastan/blob/2.x/stubs/common/Model.stub#L37

Yet `Model::save()` has not. As I see it, it is only a subset of `Model::update()` and should look a little like this:

```php
    /**
     * @param  array<int|string, mixed>  $options
     * @return bool
     */
    public function save(array $options = []);
```

Is there a reason it has not yet been stubbed?

```php

```

calebdw:
**Changes**

Hello!

This adds some stubs for Model.

Closes #1908, closes #1909 

Thanks!


szepeviktor (approved):

canvural:
Thank you!

PR #1916, opened 2024-04-14
chore: remove carbon mixin code

calebdw:
**Changes**

Hello!

This PR removes the Carbon mixin code from this package. Carbon has their own mixin system with their [own phpstan extension](https://github.com/briannesbitt/Carbon/blob/master/extension.neon) and tests---there's no need for us to override that and have unnecessary code. We actually had an issue a few months ago where *our* Carbon mixin code broke due to Carbon changes whereas if users had been using Carbon's phpstan plugin all along then that would not have been an issue.

If users use the [phpstan extension-installer](https://github.com/phpstan/extension-installer) then the Carbon plugin is automatically used, otherwise they would need to manually include it in their phpstan configs.

Thanks!

**Breaking changes**

N/A


Roardom:
> If users use the [phpstan extension-installer](https://github.com/phpstan/extension-installer) then the Carbon plugin is automatically used, otherwise they would need to manually include it in their phpstan configs.

Is there any chance you can document this? As someone who doesn't use the extension installer, this sounds like useful information

calebdw:
I added a line in the installation example

canvural:
Hi,

I think I'd leave this change to a major version. If people don't have this extension installed or don't include the Carbon extension, their builds will fail.

I don't mind breaking builds in minor/patch version because Larastan/PHPstan got smarter about some code, or we fixed a bug. But this feel different.

I'll add this to the v3 list.

calebdw:
No worries, it would be nice to have a branch that I could start making PRs against! :wink:  :upside_down_face:

PR #1920, opened 2024-04-16
chore: add test for renameColumn

### LINKED ISSUE #1919: Larastan doesn't detect a renameColumn name
- Larastan Version: v2.9.4
- Laravel Version: v10.48.7
- Level: 1

### Description

In a migration, I create a column `last_changed_at`, and in a later migration I do a `$table->renameColumn('last_changed_at', 'first_detection_at')`. `last_changed_at` isn't used anymore, but `first_detection_at` is.

But phpstan gives me an error:

> Access to an undefined property App\Models\SiteCheck::$first_detection_at.

Only that property. All the others are detected OK. It seems the `renameColumn()` doesn't work.

If I edit the migration and add in `up()`:

```php
// larastan
return;
\Schema::table('site_checks', function(Blueprint $table) {
	$table->timestamp('first_detection_at');
});
```

Larastan gets it. But that's too hacky, even for me.

If I add some debug in `SchemaAggregator`:

```php
case 'renamecolumn':
	dump($columnName, $secondArg->value);
```

and run phpstan with `--debug`, it seems Larastan does understand the renameColumn():

```
"last_changed_at" // vendor/larastan/larastan/src/Properties/SchemaAggregator.php:412
"first_detection_at" // vendor/larastan/larastan/src/Properties/SchemaAggregator.php:412
```

but it still doesn't apply the change...?

Should I read the migrations as bootstrap files or something?

<s>I'll try to upgrade Laravel and Larastan to try the newest version.</s> Same error with newest versions.

calebdw:
**Changes**

Hello!

This adds a test for `renameColumn`.

Thanks!

### DIFF
--- a/tests/Unit/MigrationHelperTest.php
+++ b/tests/Unit/MigrationHelperTest.php
@@ -112,7 +112,7 @@ class MigrationHelperTest extends PHPStanTestCase
     }
 
     /** @test */
-    public function it_can_handle_alter_table_rename(): void
+    public function it_can_handle_alter_table_and_column_rename(): void
     {
         $migrationHelper = new MigrationHelper($this->parser, [
             __DIR__ . '/data/rename_migrations',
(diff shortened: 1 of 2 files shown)


canvural:
Hi,

Thanks for the test. But I removed the issue closing keyword. We'll keep discussing that in the issue.

PR #1921, opened 2024-04-17
refactor: assertType calls

calebdw:
Hello!

This PR goes through and refactors all the `assertType` tests for consistency in the codebase:

- all `assertType` calls are now inside a function instead of in the top level
- the name for all these functions is `test`, I felt this was a better representation of what it actually is for than `foo`/`dummy`
  - where it makes sense, some calls are inside a `test<Description>` method 
- any types needed inside the function are passed as parameters

Thanks!


canvural:
Thank you!

PR #1929, opened 2024-04-19
fix: eloquent collection generics

### LINKED ISSUE #1927: Larastan doesn't detect custom model collection after `map()`
- Larastan Version: 2.9.5
- PHPStan Version: 1.10.67
- Laravel Version: 10.48.8

### Description

My `User` model overrides `newCollection()` to return custom `ModelCollection` that extends `Eloquent\Collection`. The user model has a relationship `created_by_user` to `User`. I want to fetch a few users, and map them to their creator users. The fetched users are a `ModelCollection` 👍 I expect the resulting collection to also be a `ModelCollection` because it's still a collection of models. Instead PHPStan thinks it's a Eloquent\Collection.

### Code

```php
$models = User::query()->get();
\PHPStan\dumpType($models); // 👍 App\Base\ModelCollection<int, App\Models\User>
$users = $models->map(fn(User $model) => $model->created_by_user);
\PHPStan\dumpType($users); // 👎 Illuminate\Database\Eloquent\Collection<int, App\Models\User>
```

If I make the closure more verbose:

```
$users = $models->map(function(User $model) : User {
    return $model->created_by_user;
});
```

the result is the same, because the `User` bit isn't the problem (Larastan knows the result is a collection of User), but the collection is.

### Demo

`MapController` in https://github.com/rudiedirkx/larastan-demo

It tests with `User` and `Post`. `User` has the custom collection, `Post` doesn't. The `Post` map result is fine, but the `User` map result is not.

    $ ./vendor/bin/phpstan analyze app/Http/Controllers/MapController.php

```
  16     Dumped type: Illuminate\Database\Eloquent\Collection<int, App\Models\Post>
  18     Dumped type: Illuminate\Database\Eloquent\Collection<int, App\Models\User>
  24     Dumped type: App\Base\ModelCollection<int, App\Models\User>
  28     Dumped type: Illuminate\Database\Eloquent\Collection<int, App\Models\User>
```

calebdw:
- [x] Added or updated tests

Closes #1927 

**Changes**

Hello!

This PR corrects EloquentCollection generics and issues with custom EloquentClasses.

With these changes, a collection can only be an EloquentCollection if Model can be a super type of all the inner types in the collection. If this is not the case, then the collection class is downgraded to the SupportCollection.

Furthermore, if you have a custom collection that narrows the `TModel` generic, and that collection has a wider type, then the collection is downgraded to an EloquentCollection:

```php
/** @extends EloquentCollection<int, User> */
class UserCollection extends EloquentCollection
// ..
$userCollection; // this has type UserCollection
$foo = $userCollection->map(fn ($u) => new Role()); // now a EloquentCollection<int, Role>

```

Thanks!

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -299,7 +299,7 @@ services:
             - phpstan.broker.dynamicMethodReturnTypeExtension
 
     -
-        class: Larastan\Larastan\ReturnTypes\EnumerableGenericStaticMethodDynamicMethodReturnTypeExtension
+        class: Larastan\Larastan\ReturnTypes\EnumerableGenericDynamicMethodReturnTypeExtension
         tags:
             - phpstan.broker.dynamicMethodReturnTypeExtension
 
@@ -314,7 +314,7 @@ services:
             - phpstan.broker.dynamicMethodReturnTypeExtension
 
     -
-        class: Larastan\Larastan\ReturnTypes\EnumerableGenericStaticMethodDynamicStaticMethodReturnTypeExtension
+        class: Larastan\Larastan\ReturnTypes\EnumerableGenericDynamicStaticMethodReturnTypeExtension
         tags:
             - phpstan.broker.dynamicStaticMethodReturnTypeExtension
 
--- a/src/ReturnTypes/EnumerableGenericDynamicMethodReturnTypeExtension.php
+++ b/src/ReturnTypes/EnumerableGenericDynamicMethodReturnTypeExtension.php
@@ -26,7 +26,7 @@ use function array_map;
 use function array_merge;
 use function in_array;
 
-class EnumerableGenericStaticMethodDynamicMethodReturnTypeExtension implements DynamicMethodReturnTypeExtension
+class EnumerableGenericDynamicMethodReturnTypeExtension implements DynamicMethodReturnTypeExtension
 {
     public function getClass(): string
     {
--- a/src/ReturnTypes/EnumerableGenericDynamicMethodReturnTypeExtension.php
+++ b/src/ReturnTypes/EnumerableGenericDynamicMethodReturnTypeExtension.php
@@ -15,6 +15,7 @@ use PHPStan\Reflection\MethodReflection;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
 use PHPStan\Type\Generic\GenericObjectType;
+use PHPStan\Type\Generic\TemplateObjectType;
 use PHPStan\Type\IntersectionType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\StringType;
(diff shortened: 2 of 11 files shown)


canvural:
Hi,

Did not see this PR. Fixed the same issue with 1a05ac6dd0bbde4dc52b9d24c0a2a493f2c3784c

calebdw:
@canvural, this PR solves other issues with eloquent collection than just the one in the issue.

Instead of returning early with EloquentCollections or duplicating methods I was able to  just add the `find` method at the end similar to the SupportCollection.

```php
        if ($methodReflection->getDeclaringClass()->getName() === EloquentCollection::class) {
            $methods = array_merge($methods, ['find']);
        }
```

This PR:
1. Correctly downgrades an EloquentCollection to SupportCollection if the collection does not contain Models
2. Correctly downgrades a custom collection to EloquentCollection if it contains models that do not fit into the defined generics.

Please reconsider this PR, I believe it to be the more complete solution and solves other issues than just with the `map` methods.

canvural:
I checked the test cases and they are wrong. They don't represent the runtime types.

calebdw:
@canvural, the runtime types are incorrect and need to be fixed---I plan on submitting PRs to the framework to tackle that.

The test cases here are correct and safe---you cannot call EloquentCollection methods on a collection that does not contain Models, and a non-generic custom collection cannot contain anything outside the given generics

canvural:
> you cannot call EloquentCollection methods on a collection that does not contain Models

Where does that constraint coming from?

calebdw:
It's unsafe, how can you call methods like this if you're not certain that every item in the collection is a Model?

- https://github.com/laravel/framework/blob/134db76b341b7e69a57a833956c8fd6857882206/src/Illuminate/Database/Eloquent/Collection.php#L66
- https://github.com/laravel/framework/blob/134db76b341b7e69a57a833956c8fd6857882206/src/Illuminate/Database/Eloquent/Collection.php#L314

The map methods even convert `toBase` if the collection contains anything that's not a model:

https://github.com/laravel/framework/blob/134db76b341b7e69a57a833956c8fd6857882206/src/Illuminate/Database/Eloquent/Collection.php#L334-L347

More methods just need to do this, or just follow a similar pattern as `countBy` which first converts to base before calling the same method:

https://github.com/laravel/framework/blob/134db76b341b7e69a57a833956c8fd6857882206/src/Illuminate/Database/Eloquent/Collection.php#L565-L574

canvural:
We cannot do that change first. If Laravel changes it, then we can also make the change.

calebdw:
> We cannot do that change first.

Why not? The actual Collection generics indicate that the values must always be a Model:
> @template TModel of \Illuminate\Database\Eloquent\Model
https://github.com/laravel/framework/blob/134db76b341b7e69a57a833956c8fd6857882206/src/Illuminate/Database/Eloquent/Collection.php#L13-L19

Therefore an EloquentCollection *should be* downgraded to a SupportCollection if it violates this---regardless of what the runtime types are.

You can pass an EloquentCollection to a method that is typed as SupportCollection without issue so this ensures type safety despite the runtime types being incorrect.

PR #1934, opened 2024-04-22
[v1] fix: removing composer dependency

CoolGoose:
See: https://github.com/composer/composer/issues/11940

**Changes**

This removes the composer dependency of the package itself.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,16 @@
     "name": "nunomaduro/larastan",
     "type": "phpstan-extension",
     "description": "Larastan - Discover bugs in your code without running it. A phpstan/phpstan wrapper for Laravel",
-    "keywords": ["laravel", "larastan", "phpstan", "php", "package", "static analysis", "code analysis", "code analyse"],
+    "keywords": [
+        "laravel",
+        "larastan",
+        "phpstan",
+        "php",
+        "package",
+        "static analysis",
+        "code analysis",
+        "code analyse"
+    ],
     "license": "MIT",
     "authors": [
         {
@@ -40,9 +49,9 @@
             "Database\\Factories\\": "tests/Laravel8Application/database/factories/",
             "Database\\Migrations\\": "tests/Laravel8Application/database/migrations/"
         },
-      "classmap": [
-        "tests/"
-      ]
+        "classmap": [
+            "tests/"
+        ]
     },
     "minimum-stability": "dev",
     "prefer-stable": true,
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,6 @@
     "require": {
         "php": "^7.2 || ^8.0",
         "ext-json": "*",
-        "composer/composer": "^1.0 || ^2.0",
         "illuminate/console": "^6.0 || ^7.0 || ^8.0 || ^9.0",
         "illuminate/container": "^6.0 || ^7.0 || ^8.0 || ^9.0",
         "illuminate/contracts": "^6.0 || ^7.0 || ^8.0 || ^9.0",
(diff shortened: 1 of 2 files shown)


szepeviktor:
Hello @CoolGoose 👋🏻 

We don't accept changes to v1.
I am sorry.

CoolGoose:
@szepeviktor I think this should be an exception since it does break normal usage of composer > 2.7.3 because of the linked bug.

If that's still the case and you don't accept exceptions it's fine and I'll just fork it for our internal use only, but please don't block a lot of users of 1.x on never updating composer because of it.

szepeviktor:
We use Composer
https://github.com/larastan/larastan/blob/769bc6346a6cce3b823c30eaace33d9c3a0dd40e/src/ApplicationResolver.php#L96-L99

szepeviktor:
> please don't block a lot of users of 1.x

You can install Larastan this way: `composer require "larastan/larastan:^1.0" --with="composer/composer<=2.7.3"`

szepeviktor:
Or by adding `"composer/composer": "<=2.7.3",` to your composer.json.

CoolGoose:
But technically that does mean that the dependency should be `composer/class-map-generator` not `composer` itself no ? 
I'm happy to give it a try and test.

CoolGoose:
I think this might be fine now @szepeviktor , I am just not sure about what should I do with the tests, since locally at least it errors out, but not in parts of the codebase that have a relation with this changes.

I did check and the ClassMapGenerator does return data

```php
array(20) {
  [0]=>
  string(16) "App\UserResource"
  [1]=>
  string(9) "App\Admin"
  [2]=>
  string(10) "App\Thread"
  [3]=>
  string(30) "App\Providers\ConcreteProvider"
  [4]=>
  string(30) "App\Providers\AbstractProvider"
  [5]=>
  string(8) "App\User"
  [6]=>
  string(15) "App\Transaction"
  [7]=>
  string(25) "App\TransactionCollection"
  [8]=>
  string(12) "App\Importer"
  [9]=>
  string(11) "App\Address"
  [10]=>
  string(8) "App\Post"
  [11]=>
  string(18) "App\RoleCollection"
  [12]=>
  string(8) "App\Role"
  [13]=>
  string(21) "App\AccountCollection"
  [14]=>
  string(19) "App\Traits\HasOwner"
  [15]=>
  string(20) "App\Traits\HasParent"
  [16]=>
  string(28) "App\Traits\NestedSoftDeletes"
  [17]=>
  string(11) "App\Account"
  [18]=>
  string(9) "App\Group"
  [19]=>
  string(16) "App\GuardedModel"
}
```

szepeviktor:
We do not work on v1.
https://github.com/larastan/larastan/commits/1.x/

CoolGoose:
The change is against the 1.x branch .
To fix 2.x you want me to cherry pick and re-apply on 2.x ? 

I assume for the time-being for 1.x people should just fork and apply the same patch ?

szepeviktor:
You find all I can say in [the above comments](https://github.com/larastan/larastan/pull/1934#issuecomment-2068855431).

CoolGoose:
I respect your decision as the maintainer to not merge against 1.x, but that's not a valid fix, it's more of a crutch workaround.

It does seem composer itself might try to fix this so people don't have the issue with newer composer versions, but in any case my plans are the following

1. Open a changeset against 2.x to remove this going forward for new installs on the maintained codebase
2. Fork and do a new 1.x release for who needs it.

canvural:
We don't have a dependency on Composer in 2.x It was removed some time ago. 

And like @szepeviktor said 1.x is not maintained any more. Although I understand your point, I think it's normal to encounter issues if you use older or not anymore maintained versions of any software.

szepeviktor:
@CoolGoose You've got your fix: https://github.com/composer/composer/releases/tag/2.7.4

PR #1939, opened 2024-04-29
Add model timestamps as array

tvup:
- [x] Added or updated tests

**Changes**
This PR enables editing the timestamps property on an Eloquent model. This property can be set to a boolean value—true if both created_at and updated_at are used, or false if they are not. Additionally, developers can specify either created_at, updated_at, both, or neither by providing an array. Laravel will handle these settings appropriately. This PR introduces the capability to accept an array for the timestamps property.

**Breaking changes**
None to my knowledge

### DIFF
--- a/stubs/common/Database/Eloquent/HasTimestamps.stub
+++ b/stubs/common/Database/Eloquent/HasTimestamps.stub
@@ -1,0 +1,14 @@
+<?php
+
+namespace Illuminate\Database\Eloquent\Concerns;
+
+trait HasTimestamps
+{
+    /**
+     * Indicates if the model should be timestamped,
+     * or includes the timestamp fields which should be used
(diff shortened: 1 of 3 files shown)


calebdw:
The `$timestamps` property is only ever used as a bool in the framework---setting it to be an array would be truthy which is the only reason why that would work.

Also, this is just a phpdoc fix---something like this should be attempted in the framework first so there's no need for a stub.

szepeviktor:
Yes, this PR might not get merged.
I am sorry.

https://github.com/laravel/framework/blob/e38d945fce91860c29f10a6bd4a19790ff25ad84/src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php#L9-L14

PR #1942, opened 2024-05-05
fix: infinite mixin recursion through class_alias

### LINKED ISSUE #1883: Redis::connection causes segmentation fault
- Larastan Version: 2.9.2
- `--level` used: 6
- Pull request with failing test: N/A private code

### Description
Interacting with Redis::connection causes segmentation fault

This seems to be related to https://github.com/phpstan/phpstan/issues/6300 which was fixed in 
https://github.com/phpstan/phpstan-src/commit/9ab4e6fd8ef29d7bf7b0b729250e04f3dd5ad75d

Running tests with just phpstan pass fine, but with the larastan extension included it causes a segmentation fault anywhere that `Redis::connection` is used.

I'm using phpstan 1.10.62

```php
        $redis = Redis::connection('session');
        $redis->del($basket->redisName . ':' . Session()->get('accountUser')->id);
```

sdkawata:
- [x] Added or updated tests

Resolves #1883

**Changes**

In MixinPropertiesClassReflectionExtension, there's a mechanism to prevent infinite mixin recursion. However, within ReflectionHelper, because a new instance of MixinPropertiesClassReflectionExtension is created each time, it wasn't functioning correctly.
In https://github.com/larastan/larastan/pull/1839, an attempt was made to prevent infinite recursion where a class is mixin is applied to itself, but it failed to prevent infinite recursion when mixin was applied via class_alias. This PR addresses this issue by modifying the code to reuse MixinPropertiesClassReflectionExtension instances instead of creating new ones each time, thus effectively preventing recursion over a wider range of mixin scenarios.

**Breaking changes**

Probably No

### DIFF
--- a/src/Reflection/ReflectionHelper.php
+++ b/src/Reflection/ReflectionHelper.php
@@ -12,6 +12,8 @@ use function array_key_exists;
 
 final class ReflectionHelper
 {
+    private static MixinPropertiesClassReflectionExtension|null $mixinPropertiesClassReflectionExtension = null;
+
     /**
      * Does the given class or any of its ancestors have an `@property*` annotation with the given name?
      */
@@ -27,10 +29,17 @@ final class ReflectionHelper
             }
         }
 
+        if (self::$mixinPropertiesClassReflectionExtension === null) {
+            /** @phpstan-ignore-next-line */
+            self::$mixinPropertiesClassReflectionExtension = (new MixinPropertiesClassReflectionExtension([]));
+        }
+
         /** @phpstan-ignore-next-line */
-        return (new MixinPropertiesClassReflectionExtension([$classReflection->getName()]))
+        return self::$mixinPropertiesClassReflectionExtension
             ->hasProperty($classReflection, $propertyName);
     }
+
+    private static MixinMethodsClassReflectionExtension|null $mixinMethodsClassReflectionExtension = null;
 
     /**
      * Does the given class or any of its ancestors have an `@method*` annotation with the given name?
@@ -47,8 +56,13 @@ final class ReflectionHelper
             }
         }
 
+        if (self::$mixinMethodsClassReflectionExtension === null) {
+            /** @phpstan-ignore-next-line */
+            self::$mixinMethodsClassReflectionExtension = new MixinMethodsClassReflectionExtension([]);
+        }
+
         /** @phpstan-ignore-next-line */
-        return (new MixinMethodsClassReflectionExtension([$classReflection->getName()]))
+        return self::$mixinMethodsClassReflectionExtension
             ->hasMethod($classReflection, $methodName);
     }
 }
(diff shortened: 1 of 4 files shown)


calebdw:
Why would a class mixin an alias of itself?

sdkawata:
> Why would a class mixin an alias of itself?

`\Illuminate\Support\Facades\Redis` mixin `\Redis`.
https://github.com/larastan/larastan/blob/66db15f82209d971e9bcca99ea3b2cb031be9d52/stubs/common/Facades.stub#L8

And,  `\Redis` is an alias of `\Illuminate\Support\Facades\Redis`.
https://github.com/laravel/laravel/blob/3923e7f7c40368a5e78c1a33610191be8ad91e3b/config/app.php#L218

This recursion cause https://github.com/larastan/larastan/issues/1883

calebdw:
I'm fairly certain this is (or rather should be) mixing in the root Redis class/client from phpredis and not the alias:

https://github.com/laravel/framework/blob/5710fdb8bf36df501405c705f7d2ea4720d2b9d3/src/Illuminate/Redis/Connections/Connection.php#L18-L23

There's no reason for a class to mixin an alias to itself

sdkawata:
While I agree that this intended to be mixing the Redis class from phpredis,
this unintentionally be mixing an alias in the environment where phpredis is not installed,
and larastan crashes in such an environment.
This happens for laravel default installation (-7.0), thus this fix may benefit many users.

This PR not only fixes the issue of infinite mixin recursion through class_alias
but also prevents infinite mixin recursion in other scenarios (such as #1830).
Therefore, it may benefit users who might encounter similar issues under different circumstances.

canvural:
Hi,

I was suspecting this would not cause the errors you mentioned and I removed your changes for `ReflectionHelper` but kept the tests and tests still passed 🙂 Also this test might be more suitable to `Integration` folder.

So first please try to create the error in the CI then we can talk about the fix.

sdkawata:
> I was suspecting this would not cause the errors you mentioned and I removed your changes for ReflectionHelper but kept the tests and tests still passed 🙂 Also this test might be more suitable to Integration folder.

I moved the test to `Integration` folder and confirmed that the added test fails in CI,

canvural:
> and confirmed that the added test fails in CI,

It did not fail for the correct reason 😄 Looks like it's a memory limit issue in PHPUnit. You can try to increase the memory limit for PHPUnit or set it to `-1` And please just keep the failing test. I want to see the same segmentation fault error in the CI that was mentioned in #1883

calebdw:
Note that this is also just testing an unknown class on alias mixin---this is not testing the specific code that caused the  segfault:

https://github.com/larastan/larastan/issues/1883#issuecomment-1999647373

canvural:
Hi,

I just remembered that there is an undocumented PHPStan config option just for this case! It's called `mixinExcludeClasses` So you can add the following to your `phpstan.neon` file
```neon
mixinExcludeClasses:
	Redis
```

and it should solve the problem 👍🏽

PR #1944, opened 2024-05-08
Allow Backed Enums to be used for the auth provider

lancepioch:
- [x] Added or updated tests

**Changes**

If you are using a BackedEnum for the auth provider, this resolves issues with that.

**Breaking changes**

There are no BC

### DIFF
--- a/src/Concerns/LoadsAuthModel.php
+++ b/src/Concerns/LoadsAuthModel.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Larastan\Larastan\Concerns;
 
+use BackedEnum;
 use Illuminate\Contracts\Config\Repository as ConfigRepository;
 
 use function array_keys;
@@ -32,7 +33,7 @@ trait LoadsAuthModel
                     return $carry;
                 }
 
-                if ($provider instanceof \BackedEnum) {
+                if ($provider instanceof BackedEnum) {
                     $provider = $provider->value;
                 }
 
--- a/src/Concerns/LoadsAuthModel.php
+++ b/src/Concerns/LoadsAuthModel.php
@@ -32,6 +32,10 @@ trait LoadsAuthModel
                     return $carry;
                 }
 
+                if ($provider instanceof \BackedEnum) {
+                    $provider = $provider->value;
+                }
+
                 $authModel = $providers[$provider]['model'] ?? null;
 
                 if (! $authModel || in_array($authModel, $carry, strict: true)) {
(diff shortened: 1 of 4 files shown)


calebdw:
What is a valid use case for this and what does the config look like?

canvural:
Could you add a test case that shows how this can be used?

lancepioch:
Yeah, let me take a look at the test suite and see!

lancepioch:
@canvural updated the tests. My use case is that the app has many different guards and we organize it via an enum instead of having the magic strings everywhere.

calebdw:
@lancepioch,

Given the following config:

```php
return [
        // ...
        'enum' => [
            'driver' => 'session',
            'provider' => App\ProviderEnum::Users,
        ],
    ],

    'providers' => [
        'users' => [
            'driver' => 'eloquent',
            'model' => App\User::class,
        ],
       // ...
```

The `provider` value has to match the key in the `providers` array, therefore unless you're also using the enum as the key in the providers array you're going to have an inconsistent definition and you're still relying on "magic strings".

```diff
return [
        // ...
        'enum' => [
            'driver' => 'session',
            'provider' => App\ProviderEnum::Users,
        ],
    ],

    'providers' => [
-        'users' => [
+       App\ProviderEnum::Users->value => [
            'driver' => 'eloquent',
            'model' => App\User::class,
        ],
       // ...
```

Note that you can also just use the following and no code changes to Larastan would be required:



```diff
return [
        // ...
        'enum' => [
            'driver' => 'session',
-           'provider' => App\ProviderEnum::Users,
+           'provider' => App\ProviderEnum::Users->value,
        ],
    ]
```

lancepioch:
> The provider value has to match the key in the providers array, therefore unless you're also using the enum as the key in the providers array you're going to have an inconsistent definition and you're still relying on "magic strings".

That's correct, we are and we're also using the enum in several other parts of the code instead of using "->value" everywhere.

calebdw:
Where else in the code are the `providers` keys used? To my understanding, it is only referenced in the guard `provider` definition---and the guard name does not have to match the provider name (we have different guards that use the same provider).

Does the Laravel framework actually support using an enum as the provider value? To me this seems like such a niche (and quite frankly, useless) edge case given that the *providers names are only referenced in the same config file as the guards*.

Now I might could see using an enum for the **guard** name as those are used around the codebase in the `::guard()` methods and `auth:` middleware, but this PR is solely focused on the provider and not the guard

canvural:

PR #1950, opened 2024-05-16
feat: convert array-key to int|string for collection keys

### LINKED ISSUE #1794: Enumerable template is not used for search method
- Larastan Version: v2.6.4
- `--level` used: 8

### Description

Can't infer key type from phpdoc and it throws following error:

```
Parameter #1 $items of method Illuminate\Support\Enumerable<int,Modules\Shipment\Entities\ShipmentProduct>::replace() expects  
         Illuminate\Contracts\Support\Arrayable<int, Modules\Shipment\Entities\ShipmentProduct>|iterable<int,                           
         Modules\Shipment\Entities\ShipmentProduct>, non-empty-array<int|string, Modules\Shipment\Entities\ShipmentProduct> given.
```

```php
/**
 * @property int $id
 * @property Enumerable<int, ShipmentProduct> $products
 */
class Shipment extends Model {
    public function add(ShipmentProduct $shipmentProduct) {
            $index = $this->products->search(fn (ShipmentProduct $sp) => $sp->id === $shipmentProduct->id);

            if ($index === false) {
                throw new \Exception();
            }

            $this->products->replace([$index => $shipmentProduct]);
    }
}
```

If I explicitly set $index type it will be alright:

```
 /** @var int|false $index*/
```

Enumerable `search` method uses template and returns `TKey`. Don't know why it doesn't work automatically.

```php
/**
     * Search the collection for a given value and return the corresponding key if successful.
     *
     * @param  TValue|callable(TValue,TKey): bool  $value
     * @param  bool  $strict
     * @return TKey|bool
     */
    public function search($value, $strict = false);
```

canvural:
- [x] Added or updated tests

**Changes**

PHPStan has a bug around creating union types with benevolent union types. `array-key` in this case. So this PR changes `array-key` to `int|string` which should be the same behavior.

Fixes #1794 

**Breaking changes**

### DIFF
--- a/stubs/common/Collection.stub
+++ b/stubs/common/Collection.stub
@@ -7,7 +7,7 @@ use Illuminate\Contracts\Support\CanBeEscapedWhenCastToString;
 use Illuminate\Support\Traits\EnumeratesValues;
 
 /**
- * @template TKey of array-key
+ * @template TKey of int|string
  * @template TValue
  *
  * @implements \ArrayAccess<TKey, TValue>
@@ -59,7 +59,7 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     /**
      * Push all of the given items onto the collection.
      *
-     * @template TConcatKey of array-key
+     * @template TConcatKey of int|string
      * @template TConcatValue
      *
      * @param  iterable<TConcatKey, TConcatValue>  $source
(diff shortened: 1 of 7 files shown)


szepeviktor:
Definitely an improvement: https://github.com/larastan/larastan/actions/runs/9119477663/job/25074914522?pr=1950

PR #1953, opened 2024-05-20
fix: child builders with non generic parents

### LINKED ISSUE #1952: Generic parent methods erroneously treated as private
- Larastan Version: 2.9.6
- Laravel Version: 11.7.0

### The Surface Problem

I've created an [example project](https://github.com/treed-eaglemetal/larastan-builder-generics) to replicate this issue. In short, I'm faced with the following error:

> Call to private method orderBy() of parent class App\Builders\UserBuilder.

My `UserBuilder` class extends `Illuminate\Database\Eloquent\Builder`, where the `orderBy` method is sourced from the `Illuminate\Database\Query\Builder` class via mixin. Anyone familiar with Laravel will know that the `orderBy` method is not private.

The code in question being blamed is also seemingly harmless:

```php
public function index(Request $request): View
{
    $users = PortalUser::query()
        ->whereType($request->type)
        ->orderBy('name') // <-- Blame
        ->get();

    return view('users.index', compact('users'));
}
```

Without knowing the setup behind the scenes to cause this, the error looks like a fluke. This is why I took my working problem, and created the example project to ensure the issue isn't related to my specific setup.

### The Deeper Problem

This happens when generics are involved in a particular manner. To simplify the example project even further, here are the docblocks for the involved classes:

[Larastan Eloquent Stub](https://github.com/larastan/larastan/blob/2.x/stubs/common/EloquentBuilder.stub):
```php
/**
 * @template TModelClass of Model
 * @property-read static $orWhere
 */
class Builder
```

[Example Parent Builder](https://github.com/treed-eaglemetal/larastan-builder-generics/blob/main/app/Builders/UserBuilder.php):
```php
/** @extends Builder<User> */
```

[Example Child Builder](https://github.com/treed-eaglemetal/larastan-builder-generics/blob/main/app/Builders/PortalUserBuilder.php):
```php
class PortalUserBuilder extends UserBuilder
```

Note that the child builder has no generics. The parent builder references generics, but is not itself generic.

When this specific combination is in play, any method call to the underlying Database Query Builder instance is erroneously treated as private.

### The Likely Culprit

After digging around through Larastan's source code, I came across [this](https://github.com/larastan/larastan/blob/f934f35d51d120d650a678433d9e914353623ae0/src/Methods/EloquentBuilderForwardsCallsExtension.php#L74-L79):

```php
// Generic type is not specified
if ($modelType === null) {
    if (! $classReflection->isGeneric() && $classReflection->getParentClass()?->isGeneric()) {
        $modelType = $classReflection->getParentClass()->getActiveTemplateTypeMap()->getType('TModelClass');
    }
}
```

If I'm reading this right, when the class itself is not generic, _and the parent itself isn't generic either_, the `$modelType` stays `null`. However, had this code checked the grandparent, it would have found `Illuminate\Database\Eloquent\Model` (as the grandparent is the Eloquent Builder class).

I'm not saying that this code exactly is to blame, but this reads like Larastan expects custom eloquent builders to be a direct descendant of a generic eloquent builder. In the example scenario I've laid out, that isn't true, and I start getting nonsensical output (such as public methods being treated as private).

### The Workaround

I can definitely work around this bug by passing the generics through the parent builder, such that the child builder must reference the generics. This works, but has limitations.

In my actual scenario at work, the parent builder is defined within a package, and is therefore not open for modification. When my hands are tied in this manner, I'm left with either grandiose modifications of a package, or ignoring the private method error.

I would prefer that Larastan understand this sort of situation. While niche, it does happen.

calebdw:
- [x] Added or updated tests

**Changes**

Closes #1952 

Hello!

This PR updates the `EloquentBuilderForwardsCallsExtension` to continue iterating up the builder parents to find the active model template instead of just checking the immediate parent.

Thanks!

**Breaking changes**

N/A

### DIFF
--- a/src/Methods/EloquentBuilderForwardsCallsExtension.php
+++ b/src/Methods/EloquentBuilderForwardsCallsExtension.php
@@ -69,14 +69,17 @@ final class EloquentBuilderForwardsCallsExtension implements MethodsClassReflect
             return null;
         }
 
-        $modelType = $classReflection->getActiveTemplateTypeMap()->getType('TModelClass');
+        $loopReflection = $classReflection;
 
-        // Generic type is not specified
-        if ($modelType === null) {
-            if (! $classReflection->isGeneric() && $classReflection->getParentClass()?->isGeneric()) {
-                $modelType = $classReflection->getParentClass()->getActiveTemplateTypeMap()->getType('TModelClass');
+        do {
+            $modelType = $loopReflection->getActiveTemplateTypeMap()->getType('TModelClass');
+
+            if ($modelType !== null) {
+                break;
             }
-        }
+
+            $loopReflection = $loopReflection->getParentClass();
+        } while ($loopReflection !== null);
 
         if ($modelType === null) {
             return null;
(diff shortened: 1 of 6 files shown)


calebdw:
The e2e failure is unrelated, I'll try to open a PR to fix later tonight

canvural:
Thank you!

PR #1955, opened 2024-05-21
fix: e2e tests

calebdw:
**Changes**

Hello!

This should fix the e2e tests---I also scheduled the e2e tests, but I can remove this if it's not desired.

Thanks!

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -12,6 +12,8 @@ on:
     paths-ignore:
       - "**.md"
       - 'docs/**'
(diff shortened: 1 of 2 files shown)


canvural:
Actually I'm thinking to revert the commit that caused this. Need to make some more tests first.

But the error in `larastan-test` repo is interesting. It should not be ignored and fixed instead.

calebdw:
Ah, I did not catch that it was a macro

canvural:
I updated commits and fixed all the e2e tests. Closing this then. 


Thank you.

PR #1957, opened 2024-05-24
feat: Infer type of auth()->authenticate() and Auth::authenticate() calls

robchett:
- [x] Added or updated tests

**Changes**

Add return type detection to 

- `Auth::authenticate()`
- `auth()->authenticate()`

**Breaking changes**


tests/Type/GeneralTypeTest.php
calebdw, line 1: All of these removals look to be on accident?

robchett: Whoops! Undone. Was trying to debug `Auth::guard('x')->authenticate()` before realising that Laravel doesn't have that method at a guard level.

tests/Type/data/auth.php
```diff
@@ -20,4 +20,5 @@ function test(User $user): void
     assertType('App\Admin|null', Auth::guard('admin')->user());
     assertType(SessionGuard::class, Auth::guard('session'));
     assertType('null', Auth::guard()->login($user));
+    assertType('App\Admin|App\User', Auth::authenticate());
```
calebdw: Can you add tests for the guards?

```suggestion
    assertType('App\Admin|App\User', Auth::authenticate());
    assertType('App\User', Auth::guard('web')->authenticate());
    assertType('App\Admin', Auth::guard('admin')->authenticate());
```

tests/Type/GeneralTypeTest.php
calebdw: The `authenticate` method comes from the `GuardHelpers` trait which is applied to the guards (at least on 11.x).

src/ReturnTypes/AuthExtension.php
```diff
@@ -31,7 +32,7 @@ public function getClass(): string
 
     public function isStaticMethodSupported(MethodReflection $methodReflection): bool
     {
-        return $methodReflection->getName() === 'user';
+        return in_array($methodReflection->getName(), ['user', 'authenticate']);
```
calebdw: Just a nit, but I generally prefer to pass `strict: true`/`true` to my `in_array` calls (`===` instead of `==` equivalent)---up to you though


```suggestion
        return in_array($methodReflection->getName(), ['user', 'authenticate'], strict: true);
```

calebdw (approved):

tests/Type/data/auth.php
robchett: I'll revisit this one as `Auth::guard('web')` has the type `Illuminate\Contracts\Auth\StatefulGuard` right now which doesn't have the method in the interface so phpstan doesn't provide any way to bind a ReturnType extension. 
If I improve the return type of `Auth::guard('web')` to the concrete implementation then I should be able to attach a return type to `Auth::guard('web')->authenticate()`

calebdw: Ah that makes sense. Laravel has a really bad habit of typing interfaces and then proceeding to define and use extra methods on the runtime implementation that don't adhere to the contract---very frustrating from a static analysis point of view.

robchett: All sorted, there was a bug in the return type provider of `Auth::guard($x)`, it was using `$x` as the driver, not what was set in the config.

Tweaked a couple of bits further so the same applies to `auth($x)` as well.

robchett:
@calebdw how would you deal with the change from 
 `@method static \Illuminate\Contracts\Auth\Authenticatable|bool loginUsingId(mixed $id, bool $remember = false)`
to 
`@method static \Illuminate\Contracts\Auth\Authenticatable|false loginUsingId(mixed $id, bool $remember = false)`
in different versions of laravel?

src/ReturnTypes/AuthManagerExtension.php
```diff
@@ -49,11 +68,66 @@ public function getTypeFromMethodCall(
...
+            return $defaultReturnType;
+        }
+
+        return $this->findTypeFromGuardDriver($driver) ?? $defaultReturnType;
+    }
+
+    private function findTypeFromGuardDriver(string $driver): Type|null
+    {
+        return match ($driver) {
+            'session' => new ObjectType('Illuminate\Auth\SessionGuard'),
```
calebdw: This can also be a `::class` string?

calebdw:
> @calebdw how would you deal with the change from `@method static \Illuminate\Contracts\Auth\Authenticatable|bool loginUsingId(mixed $id, bool $remember = false)` to `@method static \Illuminate\Contracts\Auth\Authenticatable|false loginUsingId(mixed $id, bool $remember = false)` in different versions of laravel?


It depends, if you need to correct the types then you would need to create new versioned stubs (look in the `stubs/10.0.0` dir for examples).

If the types are correct and you're just having issues with asserting the type returned from a method, then you'll have to only include your assertion files for certain versions. For example, in the GeneralTypeTest:

image

src/ReturnTypes/AuthManagerExtension.php
robchett: Yes, I hadn't spotted that when moving it from src/ReturnTypes/GuardDynamicStaticMethodReturnTypeExtension.php

robchett:
Thanks @calebdw, I'll keep that in mind for next time. I think I'll leave it as is as there isn't any larastan logic applied to loginById

canvural:

PR #1958, opened 2024-05-25
Add support for `whereNot` and `orWhereNot`

ganyicz:
✅ Added or updated tests
✅ Documented user facing changes

**Changes**

This PR adds support for all HigherOrderBuilder proxies. Currently Larastan only supports `orWhere` but there are two more: https://github.com/laravel/framework/blob/e2d55af66635941d931b8e89af17553625c9699d/src/Illuminate/Database/Eloquent/Builder.php#L26-L28

The change was fairly simple, I followed the previous addition of `orWhere` from 2021: https://github.com/larastan/larastan/pull/884

**Breaking changes**

No breaking changes as far as I know


canvural:
Thank you!

PR #1959, opened 2024-05-25
feat: better generic model property type

canvural:
- [ ] Added or updated tests

**Changes**

**Breaking changes**


calebdw:
@canvural, 

I'm just curious, what all does this do and how is it better?

markuskooche:
> @canvural,
>
> I'm just curious, what all does this do and how is it better?

Here is a simple example with a factory class. Normally the code would look something like this:

```php
/**
 * @extends Factory<User>
 */
class UserFactory extends Factory
{
    /**
     */
    #[Override]
    public function definition(): array
    {
        return [
            'name' => fake()->name(),
            'email' => fake()->unique()->safeEmail(),
            'password' => Str::random(32),
        ];
    }
}
```

When I now run larastan with the new version `2.9.8`, I get the following error message:

```
Return type (array<string, mixed>) of method Database\Factories\UserFactory::definition() should be compatible with return type (array<model property of App\Models\User, mixed>) of method Illuminate\Database\Eloquent\Factories\Factory<App\Models\User>::definition()                                                                                                  
```

Now I can adapt the code accordingly. This improvement allows you to check model properties. Here is an example:

```php
/**
 * @extends Factory<User>
 */
final class UserFactory extends Factory
{
    /**
     * @return array<model-property<User>, mixed>  // <------- CHANGED -------
     */
    #[Override]
    public function definition(): array
    {
        return [
            'name' => fake()->name(),
            'email' => fake()->unique()->safeEmail(),
            'password' => Str::random(32),
            'foo' => 'bar',  // <------------------------------ ADDED --------
        ];
    }
}
```

Now I get the following error message:

```
Method Database\Factories\UserFactory::definition() should return array<model property of App\Models\User, mixed> but returns array<string, mixed>.  
💡 The given string should be a property of App\Models\User, foo given.                                                                                                   
✏️  database/factories/UserFactory.php 
```

This happens because the `User` model does not have a `foo` property. With the new larastan version `2.9.8` we can check this now. Earlier larastan would have said that the code is correct. If we had then executed `php artisan migrate:fresh`, an error similar to this would have occurred:

```
SQLSTATE[42703]: Undefined column: 7 ERROR:  column "foo" of relation "users" does not exist
```

Now we can avoid this error by using the new larastan version `2.9.8`.

calebdw:
@markuskooche, you should not be adding the php docs to the `definition` method as they already exist on the parent---you're creating more work for yourself

Also, there's no need for the `#[Override]` as the `definition` method is an abstract method that **must** be implemented

markuskooche:
> @markuskooche, you should not be adding the php docs to the `definition` method as they already exist on the parent---you're creating more work for yourself
> 
> 
> 
> Also, there's no need for the `#[Override]` as the `definition` method is an abstract method that **must** be implemented

1. But that's exactly the advantage of PR. 😉 It allows Larastan to recognize more mistakes. Therefore, you MUST add the PHP documents with `model-property<User>`.

2. Thanks, but I didn't add this. This is a [rector](https://github.com/rectorphp/rector) rule.

calebdw:
> But that is precisely the advantage of the PR. 😉 It allows Larastan to recognize more mistakes. Therefore, you MUST add the PHP documents with ˋmodel-propertyˋ.

No, you do not need to add `model-property`, the stub already has the model-property type applied with User since you specify `@extends Factory<User>`. Methods automatically inherit the PHPDocs so there is no need to override

KennedyTedesco:
One issue I'm having (and I believe is related to this change) is when I do some joins, for example:

```php
MyTable::query()->join('other_table', ...)->where('my_table.status', 1)
```
It doesn't recognize `my_table.status` as an attribute of `MyTable` because of the prefix `my_table` in the `where()` clause.

So, I'm getting:

```
Parameter #1 $column of method Illuminate\Database\Eloquent\Builder<App\Models\MyTable>::where() expects array<int|model property of App\Models\MyTable,
         mixed>|(Closure(Illuminate\Database\Eloquent\Builder<App\Models\MyTable>):
         Illuminate\Database\Eloquent\Builder<App\Models\MyTable>)|(Closure(Illuminate\Database\Eloquent\Builder<App\Models\MyTable>): void)|Illuminate\Contracts\Database\Query\Expression|model property of App\Models\MyTable, 'my_table.status' given.
```

johnbacon:
We are getting what we strongly believe to be false positives as a result of this pull request. Here's a minimal example:

**Promokit.php model:**
```php
<?php

namespace App\Models;

use App\Models\Bases\GigSaladModel; /* Extends base Laravel Model */

/**
 * @mixin IdeHelperPromokit
 */
class Promokit extends GigSaladModel
{
	protected $casts = [
		'gigsalad_active' => 'bool',
		'approved' => 'bool',
	];

	public function scopeWherePublic(Builder $query): Builder
	{
		return $query->where('gigsalad_active', 1)
			->where('approved', 1);
	}
}
```

**Abbreviated _ide_helper_models.php mixin:**
```php
<?php

namespace App\Models{
/**
 * 
 *
 * @property bool $gigsalad_active
 * @property bool $approved
 * @method static \Illuminate\Database\Eloquent\Builder|Promokit whereApproved($value)
 * @method static \Illuminate\Database\Eloquent\Builder|Promokit whereGigsaladActive($value)
 * @mixin \Eloquent
 */
	#[\AllowDynamicProperties]
	class IdeHelperPromokit {}
}
```

As you can see, `gigsalad_active` and `approved` are both defined as properties of the model. However, we get errors suggesting otherwise:

```
Parameter #1 $column of method Illuminate\Database\Eloquent\Builder<Illuminate\Database\Eloquent\Model>::where() expects array<int|model property of Illuminate\Database\Eloquent\Model, mixed>|(Closure(Illuminate\Database\Eloquent\Builder): Illuminate\Database\Eloquent\Builder)|(Closure(Illuminate\Database\Eloquent\Builder): void)|Illuminate\Contracts\Database\Query\Expression|model property of Illuminate\Database\Eloquent\Model, 'gigsalad_active' given.
💡 The given string should be a property of Illuminate\Database\Eloquent\Model, gigsalad_active given.
```

```
Parameter #1 $column of method Illuminate\Database\Eloquent\Builder<Illuminate\Database\Eloquent\Model>::where() expects array<int|model property of Illuminate\Database\Eloquent\Model, mixed>|(Closure(Illuminate\Database\Eloquent\Builder): Illuminate\Database\Eloquent\Builder)|(Closure(Illuminate\Database\Eloquent\Builder): void)|Illuminate\Contracts\Database\Query\Expression|model property of Illuminate\Database\Eloquent\Model, 'approved' given.
💡 The given string should be a property of Illuminate\Database\Eloquent\Model, approved given.
```

This is a result of these two lines: https://github.com/larastan/larastan/pull/1959/files#diff-7e443364f92b68ce916afad09165775d6dc1f77fbac96b6767e91ecb618520aeR89-R91

This _only_ occurs within scopes, and is happening 20 times throughout our codebase. We've resorted to ignoring this error now, and `checkModelProperties: true` works otherwise as it did previously.

cc: @jeramyhing

calebdw:
@johnbacon, 


You have two issues:
1. Larastan and IDE-Helper don't mix. You can certainly use the IDE-Helper to *help your IDE*, but Larastan **should not** be analyzing its stubs (your `@mixin` might be conflicting with this, you might need to remove it). Trying to use IDE-Helper with Larastan is not supported
3. Your scope needs the proper generics

```php
class Promokit extends GigSaladModel
{
    protected $casts = [
        'gigsalad_active' => 'bool',
        'approved' => 'bool',
    ];

    /**
     * @param Builder<static> $query
     * @return Builder<static>
     */
    public function scopeWherePublic(Builder $query): Builder
    {
        return $query->where('gigsalad_active', 1)
            ->where('approved', 1);
    }
}
```

johnbacon:
Hi @calebdw, thank you for the quick response!

Since then we've eliminated IDE Helper as the culprit for _this_ particular issue, and we haven't had any issues using it with Larastan in the past. Actually, it's saved us a lot of time!

Your second point does indeed cause the errors to go away. However, we don't feel there's any benefit to this. If the point is to identify properties that do not exist, we believe this is still a false positive, causes more work through enforced maintenance of unnecessary docblocks, and ultimately regresses the value and accuracy of Larastan.

calebdw:
@johnbacon, 

You can use ide-helper with Larastan if you want, but it's been the cause of many weird bugs and it's not supported. Larastan can understand Laravel just fine without it.

The problem at hand is that Larastan does not know what the generic model for the `$query` parameter is without the PHPDocs. Given that this is static analysis, you're going to need to add the types in order for Larastan to work properly.

Your options are:
- ignore the errors---this is not good and can lead to actual bugs
- add the docs to the scopes
- use custom Builder classes (my preference), this reduces the magic behind Laravel scopes and is better suited to static analysis

Builder class example (Laravel 11 has the HasBuilder trait):

```php

/** @extends \Illuminate\Database\Eloquent\Builder<Promokit> */
class PromokitBuilder extends Builder
{
    /**  @return $this */
    public function wherePublic(): static
    {
        return $this->where('gigsalad_active', 1)
            ->where('approved', 1);
    }
}

class Promokit extends GigSaladModel
{
    /** @use HasBuilder<PromokitBuilder> */
    use HasBuilder;

    protected static string $builder = PromokitBuilder::class;
}
```

PR #1969, opened 2024-06-09
Add chunk* and each* methods to Eloquent Builder stub

leo108:
- [x] Added or updated tests

### DIFF
--- a/stubs/common/EloquentBuilder.stub
+++ b/stubs/common/EloquentBuilder.stub
@@ -544,6 +544,47 @@ class Builder
      * @return bool
      */
     public function chunk($count, $callback);
+
+    /**
+     * Run a map over each item while chunking.
+     *
+     * @template TReturn of mixed
+     * @param  callable(TModelClass): TReturn $callback
+     * @param  int  $count
+     * @return \Illuminate\Support\Collection<int, TReturn>
+     */
+    public function chunkMap(callable $callback, $count = 1000);
+
+    /**
+     * Chunk the results of a query by comparing IDs.
+     *
+     * @param  int  $count
+     * @param  callable(\Illuminate\Database\Eloquent\Collection<int, TModelClass>, int): mixed $callback
+     * @param  string|null  $column
+     * @param  string|null  $alias
+     * @return bool
+     */
+    public function chunkById($count, callable $callback, $column = null, $alias = null);
+
+    /**
+     * Execute a callback over each item while chunking.
+     *
+     * @param  callable(TModelClass, int): mixed $callback
+     * @param  int  $count
+     * @return bool
+     */
+    public function each(callable $callback, $count = 1000);
+
+    /**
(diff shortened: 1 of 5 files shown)


leo108:
@canvural any chance to check this PR? let me know if you have any idea, thanks!

stubs/common/EloquentBuilder.stub
```diff
@@ -544,6 +544,47 @@ class Builder
...
+     * @param  string|null  $column
+     * @param  string|null  $alias
+     * @return bool
+     */
+    public function chunkById($count, callable $callback, $column = null, $alias = null);
+
+    /**
+     * Execute a callback over each item while chunking.
+     *
+     * @param  callable(TModelClass): mixed $callback
```
calebdw: The callable takes two args (the other method and files need to fixed as well).

```suggestion
     * @param  callable(TModelClass, int): mixed $callback
```

calebdw (commented):
Missing some args

leo108:
@calebdw Thanks for the review, the missing arg has been added, I also add the missing arg to eachById method, please check again.

leo108:
@canvural Can you check this PR when you have chance?

tests/Type/GeneralTypeTest.php
```diff
@@ -74,6 +74,10 @@ public static function dataFileAsserts(): iterable
             yield from self::gatherAssertTypes(__DIR__ . '/data/query-builder-l10-24.php');
         }
 
+        if (version_compare(LARAVEL_VERSION, '10.27.0', '>=')) {
+            yield from self::gatherAssertTypes(__DIR__ . '/data/eloquent-builder-l10-27.php');
+        }
```
canvural, line 79: We are adding the stubs to `10.0.0` anyway. So it'll be available for 10.x. No need to check for 10.27 specifically.

If there is a big change before and after 10.27, then you need to create a folder `10.27.0` in `stubs` directory and put the stubs there.

tests/Type/data/eloquent-builder-l10-27.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace ModelRelations;
```
canvural, line 3:
```suggestion
namespace EloquentBuilderL1027;
```

or something like that

stubs/common/EloquentBuilder.stub
```diff
@@ -544,6 +544,47 @@ class Builder
      * @return bool
      */
     public function chunk($count, $callback);
+
+    /**
+     * Run a map over each item while chunking.
+     *
+     * @template TReturn of mixed
```
canvural, line 551:
```suggestion
     * @template TReturn
```

you don't need that last part.

canvural (changes requested):

rudiedirkx:
Seems like all of this is in `2.x` already?

PR #1980, opened 2024-07-01
Prevent error Argument #2 ($array) must be of type array

sangnguyenplus:
This PR will fix the following error when checking model macros:

`Uncaught TypeError: array_key_exists(): Argument #2 ($array) must be of type array, null given in ~/vendor/larastan/larastan/src/Methods/MacroMethodsClassReflectionExtension.php:135
`

### DIFF
--- a/src/Methods/MacroMethodsClassReflectionExtension.php
+++ b/src/Methods/MacroMethodsClassReflectionExtension.php
@@ -128,13 +128,15 @@ class MacroMethodsClassReflectionExtension implements MethodsClassReflectionExte
                 $refProperty = $macroClassReflection->getNativeReflection()->getProperty($macroTraitProperty);
                 $refProperty->setAccessible(true);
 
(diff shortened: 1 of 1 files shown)


szepeviktor:
Hello Sang!
Why don't you simply pass an array in your code?

sangnguyenplus:
> Hello Sang! Why don't you simply pass an array in your code?

I don't know why, but in my application `$refProperty->getValue()` can be null, not an array.

By this change, it works for my app.

Maybe it is an issue from my app, but I think it's fine to check it in Larastan package too.

szepeviktor:
Thank you.

Could you share your macro code?

sangnguyenplus:
> Thank you.
> 
> Could you share your macro code?

I'm using this package https://github.com/javoscript/laravel-macroable-models for Laravel model macroable.

image

It was working fine before, but I want to override magic method __get(), then this issue happens.

canvural:
Hi,

Would be good to have a test for this.

PR #1983, opened 2024-07-04
fix: Allow custom Laravel bootstrap file path

NickSdot:
### Changes

There where previous attempts (https://github.com/larastan/larastan/pull/851, https://github.com/larastan/larastan/pull/852, https://github.com/larastan/larastan/pull/857) to make the detection of the Laravel bootstrap file better. However, after reviewing the code and discussions it appears to me that the follow up PR never landed. 

That said, there is no ultimate way to guess where people have their bootstrap file located. For that very reason this PR introduces a similar approach as Laravel uses [here ](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Foundation/Application.php#L595) for the storage path.

**Added value:**

This PR allows to either define the path statically in your environment

```bash
# ~/.bashrc or ~/.zshrc
export LARAVEL_BOOTSTRAP_FILE_PATH=Framework/bootstrap/app.php
```
or to prepend it to the command.

```console
export LARAVEL_BOOTSTRAP_FILE_PATH=Framework/bootstrap/app.php; ./vendor/bin/phpstan analyse
```

### Breaking changes

I chosed a currently unused var name. No B/C implications, imho.

### Notes
- Not sure where to document this. Keen to add a paragraph where ever you'd like to have it.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -12,7 +12,7 @@ if (! defined('LARAVEL_START')) {
     define('LARAVEL_START', microtime(true));
 }
 
-if (false !== $applicationPath = $_ENV['LARAVEL_BOOTSTRAP_FILE_PATH'] ?? $_SERVER['LARAVEL_BOOTSTRAP_FILE_PATH'] ?? false) { // Custom path in environment
+if (false !== $applicationPath = getenv('LARAVEL_BOOTSTRAP_FILE_PATH')) { // Custom path in environment
     if (file_exists($applicationPath)) {
         $app = require $applicationPath;
     }
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -12,7 +12,11 @@ if (! defined('LARAVEL_START')) {
     define('LARAVEL_START', microtime(true));
 }
 
-if (file_exists($applicationPath = getcwd().'/bootstrap/app.php')) { // Applications and Local Dev
+if (false !== $applicationPath = $_ENV['LARAVEL_BOOTSTRAP_FILE_PATH'] ?? $_SERVER['LARAVEL_BOOTSTRAP_FILE_PATH'] ?? false) { // Custom path in environment
+    if (file_exists($applicationPath)) {
+        $app = require $applicationPath;
+    }
+} elseif (file_exists($applicationPath = getcwd().'/bootstrap/app.php')) { // Applications and Local Dev
     $app = require $applicationPath;
 } elseif (file_exists($applicationPath = dirname(__DIR__, 3).'/bootstrap/app.php')) { // Relative path from default vendor dir
     $app = require $applicationPath;


szepeviktor:
Thank you!

That command line is really
```shell
LARAVEL_BOOTSTRAP_FILE_PATH="Framework/bootstrap/app.php" ./vendor/bin/phpstan analyse
```

bootstrap.php
```diff
@@ -12,7 +12,11 @@
     define('LARAVEL_START', microtime(true));
 }
 
-if (file_exists($applicationPath = getcwd().'/bootstrap/app.php')) { // Applications and Local Dev
+if (false !== $applicationPath = $_ENV['LARAVEL_BOOTSTRAP_FILE_PATH'] ?? $_SERVER['LARAVEL_BOOTSTRAP_FILE_PATH'] ?? false) { // Custom path in environment
```
szepeviktor:
```suggestion
if (false !== $applicationPath = getenv('LARAVEL_BOOTSTRAP_FILE_PATH')) { // Custom path in environment
```

szepeviktor:
How about matching the variable name in the framework??
https://github.com/laravel/framework/blob/1a1a61068bc3c5594376a559e424ae09ec3fe64a/src/Illuminate/Foundation/Application.php#L253-L259
🤔

NickSdot:
> How about matching the variable name in the framework?? https://github.com/laravel/framework/blob/1a1a61068bc3c5594376a559e424ae09ec3fe64a/src/Illuminate/Foundation/Application.php#L253-L259 🤔

Pre-bootstrapping Laravel uses the `LARAVEL_` prefix. I followed that convention.
https://github.com/laravel/framework/blob/11.x/src/Illuminate/Foundation/Application.php#L598

I believe this makes sense, because at this point the app does not yet exist – it's about to be bootstrapped.

Edit: I hope I got you right that this is about `LARAVEL_` vs. `APP_`?

bootstrap.php
NickSdot: We don't want to check `$_SERVER`? 

Eg: Some people might prefer a decorator in their `public/index.php`. By checking for `$_SERVER` everyone could do how they prefer.

NickSdot: After a second thought, you are absolutely right – we don't need that. 🤝

szepeviktor (approved):

canvural:
You can already add your bootstrap file to the PHPStan's [config](https://phpstan.org/config-reference#bootstrap). Does that not work?

szepeviktor:
> You can already add your bootstrap file

There is a misunderstanding.
This is not PHPStan's "bootstrap" but a file for the Laravel framework.
https://github.com/laravel/laravel/blob/11.x/bootstrap/app.php

canvural:
Yeah, that's what I meant. If Larastan can't bootstrap your app for some reason, you can write your own bootstrap file that boots up your app. And that file to the config.

szepeviktor:
If you include a custom `bootstrap/app.php` through PHPStan's "bootstrap" our bootstrap.php (modified in this PR) will throw an exception.

NickSdot:
> If you include a custom `bootstrap/app.php` through PHPStan's "bootstrap" our bootstrap.php (modified in this PR) will throw an exception.

This is correct. I didn't find a way around, even it was mentioned as an option in one of the discussions linked above.

NickSdot:
> Yeah, that's what I meant. If Larastan can't bootstrap your app for some reason, you can write your own bootstrap file that boots up your app. And that file to the config.

Following up here just in case it's not really clear what the problem is. Larastan – via it's `vendor/larastan/larastan/extension.neon` loads `vendor/larastan/larastan/bootstrap.php`. Even if I add

```NEON
parameters:
	bootstrapFiles:
		- Framework/bootstrap/app.php
```

to my main `phpstan.neon`, Larastans `extension.neon` loads the `vendor/larastan/larastan/bootstrap.php` which will `throw new Exception('Could not find Laravel bootstrap file nor Testbench is installed. Install orchestra/testbench if analyzing a package.');` because non of the defined bootstrap paths are valid in our application.

There is two solutions I can think of:

1) The one I proposed with this PR.
2) Modify `vendor/larastan/larastan/bootstrap.php` to something like the following, to no longer throw an exception if the bootstrap file wasn't found and allow the applications `phpstan.neon` to do it's own bootstrapping after. 

```php
if (file_exists($applicationPath = getcwd().'/bootstrap/app.php')) { // Applications and Local Dev
    $app = require $applicationPath;
} elseif (file_exists($applicationPath = dirname(__DIR__, 3).'/bootstrap/app.php')) { // Relative path from default vendor dir
    $app = require $applicationPath;
} elseif (trait_exists(CreatesApplication::class)) { // Packages
    $app = ApplicationResolver::resolve();
}

if(isset($app)) {
    if ($app instanceof Application) {
        $app->make(Kernel::class)->bootstrap();
    } elseif ($app instanceof LumenApplication) {
        $app->boot();
    }

    if (! defined('LARAVEL_VERSION')) {
        define('LARAVEL_VERSION', $app->version());
    }
}
```

If there is anything I can do to bring this forward, please let me know. 🙏

StrykeSlammerII:
> There is two solutions I can think of:
> 
>     1. The one I proposed with this PR.
> 
>     2. Modify `vendor/larastan/larastan/bootstrap.php` to something like the following, to no longer throw an exception if the bootstrap file wasn't found and allow the applications `phpstan.neon` to do it's own bootstrapping after.

Something to consider from a user's perspective:
Netbeans expects all options to be in the config file.
I don't know whether other IDEs will allow for prepending arguments to the CLI or not.

This PR allows defining the path in the environment, which is a step forward, but following PHPStan's options IMO is prefereable.

(In the meantime I'm investigating whether I need to install `orchestral/testbench` or whether manually patching the second option above will work for me.)

NickSdot:
> Something to consider from a user's perspective:
> 
> Netbeans expects all options to be in the config file.
> 
> I don't know whether other IDEs will allow for prepending arguments to the CLI or not.
> 
> This PR allows defining the path in the environment, which is a step forward, but following PHPStan's options IMO is prefereable.

Unfortunately, I don't know how Netbeans handles things and I cannot add any valuable answer here.

But I am happy to modify my PR to the second approach I proposed, if the maintainers would like me to.

> (In the meantime I'm investigating whether I need to install `orchestral/testbench` or whether manually patching the second option above will work for me.)

My temporary solution looks something like the below. On each auto-load dump delete the bootstrap file in the vendor folder; then symlink the actual bootstrap file.

Not beautiful, and hopefully really only a temporary solution, but it works. 

**composer.json**

```json
    "scripts": {
        "pre-autoload-dump": [
            "[ $COMPOSER_DEV_MODE -eq 0 ] || rm -f vendor/larastan/larastan/bootstrap.php && ln -s ./path/to/actual/bootstrap.php vendor/larastan/larastan/bootstrap.php"
        ],
```

Maybe this helps.

--
Edit: added a part I missed to copy before

NickSdot:
Fixed in https://github.com/larastan/larastan/pull/2077

PR #1988, opened 2024-07-05
Add types for first methods in HasManyThrough

korridor:
- [X] Added or updated tests

I did not yet create an issue for this PR. If this is required, please notify me.

**Changes**

Added types for the `first`, `firstOrFail` and `firstOr` methods in the `HasManyThrough` relation. I added tests to show what my changes are fixing. The tests failed before I added the method stubs.

**Breaking changes**

No breaking changes

### DIFF
--- a/stubs/10.0.0/HasManyThrough.stub
+++ b/stubs/10.0.0/HasManyThrough.stub
@@ -49,6 +49,34 @@ class HasManyThrough extends Relation
     public function cursorPaginate($perPage = null, $columns = ['*'], $cursorName = 'cursor', $cursor = null);
 
     /**
+     * Execute the query and get the first related model.
+     *
+     * @param  array<int, string>  $columns
+     * @return TRelatedModel|null
+     */
+    public function first($columns = ['*']);
+
+    /**
+     * Execute the query and get the first result or throw an exception.
+     *
+     * @param  array<int, string>  $columns
+     * @return TRelatedModel
+     *
+     * @throws \Illuminate\Database\Eloquent\ModelNotFoundException<TRelatedModel>
+     */
+    public function firstOrFail($columns = ['*']);
+
+    /**
+     * Execute the query and get the first result or call a callback.
+     *
+     * @template TReturnClosure
+     * @param  (\Closure(): TReturnClosure)|array<int, string>  $columns
+     * @param  (\Closure(): TReturnClosure)|null  $callback
+     * @return TRelatedModel|TReturnClosure
+     */
+    public function firstOr($columns = ['*'], ?\Closure $callback = null);
+
+    /**
      * Find a related model by its primary key.
      *
      * @param  mixed  $id
@@ -72,4 +100,5 @@ class HasManyThrough extends Relation
      * @return HasOneThrough<TRelatedModel>
      */
     public function one();
+
 }
--- a/tests/Type/data/custom-eloquent-collection.php
+++ b/tests/Type/data/custom-eloquent-collection.php
@@ -116,6 +116,9 @@ function test(): void
     assertType('App\TransactionCollection<int, App\Transaction>', (new User)->transactions()->find([1, 2]));
     assertType('Illuminate\Database\Eloquent\Collection<int, App\User>', (new Role)->users()->find([1, 2]));
     assertType('App\Transaction|null', (new User)->transactions()->find(1));
+    assertType('App\Transaction|null', (new User)->transactions()->first());
+    assertType('App\Transaction', (new User)->transactions()->firstOrFail());
+    assertType('App\Transaction|string', (new User)->transactions()->firstOr(callback: fn () => ''));
     assertType('App\TransactionCollection<int, App\Transaction>', (new User)->transactions()->findOrFail([1, 2]));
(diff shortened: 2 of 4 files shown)


calebdw:
This has already been fixed in the [framework](https://github.com/laravel/framework/pull/51851) (for 11.x), and I'm working on a PR to update Larastan

szepeviktor:
Thank you.

szepeviktor:
> This has already been fixed

@calebdw _This will be_

szepeviktor:
@korridor We still support L9.

stubs/10.0.0/HasManyThrough.stub
```diff
@@ -48,6 +48,34 @@ class HasManyThrough extends Relation
...
+    public function firstOrFail($columns = ['*']);
+
+    /**
+     * Execute the query and get the first result or call a callback.
+     *
+     * @template TReturnClosure
+     * @param  \Closure(): TReturnClosure|array<int, string>  $columns
+     * @param  \Closure(): TReturnClosure|null  $callback
+     * @return TRelatedModel|TReturnClosure
+     */
```
calebdw: This need to be:

```suggestion
    /**
     * Execute the query and get the first result or call a callback.
     *
     * @template TValue
     *
     * @param  (\Closure(): TValue)|array  $columns
     * @param  (\Closure(): TValue)|null  $callback
     * @return ($callback is null ? TRelatedModel|null : TRelatedModel|TValue)
     */
```

korridor: Thanks for the suggestion, I originally had exactly that, but I think this is incorrect.
If you look at the implementation, it is implicitly required to provide a callback. The only reason the second argument is nullable is that it is possible to provide a callback in the first argument.

If you would run `->firstOr()` and the query returns no result, you would get an exception/warning, not `null`, since `$callback` is `null` and `null` is not callable.

You can test that by running `$callback = null; $callback();` in a PHP console, f.e. `php -a`.

```php
public function firstOr($columns = ['*'], ?Closure $callback = null)
{
    if ($columns instanceof Closure) {
        $callback = $columns;

        $columns = ['*'];
    }

    if (! is_null($model = $this->first($columns))) {
        return $model;
    }

    return $callback();
}
```

korridor:
@szepeviktor This is my first contribution, so I'm not quite sure what to do with the fact that you support L9. You are referring to the failing GitHub actions right?

szepeviktor:
> This is my first contribution, so I'm not quite sure what to do with the fact that you support L9. You are referring to the failing GitHub actions right?

Welcome to Larastan 👋🏻 

Yes, assertion should not run on L9.

tests/Type/data/custom-eloquent-collection.php
```diff
@@ -116,6 +116,9 @@ function test(): void
     assertType('App\TransactionCollection<int, App\Transaction>', (new User)->transactions()->find([1, 2]));
     assertType('Illuminate\Database\Eloquent\Collection<int, App\User>', (new Role)->users()->find([1, 2]));
     assertType('App\Transaction|null', (new User)->transactions()->find(1));
+    assertType('App\Transaction|null', (new User)->transactions()->first());
+    assertType('App\Transaction', (new User)->transactions()->firstOrFail());
+    assertType('App\Transaction|string', (new User)->transactions()->firstOr(callback: fn () => ''));
```
calebdw, line 121: You should also have similar tests in one of the `model-relations` files

korridor: Thanks, I added something similar to the `model-relations` file. Please check if this is what you had in mind.

korridor:
@szepeviktor I just pushed changes that should make my changes compatible to L9. Please check if this is what you had in mind.

szepeviktor (approved):

tests/Type/data/custom-eloquent-collection.php
calebdw: Thanks, now that I think about it these tests can be removed now that you added similar ones to model-relations since they really have nothing to do with *custom collections*

stubs/10.0.0/HasManyThrough.stub
```diff
@@ -48,6 +48,34 @@ class HasManyThrough extends Relation
...
+     *
+     * @throws \Illuminate\Database\Eloquent\ModelNotFoundException<TRelatedModel>
+     */
+    public function firstOrFail($columns = ['*']);
+
+    /**
+     * Execute the query and get the first result or call a callback.
+     *
+     * @template TReturnClosure
+     * @param  \Closure(): TReturnClosure|array<int, string>  $columns
```
calebdw: All of these closure generics are incorrect (same for other stub), they should be

```php
     * @param  (\Closure(): TReturnClosure)|array<int, string>  $columns
     * @param  (\Closure(): TReturnClosure)|null  $callback
```

korridor: I think that at least PHPStan assumes the braces otherwise the tests would fail, but I agree that your way is more readable. I just pushed the changes.

canvural:

PR #1989, opened 2024-07-08
New rule: Ensure access to defined filesystems

aictur:
- [x] Added or updated tests

**Changes**

Added a rule that ensures access to a defined filesystem or 'disk' configured in `config/filesystems.php` when using Laravel Storage abstraction API.

**Breaking changes**

No breaking changes.

### DIFF
--- a/src/Rules/NoUndefinedFilesystemsRule.php
+++ b/src/Rules/NoUndefinedFilesystemsRule.php
@@ -1,0 +1,112 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Larastan\Larastan\Rules;
+
+use Illuminate\Support\Facades\Config;
+use Illuminate\Support\Facades\Storage;
+use PhpParser\Node;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Name\FullyQualified;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleError;
+use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\Type\ObjectType;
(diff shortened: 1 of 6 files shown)


szepeviktor:
This is very nice PR.

szepeviktor:
This is unhandled
https://github.com/filamentphp/filament/blob/7378aabb6aeacab86387662f3765f84aeb53ca0d/packages/actions/src/Exports/Models/Export.php#L92

tests/Rules/data/CorrectFilesystem.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Tests\Rules\Data;
+
+use Illuminate\Support\Facades\Storage;
+
+class CorrectFilesystem
+{
+    public function getStorage(): Storage
+    {
+        return Storage::disk('local');
```
calebdw, line 13: There's no need to have two separate test files, you can add both lines in this test (rename the file of course, and there's no need for there to be a class) and just assert that the appropriate line throws the error:

The test file can be as simple as:
```php
<?php

declare(strict_types=1);

use Illuminate\Support\Facades\Storage;

function test(): void
{
    Storage::disk('local');
    Storage::disk('this-is-not-defined');
}
```

szepeviktor: and maybe a variable: `Storage::disk($_GET['disk']);`

src/Rules/NoUndefinedFilesystemsRule.php
```diff
@@ -0,0 +1,112 @@
...
+ * For example:
+ * Storage::disk('this-is-not-defined')
+ *
+ * @implements Rule<StaticCall>
+ */
+class NoUndefinedFilesystemsRule implements Rule
+{
+    public function __construct(protected ReflectionProvider $reflectionProvider)
+    {
+    }
```
calebdw, line 33: This is not needed and can be removed

canvural:

PR #1990, opened 2024-07-10
feat: update relation generics (support Laravel >= 11.15)

calebdw:
- [x] Added or updated tests
- [x] Documented user facing changes

Closes #1284, closes #1550, closes #1676

**Changes**
Hello!

This PR updates the Relation generics to align with https://github.com/laravel/framework/pull/51851 which implemented @nagmat84's fixes and notation from his PR (#1285) which introduced two new template types, `TDeclaringModel` and `TResult` (as well as `TIntermediateModel` for the *Through relations).

Here's an example of how Relations should now look:

```php
/** @return HasOne<Address, $this> */
public function address(): HasOne
{
    return $this->hasOne(Address::class);
}

/** @return HasMany<Post, $this> */
public function posts(): HasMany
{
    return $this->hasMany(Post::class);
}

/** @return BelongsToMany<Role, $this> */
public function roles(): BelongsToMany
{
    return $this->belongsToMany(Role::class);
}

/** @return HasOne<Mechanic, $this> */
public function mechanic(): HasOne
{
    return $this->hasOne(Mechanic::class);
}

/** @return HasOneThrough<Car, Mechanic, $this> */
public function car(): HasOneThrough
{
    return $this->hasOneThrough(Car::class, Mechanic::class);
}

/** @return HasManyThrough<Part, Mechanic, $this> */
public function parts(): HasManyThrough
{
    return $this->hasManyThrough(Part::class, Mechanic::class);
}
```

Of course Larastan can still parse the file to determine the generics if the phpdocs aren't provided, but it's always good to provide the generics.

There was a good bit of code I was able to clean up and greatly simplify along the way, as well as various bugs I encountered that needed to be fixed.

**Breaking changes**
I've tried my best to keep the breaking changes on 11.x, the older version still use just the `TRelatedModel` template (with the exception of the BelongsTo relation), but some changes were necessary to keep the code from becoming too complicated:
- renamed `TModelClass` to `TModel`
- renamed `TChildModel` to `TDeclaringModel`

PR #1995, opened 2024-07-15
Add typehint to `env` when specifying default value

bram-pkg:
- [x] Added or updated tests

**Changes**

This PR adds a generic value to the `env()` helper when a second parameter is specified with a default value. When this value is non-null, it will be used as the `env()` return type.

**Breaking changes**

I am not sure this is a breaking change, please advise.

### DIFF
--- a/stubs/common/Helpers.stub
+++ b/stubs/common/Helpers.stub
@@ -197,6 +197,6 @@ function blank($value) {}
  *
  * @param  string  $key
  * @param  TDefault  $default
- * @return ($default is null ? mixed : TDefault)
+ * @return bool|string|TDefault
  */
 function env($key, $default = null) {}
--- a/stubs/common/Helpers.stub
+++ b/stubs/common/Helpers.stub
@@ -189,3 +189,14 @@ function filled($value) {}
  * @return bool
  */
 function blank($value) {}
+
+/**
+ * Gets the value of an environment variable.
+ *
+ * @template TDefault
+ *
+ * @param  string  $key
+ * @param  TDefault  $default
+ * @return ($default is null ? mixed : TDefault)
+ */
+function env($key, $default = null) {}
--- a/tests/Type/data/helpers.php
+++ b/tests/Type/data/helpers.php
@@ -168,7 +168,8 @@ function test(?int $value = 0): void
         assertType('int', $value);
     }
 
-    assertType('int', env('SESSION_LIFETIME', 120));
-    assertType('string', env('APP_NAME', 'Laravel'));
-    assertType('mixed', env('APP_NAME'));
+    assertType('bool|string|null', env('foo'));
+    assertType('bool|string|null', env('foo', null));
+    assertType('bool|int|string', env('foo', 120));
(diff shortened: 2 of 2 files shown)


bram-pkg:
@calebdw are the test failures related to my change? 🤔

calebdw:
No, the ci is failing on >= Laravel 11.15 until #1990 is finished up and merged.

However, can you please write some type tests for this (in `tests/Type/data/helpers.php`)?

bram-pkg:
Yeah, ofcourse! I will attempt to write some tests.

bram-pkg:
@calebdw added two types of test, one to check if the specific "cannot cast mixed to string" (or any `mixed to X` casting for that matter) does not reoccur, and one that validates output types.

tests/Integration/data/helpers.php
```diff
@@ -17,3 +17,8 @@ function retryAcceptsCallableAsSleepMilliseconds(): void
         return true;
     });
 }
+
+function envReturnsGenericValue(): void
+{
+    (string) env('APP_NAME', 'Laravel');
+}
```
calebdw: There's really no need for this test if the other types test pass because it would already be a string---no need to cast

bram-pkg: Ah alright, I can remove it then. I guess that integration file is reserved for more niche scenarios.

bram-pkg: Squashed, dropped that commit and force pushed.

calebdw:
Looks good to me, however, it seems to me like the root cause for this change is that the return type `mixed` is too broad and is throwing phpstan errors when trying to cast it. All env values are essentially strings---however, the env helper does cast boolean strings to `bool`, so perhaps a type of `bool|string|null` is a better type than `mixed`?

stubs/common/Helpers.stub
```diff
@@ -189,3 +189,14 @@ function filled($value) {}
...
 function blank($value) {}
+
+/**
+ * Gets the value of an environment variable.
+ *
+ * @template TDefault
+ *
+ * @param  string  $key
+ * @param  TDefault  $default
+ * @return ($default is null ? mixed : TDefault)
```
calebdw: What about the following?

```suggestion
 * @return bool|string|TDefault
```

This would yield the following types:

```php
assertType('bool|string|null', env('foo'));
assertType('bool|string|null', env('foo', null));
assertType('bool|string', env('foo', ''));
assertType('bool|int|string', env('foo', 0));
assertType('bool|string', env('foo', false));
```

bram-pkg: Looking at the internals of `env()`, you seem to be correct.

That's even cleaner indeed. I will apply your suggestion.
Sorry for the back and forth, I am not too familiar yet contributing to PHPStan and the like (static analysis and ast tools)

bram-pkg:
Suggestions have been applied.

calebdw (approved):

canvural:
Thank you!

bram-pkg:
Thank you for the rapid correspondence!

PR #1999, opened 2024-07-17
fix: handle model property aliases

### LINKED ISSUE #1998: false error when using alias for selecting Model properties
- Larastan Version: 2.9.8
- Laravel Version: 11.16.0

### Description

After updating my laravel to the latest version phpstan threw new errors at me I fixed lots of them by adding generic types to variables of the Builder class. 

But I couldn't find a solution for this error. 

Here is the code that I use.

```php
Example::find($id,['id as page_unique', 'title', 'description']);
```
And here is the error:

```

Parameter #2 $columns of method Illuminate\Database\Eloquent\Builder<App\Models\Example>::find() expects array<int, model property of App\Models\Example>|model     
         property of App\Models\Example, array<int, string> given.
         💡 Type #1 from the union: The given string should be a property of App\Models\Example, id as page_unique given.

```

calebdw:
When a model property is aliased, the alias should be removed before

- [x] Added or updated tests

**Changes**
Hello!

This closes #1998 

Thanks!

### DIFF
--- a/src/Types/ModelProperty/GenericModelPropertyType.php
+++ b/src/Types/ModelProperty/GenericModelPropertyType.php
@@ -68,6 +68,10 @@ class GenericModelPropertyType extends StringType
                 return AcceptsResult::createNo();
             }
 
+            if (str_contains($givenString, ' as ')) {
+                $givenString = explode(' as ', $givenString)[0];
+            }
+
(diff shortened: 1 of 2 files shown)


canvural:
Makes sense I guess. Thank you.

calebdw:
Thank you!

PR #2004, opened 2024-07-22
[3.x] feat: add support for config array shapes

calebdw:
- [x] Added or updated tests

**Changes**

Hello!

I got tired of having to do this number when accessing config values:

```php
            /** @var array{datacenter_id: int, worker_id: int, epoch: string} $config */
            $config = config('laraflake');
```

So I decided to add support for config array shapes (if they exist):

```php
    assertType('array{guard: string, passwords: string}|null', config('auth.defaults'));
```

Thanks! 

**Breaking changes**
None

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -175,6 +175,11 @@ services:
             - phpstan.broker.dynamicStaticMethodReturnTypeExtension
 
     -
+        class: Larastan\Larastan\ReturnTypes\ConfigDynamicStaticMethodReturnTypeExtension
+        tags:
+            - phpstan.broker.dynamicStaticMethodReturnTypeExtension
+
+    -
         class: Larastan\Larastan\ReturnTypes\AuthExtension
         tags:
             - phpstan.broker.dynamicStaticMethodReturnTypeExtension
@@ -254,6 +259,11 @@ services:
 
     -
         class: Larastan\Larastan\ReturnTypes\Helpers\CollectExtension
+        tags:
+            - phpstan.broker.dynamicFunctionReturnTypeExtension
+
+    -
+        class: Larastan\Larastan\ReturnTypes\Helpers\ConfigDynamicFunctionReturnTypeExtension
         tags:
             - phpstan.broker.dynamicFunctionReturnTypeExtension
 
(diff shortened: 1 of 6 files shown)


Roardom:
I'm using your fork with this patch included and have noticed it generates a bunch of

```
         Negated boolean expression is always true.                                                                                                  
         🪪  booleanNot.alwaysTrue                                                                                                                   
         💡 Because the type is coming from a PHPDoc, you can turn off this check by setting treatPhpDocTypesAsCertain: false in your phpstan.neon. 
```

However, my boolean config values are intended to be editable and switched between true/false for various occasions, such as enabling/disabling various systems or for global site state that would be dumb to put into a one-row 500-column database site config table. Adding `/** @var array{'configkeyname': bool} */` above the returned array in the config doesn't seem to clear the error, but that would be tedious and error prone to duplicate every key anyways. Perhaps there's a better way?

calebdw:
Ah, I see what the problem is---thanks for reporting!

This is an overlooked edge-case, but I'll get this fixed soon!

calebdw:
@Roardom, this should be fixed!

Roardom:
Thank you! It's recognized as boolean now.

For what's it's worth, we have a select few other config options that are interpreted differently depending on if they're a string or an int (e.g. To specify the default chatroom, if an int is provided in the config, the chatroom is selected by id, otherwise if a string is provided in the config, it's selected by its unique name), but I'd call that bad practice on our side. Maybe that edge case isn't common enough to worry about.

calebdw:
> For what's it's worth, we have a select few other config options that are interpreted differently depending on if they're a string or an int

Yeah that's rather tricky---I set the type based on the value to clean up always having to narrow the `mixed` type when most config options will likely only have a single type.

I suppose I could just set the type across the board to `bool|float|int|string|null` if it's not an array---but this would still require you to narrow the types at higher phpstan levels

canvural:
This is implemented under `checkConfigTypes` config parameter.

calebdw:
Do note that I opened this PR well over a year ago which was long before the `checkConfigTypes` existed

PR #2008, opened 2024-07-27
fix eloquent generics documentation

arxeiss:
- [ ] Added or updated tests

**Changes**

The documentation mention, we should be using `BelongsTo<Model, $this>` but when I use it, I got this error (stripped own FQDN)

```
Method Absence::user() should return
Illuminate\Database\Eloquent\Relations\BelongsTo<User, $this(Absence)> but returns  
Illuminate\Database\Eloquent\Relations\BelongsTo<User, Absence>
```

After changing `$this` to `self` it works well.

**Breaking changes**

None

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -45,7 +45,7 @@ In order for Larastan to recognize Model relationships:
 If the above conditions are not met, then adding the `@return` docblock can help
 
 ```php
-/** @return BelongsTo<User, $this> */
+/** @return BelongsTo<User, self> */
 public function user(): BelongsTo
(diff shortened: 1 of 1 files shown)


Afrowson:
For me this works:

   ```php
    /**
     * @return HasMany<BookingDetail, $this>
     */
    public function bookingDetails(): HasMany
    {
        return $this->hasMany(BookingDetail::class, 'bookings_id');
    }
```

calebdw:
Yeah, this PR can be closed now

PR #2013, opened 2024-08-02
Add arrow function and first class callable support for value helper

netpok:
- [x] Added or updated tests

**Changes**
This PR allows return type resolution for arrow functions and first class callables.

```php
// This PR adds support for:
value(fn () => 5);
value(intval(...), 5);
value(\Closure::fromCallable('intval'), 5);
value(SomeRandomClass::doMagic(...));

// to the already handled formats:
value(5);
value(function () {
    return 5;
});
```

Test are also added for the new formats as well as the missing closure format.

**Breaking changes**
None

### DIFF
--- a/src/ReturnTypes/Helpers/ValueExtension.php
+++ b/src/ReturnTypes/Helpers/ValueExtension.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Larastan\Larastan\ReturnTypes\Helpers;
 
 use PhpParser\Node\Expr\ArrowFunction;
+use PhpParser\Node\Expr\CallLike;
 use PhpParser\Node\Expr\Closure;
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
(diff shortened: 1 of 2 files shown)


netpok:
The broken tests are unrelated to this PR

canvural:
Looks good I think. It can be merged when it's out of draft.

netpok:
I put this MR in draft because it failed on our project when I tested with an internal error, the problem is that I forget why.

Currently I cannot reproduce it and it is possible that it was fixed in our project or in phpstan itself.

Edit: maybe it was in a later revision where I tried to correctly resolve compound types like `(Closure(): string)|string` which should resolve to `string`

---

On the other hand I'm not even sure that this extension is necessary because the new phpdoc on the value helper produces almost the same result, the only difference is that the literal values are resolved to their types:
- with the extension `value(7)` has the type `'7'`
- without the extension `value(7)` has the type `int`

In my opinion it would be better to just simply remove this extension or put a conditional like `laravel version <= 10`  unless I or someone else finds some time to handle compond types

canvural:

PR #2014, opened 2024-08-02
fix: make TGet covariant on Attribute stub

calebdw:
- [x] Added or updated tests

**Changes**
Hello!

I'm using the PHPStan [bleeding edge](https://phpstan.org/config-reference#bleeding-edge) config, and I started getting the following errors with the Model Attributes:

```php
    /** @return Attribute<string|null, never> */
    protected function surveyedName(): Attribute
    {
        return Attribute::get(function (): ?string {
            if (! $this->surveyable_type) {
                return null;
            }

            return Str::of($this->surveyable_type)
                ->replace('.', ' ')
                ->title()
                ->toString();
        });
    }
```

```
  148    Method <class>::surveyedName() should return Illuminate\Database\Eloquent\Casts\Attribute<string|null, never> but returns                        
         Illuminate\Database\Eloquent\Casts\Attribute<string|null, never>.                                                                                                    
         🪪  return.type                                                                                                                                                      
         💡 Template type TGet on class Illuminate\Database\Eloquent\Casts\Attribute is not covariant. Learn more: https://phpstan.org/blog/whats-up-with-template-covariant  
```

This PR updates the `TGet` template to be covariant and fixes some other issues with the stub.

Thanks!

**Breaking changes**
None

### DIFF
--- a/stubs/common/Attribute.stub
+++ b/stubs/common/Attribute.stub
@@ -3,7 +3,7 @@
 namespace Illuminate\Database\Eloquent\Casts;
 
 /**
- * @template TGet
+ * @template-covariant TGet
  * @template TSet
  */
 class Attribute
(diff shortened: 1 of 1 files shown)


szepeviktor:
Is this version high enough?
https://github.com/larastan/larastan/blob/c8ec6f04f800d9425be743998574f1034e2ffb3f/composer.json#L28

calebdw:
@szepeviktor, yes I believe that should be fine, the bleeding edge config is a sneak peak at the next version of phpstan

canvural:
AFAIR this was a bug in PHPStan. Making the type covariant is not the correct solution. Closing it for now.

Issue #676, opened 2020-10-07
[Hacktoberfest 2020] Update stubs for new `model-property` type


canvural:
## The Problem
With the new `model-property` type and `ModelPropertyRule` Larastan now has the ability to check the existence of model columns that are given to a method.

Adding the `model-property` type to the Laravel's core methods are done with the help of [PHPstan stubs](https://phpstan.org/user-guide/stub-files). Some [examples](https://github.com/nunomaduro/larastan/blob/master/stubs/EloquentBuilder.stub#L14) can be seen [here](https://github.com/nunomaduro/larastan/blob/master/stubs/Model.stub#L13). Currently, not all the Laravel methods that are doing something with the model properties are updated in stubs. And we need help with updating this.

## Contributing

You can contribute to this issue by updating the stubs and tests. Below you can read the detailed instructions for how to that. In case you have any questions you can ask them in this issue, or you can reach out to me at [Twitter](https://twitter.com/can__vural)

### Preparation

* Create a [GitHub user account](https://www.wikihow.com/Create-an-Account-on-GitHub) (if you haven't got one already).
* [Fork this repository on GitHub](https://help.github.com/en/articles/fork-a-repo).
* Make sure you have both [git](https://git-scm.com/downloads) as well as [Composer](https://getcomposer.org/) installed on your machine.
* [Clone your fork to your local machine](https://help.github.com/en/articles/cloning-a-repository) using your preferred git client.
* Run `composer install` from the command line in the root directory of your local clone of this repo.

#### Helpful guides
* https://www.digitalocean.com/community/tutorials/hacktoberfest-how-to-submit-your-first-pull-request-on-github
* https://www.digitalocean.com/community/tech_talks/level-up-as-an-open-source-contributor
* https://www.nexmo.com/blog/2020/10/06/opening-multiple-pull-requests-dr

### Updating stubs

[Here](https://github.com/nunomaduro/larastan/tree/master/stubs) is the directory containing all of the stubs Larastan uses. All of the stubs that need updating are already there, so you don't need to create new files. Just updating method definitions inside.

#### Steps
1. Find a Laravel method that does not exist in our stubs, or exists but the parameter type is not yet updated. For this example, we will use this method here: https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Eloquent/Builder.php#L511

[...]

PR #2017, opened 2024-08-03
route-name-type and stubs for Illuminate\Routing

szepeviktor:
- [x] Added or updated tests

Resurrect #1734 by @NiclasvanEyk

**Changes**

Similar to how view names are checked when using view, the functions that generate URLs (route, to_route, Url::signedRoute, etc.) are now checked too. This way a developer gets an error when referring to a route that does not exist.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
         "illuminate/database": "^9.52.16 || ^10.28.0 || ^11.0",
         "illuminate/http": "^9.52.16 || ^10.28.0 || ^11.0",
         "illuminate/pipeline": "^9.52.16 || ^10.28.0 || ^11.0",
+        "illuminate/routing": "^9.52.16 || ^10.28.0 || ^11.0.0",
         "illuminate/support": "^9.52.16 || ^10.28.0 || ^11.0",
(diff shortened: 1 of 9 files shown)


calebdw:
Would it be possible to solve this without introducing a new `route-string-name` custom phpdoc type? Ideally, I would like to move away from having to override so many of Laravel's doc types through the use of stubs---they're a pain to maintain (especially when you add all the different version into the mix), they easily become out of sync, and they also have a compound effect where if you override a class you tend to have to add other stubs because you can't reference classes in a stub that are not also stubs.

I've been trying to push a lot of the stubs into the framework, and they've generally been amenable to the changes. One of the things that we can't move to the framework is the `model-property` type, but I've been thinking of some ideas on how to enable the model property check **without having to override all the docs** (because there's a ton of them). This would also solve #676 as the stubs would not need to be updated. We could still keep the custom type for users to use in app-land, but we would not have to override all the framework stubs

szepeviktor:
@calebdw I am just the guy doing copy-paste.

My wish is to detect **non-existent route names**.

canvural:

PR #2019, opened 2024-08-07
fix: don't explicitly exclude blade files

### LINKED ISSUE #2018: Shouldn't excludePaths: *.blade.php
- Larastan Version: Any
- Laravel Version: Any

### Description

Larastan's extension file has below:

https://github.com/larastan/larastan/blob/c8ec6f04f800d9425be743998574f1034e2ffb3f/extension.neon#L8-L9

Since Neon config can't override the parameter, this make it impossible to analyse blade file any more even with bladestan installed.

calebdw:
Closes #2018 

Thanks!

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -5,8 +5,6 @@ parameters:
     earlyTerminatingFunctionCalls:
         - abort
         - dd
-    excludePaths:
-        - *.blade.php
     mixinExcludeClasses:
         - Eloquent
     bootstrapFiles:


szepeviktor:
This is not so easy.

calebdw:
Why not? The actual paths that are analyzed should be specified by the end user, e.g.,:

```yaml
parameters:
  level: 2
  paths:
    - app
```

szepeviktor:
It is there for a reason.

johntheteam:
> It is there for a reason.

What's the reason?
If the reason is known, it should be specified and handled in a way that allow to change by the client.

MKodde:
@szepeviktor can you be more specific? I'd love to use Bladestan in conjunction with Larastan, but this is not possible because all Blades are excluded by default in the Larastan config.

Or is there another way to achieve this goal that you know of? Or did you get any further in the meantime @johntheteam ?

szepeviktor:
I do not know why.

MKodde:
@nunomaduro and/or @canvural can you shed a ligth on this? Or @ThibaudDauce ?

canvural:
There is nothing to explain here 🤔

I'll have time tomorrow to work on Larastan. So probably tomorrow it'll be released 👍

AJenbo:
Bladestan maintainer here. It looks like the title for this PR is based on a miss understanding of statements in the refereed issue.

Bladestan works just fine with this excludePaths, it doesn't analyze the *.blade.php file directly, rather it compiles the templates in context of the view calls and analyzes the resulting PHP code. [johntheteam](https://github.com/johntheteam) did not read the documentation and made wrong assumptions about how Bladestan should be used / works: https://github.com/TomasVotruba/bladestan/issues/104

I think what [szepeviktor](https://github.com/szepeviktor) is stating in his comment is that he analyzes *.blade.php with plane PHPStan, and isn't making any statement or confirmation of this causing issues for Bladestan.

So please change the title of the PR so it doesn't confuse more people like [MKodde](https://github.com/MKodde), Larastan dosn't cause any issues for Bladestan :)

Excluding `*.blade.php` is only an issue for someone that wants to analyze uncompiled Blade files, as if they where PHP, but that really isn't a good idea since that isn't how they will be run in reality. TBH you might just want to close this PR and the issue.

calebdw:
@AJenbo, 

Thanks for the clarification. However, I still don't think that we should specifically exclude blade files from PHPStan's analysis---that is really on the end user to decide what paths should be analyzed and which paths should be ignored. Blade files are typically stored under `views/` and if a user has blade files intermingled with their source code then that is on them

AJenbo:
Yeah that is totally reasonable :+1:

AJenbo (commented):
I believe I now better understand [szepeviktor](https://github.com/szepeviktor)'s concern based on our discussion in https://github.com/larastan/larastan/issues/2018#issuecomment-2377458868.

He was simply pointing out that removing this won't make Bladestan capable of directly analyzing `.blade.php` files. However, this stems from a misunderstanding of what Bladestan does, as analyzing Blade files directly isn't possible.

So, there shouldn't be any blockers preventing you from removing this from the config. But it would probably be good to mention this change in the following release so that anyone that has blade files in there analyzed path can easily figure out how to exclude them on there end.

AJenbo (approved):

canvural:
Hi,

I totally agree that it shouldn't be in our config. It should be user's choice. But, on the other hand if we now remove it if users have their `views` or `resources` directory in the scanned paths for whatever reason it'll output lots of errors. Sure we can write it in release notes and in upgrade guide. But maybe we can wait until 3.0 to make the change. What do you think?

calebdw:
>  if users have their views or resources directory in the scanned paths for whatever reason it'll output lots of errors.

That's a big `if` given that it's really non-standard, however, this isn't really critical so I'm fine with waiting for 3.x

canvural:
People do crazy things 🤷🏽 So gonna close this one but will add it to 3.x. Thank you!

PR #2022, opened 2024-08-24
fix: wrong trigger RelationExistenceRule with return union type in relation method

JeRabix:
Sometimes i need declare union type for my relation methods, for better autocomplete in IDE. For example:
model User has method/relation `deals`, and i declare next method:

```php
public function deals(): HasMany|Deal
{
   return $this->hasMany(Deal::class);
}
```

So now i can use next code, and have autocomplete for Deal model scopes:

```php
$userDeals = $user->deals()->dealScope()->get();
```

- [x] Added or updated tests

**Changes**

When relation return union type, RelationExistenceRule can understand now this, and not trigger error.

### DIFF
--- a/src/Rules/RelationExistenceRule.php
+++ b/src/Rules/RelationExistenceRule.php
@@ -16,6 +16,7 @@ use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
+use PHPStan\Type\UnionType;
 
 use function array_map;
 use function array_merge;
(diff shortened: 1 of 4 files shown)


calebdw:
The `->deals()` method will **never** return a Deal so this is incorrect

calebdw:
For better support you should add the generics:

```php
/** @return HasMany<Deal> */
public function deals(): HasMany
{
   return $this->hasMany(Deal::class);
}
```

And for even better static support you should use custom builders instead of Model scopes

JeRabix:
> For better support you should add the generics:

Relation classes only has generics in laravel 11+, i have version 10

JeRabix:
> The `->deals()` method will **never** return a Deal so this is incorrect

But Deal class has method annotations for builder scopes

calebdw:
Larastan has stubs for the relation generics.

But regardless, IDE support is not the purpose of this tool (you can look into the IDE Helper package for that). Larastan understands the model scopes just fine and you should be adding the generics anyway given that this is static analysis and you should be defining the inner types.

The union type you have is not correct (and higher levels of phpstan will throw errors)---so this PR should not be merged

canvural:
Hi,

This is the wrong assumption. Relation methods should not return union types with models. Even a union of two relation class shouldn't be done (not sure even if it's possible)

So gonna close this one. Thank you for understanding.

Issue #2021, opened 2024-08-14
Builder doesn't have correct generic when initialized through class-string


vixducis:
- Larastan Version: 2.9.2
- Laravel Version: 11.1.1

### Description

When you initialize a `Builder` through the `::query()` method on a `class-string`, the generic on the `Builder` isn't correctly deduced from the generic of the `class-string`, and defaults to `Model`. This can be easily tested by trying to access a local scope on the query, as demonstrated by the code below. This results in a PHPStan error: `Call to an undefined method Illuminate\Database\Eloquent\Builder::test().`.

It's debatable whether this is a bug or rather a feature request.

### Laravel code where the issue was found

```php
<?php

namespace App\Models\Example;

use Illuminate\Database\Eloquent\Builder;
use Illuminate\Database\Eloquent\Model;

class Example extends Model
{
    public function scopeTest(Builder $query): void {}
}

/** @var class-string<Example> $className */
$className = Example::class;
$className::query()
    ->test()
    ->get();
```

calebdw:
:+1: I'll try to get to this in the next week or two

A similar issue I've noticed is calling `static::query()` does not handle the static instance:

```php
class AbstractModel extends Model
{
    public static function new(): static
    {
        return static::query()->create(); // fails, returns AbstractModel instead of static(AbstractModel)
    }
}

class User extends AbstractModel {}

\PHPStan\Testing\assertType('User', User::new());
```

JeRabix:
I try help for first part of this issue, but currently can't understand how fix this:

> ```
>  return static::query()->create(); // fails, returns AbstractModel instead of static(AbstractModel)
> ```


`ModelDynamicStaticMethodReturnTypeExtension`- skip all native class methods except only this: 

```php
if (in_array($name, ['get', 'hydrate', 'fromQuery'], true)) {
   return true;
}
```

when add method `new` to this except rule - it work correctly.

PR #2023, opened 2024-08-25
feat: builder doesn't have correct generic when initialized through class-string

JeRabix:
- [x] Added or updated tests

**Changes**

Changes/fixes for #2021


JeRabix:
Add stub for `Illuminate\Database\Eloquent\Model` class.
`query` method exist generic annotation only from 11 version

JeRabix:
I add Model class stub for - 9.x version too, but I doubt that this is the right decision.

stubs/9.0.0/Model.stub
calebdw, line 1: This is not needed, we already have the `stubs/common/Model.stub` and the lowest version we support is 9.x.

stubs/10.0.0/Model.stub
calebdw, line 1: There should be no need for this stub either---why was this added? It can likely just be added to the `common/Model.stub`

calebdw (changes requested):

stubs/10.0.0/Model.stub
JeRabix: I proceeded from the code that exists. And in it, many stubs are duplicated. I thought the logic is that if in version 11, for example, the declaration of a method changes, then the stub should not affect the previous version. I removed the stubs for versions 9/10, modified the common stub. Sorry, this is only my second commit to this repository. 

Let's see what the tests show now, but I think everything will be ok. I checked it on version 9 locally.

calebdw: We only create new stubs when there's something drastically different (like a new method introduced in 11.x that need its signature overriden). Unfortunately, this means that the stub has to be mostly duplicated.

stubs/common/Model.stub
```diff
@@ -66,6 +66,13 @@ abstract class Model implements \JsonSerializable, \ArrayAccess
      * @return \Illuminate\Database\Eloquent\Builder<static>
      */
     public static function with($relations);
+
+    /**
+     * Begin querying the model.
+     *
+     * @return \Illuminate\Database\Eloquent\Builder<static>
+     */
+    public static function query();
```
calebdw: This is fine, however, I'm 90% sure that this stub is not being used given that there's an extension that calculates what the return type is

calebdw (approved):
Looks good to me, thanks!

stubs/common/Model.stub
JeRabix: Without this stub method - tests in 9,10 version laravel failed

calebdw: That's because in your test you have:

```php
assertType('Illuminate\Database\Eloquent\Builder<App\User>', (new $className())->query());
```

which is calling a **static** function non-statically (and thus the static extension is not being used). This should be:

```php
assertType('Illuminate\Database\Eloquent\Builder<App\User>', (new $className())->newQuery());
```

canvural:
Thank you!

PR #1922, opened 2024-04-17
feat: support newFactory method when resolving factory

calebdw:
- [x] Added or updated tests
- [x] Documented user facing changes

Closes #1810 

**Changes**
Hello!

This PR adds support for the `newFactory` method which can point to a model that's outside the typical convention.

Thanks!

PR #2026, opened 2024-08-28
Fix get string instead of object in class detection with a complex st…

PhilippeThouvenot:
- [ ] Added or updated tests

**Changes**
get class name when an Object is return by Factory::resolveFactoryName

**Breaking changes**
None

### DIFF
--- a/src/ReturnTypes/ModelFactoryDynamicStaticMethodReturnTypeExtension.php
+++ b/src/ReturnTypes/ModelFactoryDynamicStaticMethodReturnTypeExtension.php
@@ -70,6 +70,10 @@ final class ModelFactoryDynamicStaticMethodReturnTypeExtension implements Dynami
 
         $factoryName = Factory::resolveFactoryName(ltrim($class->toCodeString(), '\\')); // @phpstan-ignore-line
 
+        if (is_object($factoryName)) {
+            $factoryName = get_class($factoryName);
+        }
+
         if (class_exists($factoryName)) {
             return new ModelFactoryType($factoryName, null, null, $isSingleModel);
         }


calebdw:
I've already fixed the issues with the factory resolution in #1922, please see if that does not solve your issue

PhilippeThouvenot:
@calebdw I've tried your package, and unfortunaly it's not fixed : 
```
Internal error: class_exists(): Argument #1 ($class) must be of type string, App\Domain\Project\Database\Factories\TaskFactory given while analysing file
     /root/digit/app/Domain/Project/Seeders/TaskSeeder.php
```

calebdw:
What does your factory and model look like, and how are you calling it in the seeder?

calebdw:
@PhilippeThouvenot, 

`Factory::resolveFactoryName()` is supposed to return a **class-string** not an object:

https://github.com/laravel/framework/blob/b9cf7d3217732e9a0fa4f00e996b3f9cc5bf7abd/src/Illuminate/Database/Eloquent/Factories/Factory.php#L857-L878

If you're getting an object then it's a bug with a custom `$factoryNameResolver` closure

PhilippeThouvenot:
I've made this repo for you https://github.com/PhilippeThouvenot/larastan-test-newfactory
You can test it.
I figure the same problem with ide helper.

calebdw:
@PhilippeThouvenot,

This is due to a bug in the LaravelDDD package. The following line below:

https://github.com/lunarstorm/laravel-ddd/blob/56eeddd572373225cf608f32de1a027e55b6abbc/src/Support/DomainAutoloader.php#L113

should be:

```diff
-               return DomainFactory::factoryForModel($modelName);
+               return DomainFactory::resolveFactoryName($modelName);
```

You can submit a PR to that package and close this one since the method should return a class-string instead of an object.

PhilippeThouvenot:
Ok thank you, I will do it.

PR #2027, opened 2024-08-29
Add support for model attributes property

maartenpaauw:
- [ ] Added or updated tests

**Changes**

The `$attributes` array is used to configure the default model attribute values. This PR adds its type as doc block annotation to the `Model.stub` file.

**Breaking changes**

### DIFF
--- a/stubs/common/Model.stub
+++ b/stubs/common/Model.stub
@@ -8,6 +8,11 @@ namespace Illuminate\Database\Eloquent;
 abstract class Model implements \JsonSerializable, \ArrayAccess
 {
     /**
+     * @var array<string, mixed>
+     */
+    protected $attributes = [];
+
+    /**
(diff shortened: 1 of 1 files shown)


maartenpaauw:
I don’t think my change let all those actions fail. Am I right?

Becky1175:
Wydm

calebdw:
@maartenpaauw, 

> I don’t think my change let all those actions fail. Am I right?

No you're good, I'm getting these same errors on my repo. This is likely due to a PHPStan update (I'll create a PR to fix).

szepeviktor:
Nightly CI runs: https://github.com/larastan/larastan/actions/runs/10624646059

canvural:
Thank you!

AJenbo:
Thanks for this change :)

PR #2028, opened 2024-08-30
fix: incorrect phpdocs and stubs

calebdw:
Hello!

This fixes some incorrect phpdocs and stubs as a result of PHPStan 1.12.0 having stricter type checks.

This fixes what can be fixed in the failing nightly ci jobs, the other failures in the job will have to be fixed by PHPStan (see my https://github.com/phpstan/phpstan/issues/11591#issuecomment-2321280743).

Thanks!


calebdw:
@szepeviktor, PHPStan just released a patch and this fixes the ci jobs---can this be merged now?

Thanks!

calebdw:
@canvural, this fix is needed so that all of the ci jobs will start passing again (with the exception of L11)

canvural:
Thank you!

PR #2029, opened 2024-09-01
Prevent error Argument #2 ($array) must be of type array

sangnguyenplus:
This PR will fix the following error when checking model macros:

```
Uncaught TypeError: array_key_exists(): Argument #2 ($array) must be of type array,
 null given in vendor/larastan/larastan/src/Methods/MacroMethodsClassReflectionExtension.php:135
```

This PR won't have any breaking changes.

The return value of `$refProperty->getValue()` can be null.

image

So with this change, it will be ignored if `$refProperty->getValue()` is null.


calebdw:
Thanks for this, can you please give an example of code that caused this to fail? Writing a test for this would be good too

canvural:
Hi,

Cannot merge this without a test. Please try to add a test case.

canvural:

PR #2030, opened 2024-09-02
fix: properly handle static called on type for model query

calebdw:
**Changes**

Hello!

This PR resolves the issue noted in https://github.com/larastan/larastan/issues/2021#issuecomment-2289617242

Thanks!


canvural:
Thank you!

PR #2031, opened 2024-09-02
[3.x] refactor: create FileHelper to clean up repeated logic

### LINKED ISSUE #1954: Support migrations in modules more flexible
# Feature Request

Larastan Version: `v2.9.6`
`--level` used: 9

Would it be possible to add wildcard support to config option `databaseMigrationsPath`, please?
I would be happy if i don't have to add all migration paths of all modules manually.

**Additional context**

I'm using Laravel 11 with `"nwidart/laravel-modules": "^11.0"`
My migrations in modules weren't scanned, so I found two issues here, #1468 and #1266
The latter one helped, after I found out that it doesn't support `*` wildcard.

I hoped I can write something like `Modules/*/database/migrations` to automatically add all migrations from all modules.

calebdw:
**Changes**

Hello!

~This PR allows wildcards to be specified in the migration/schema paths which closes #1954.~

This PR creates an internal FileHelper class that cleans up all the repeated file globing logic.

It also has a small performance tweak in the NoEnvCallsOutsideOfConfigRule---there's no need to reglob the directories on every `env` check

Thanks!

### DIFF
--- a/docs/custom-config-parameters.md
+++ b/docs/custom-config-parameters.md
@@ -12,6 +12,8 @@ By default, the default Laravel database migration path (`database/migrations`) 
 
 You can give absolute paths, or paths relative to the PHPStan config file.
 
+Paths with wildcards are also supported (passed to `glob` function).
+
 ### Example
 ```neon
 parameters:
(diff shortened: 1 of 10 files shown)


calebdw:
It looks like https://github.com/larastan/larastan/pull/2271 was duplicate PR that was merged and conflicted with this, so this has morphed into more of a refactoring PR

PR #2036, opened 2024-09-11
[3.x] docs: add docs for custom model builders, factories, and collections

calebdw:
Hello!

This documents custom model builders, factories, and collection as requested in https://github.com/larastan/larastan/discussions/2000.

Thanks!

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -11,33 +11,222 @@ The first generic type is the getter return type, and the second is the setter a
 #### Examples
 
 ```php
+<?php
 /** @return Attribute<string[], string[]> */
 protected function scopes(): Attribute
 {
-	return Attribute::make(
-		get: fn (?string $value) => is_null($value) ? [] : explode(' ', $value),
-		set: function(array $value) {
-			$set = array_unique($value);
-			sort($set);
-			return ['scopes' => implode(' ', $set)];
-		}
-	);
+    return Attribute::make(
+        get: fn (?string $value) => is_null($value) ? [] : explode(' ', $value),
+        set: function(array $value) {
+            $set = array_unique($value);
+            sort($set);
+            return ['scopes' => implode(' ', $set)];
+        }
+    );
 }
 ```
 
 ```php
+<?php
 /** @return Attribute<bool, never> */
 protected function isTrue(): Attribute
 {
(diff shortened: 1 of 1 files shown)


phh:
Using the [custom builder](https://github.com/larastan/larastan/blob/bdfb2131b02a1bcfb78e253abd4db761acd2cc8d/docs/features.md#custom-model-builders) example I'm getting:

```
PHPDoc tag @return contains generic type App\Builders\UserBuilder<App\Models\User> but class
   1      App\Builders\UserBuilder is not generic.
```

Adding builder:
```
/** @extends Builder<User> */
```

It works if I add the template:

```
/**
 * @template TModelClass of User
 *
 * @extends Builder<TModelClass>
 */
```

calebdw:
If your builder is not generic, then you need to return just `App\Builders\UserBuilder` instead of `App\Builders\UserBuilder<App\Models\User>`---or you can just make the builder generic

phh:
https://github.com/larastan/larastan/pull/2036/files#diff-aff15ced1fe37506f41f2e1d93a4d8312529259f7803f8634ffb5eac4de47823R53

Since this one extends `Illuminate\Database\Eloquent\Builder` which is generic, shouldn't the `UserBuilder` in this case also be generic?

calebdw:
No it doesn't have to be, especially since it's a **User** builder that will likely only be used with Users. You would really only need to make it generic if you plan on having models that extend from User

docs/features.md
```diff
@@ -11,33 +11,220 @@ The first generic type is the getter return type, and the second is the setter a
...
+use Illuminate\Database\Eloquent\HasBuilder;
+use Illuminate\Database\Eloquent\Model;
+
+/** @extends Builder<User> */
+class UserBuilder extends Builder
+{
+    /** @return $this */
+    public function active(): static
+    {
+        return $this->where('active', true);
```
phh: This will return `static(UserBuilder)`.

Adding the where and then returning `$this` will instead return `$this(UserBuilder)` as expected.
```suggestion
        $this->where('active', true);

        return $this;
```

lloricode-halcyon (approved):

AJenbo:
Thanks

PR #2042, opened 2024-09-14
fix: relation class forwarding

### LINKED ISSUE #2040: Builder methods are not forwarded on Relation object
- Laravel Version: ^11.23.0

### Description
After the `Collection::load*` and `Builder::with*` methods were updated in Laravel v11.23.0 to specify that a `Relation` is passed to the closures there have been reports that Builder methods called on the `Relation` object are not being forwarded and are reported as not found.

This is likely due to checking that the classReflection is a subclass of `Relation` in the forwarding extension which returns false for the base Relation

```php
function test(Builder $query): void
{
    $query->with(['foo' => fn (Relation $q) => $q->select('id')]); // fails with `select not found on Relation`
}
```

calebdw:
**Changes**

Hello!

This fixes builder method forwarding on the base `Relation` class and closes #2040.

I discovered that the `ClassReflection` class has a handy `is` method which cleans up some code.

Thanks!

### DIFF
--- a/src/Methods/EloquentBuilderForwardsCallsExtension.php
+++ b/src/Methods/EloquentBuilderForwardsCallsExtension.php
@@ -66,7 +66,7 @@ final class EloquentBuilderForwardsCallsExtension implements MethodsClassReflect
      */
     private function findMethod(ClassReflection $classReflection, string $methodName): MethodReflection|null
     {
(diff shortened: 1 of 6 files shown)


canvural:
Thank you!

PR #2047, opened 2024-09-18
Support Manager's createDriver() return types

### LINKED ISSUE #2046: Managers: driver methods return type not used
- Larastan Version: 2.9.8
- Laravel Version: 10

### Description
When dealing with Managers, the class of the object returned by the `->driver()` method is used instead of the return type defined in the `createFooDriver()` method.

If a contract is defined in the return type of the `createFooDriver`, larastan will not consider that the manager returns the contract, but the actual implementation itself instead.

Example:

If we have the following Manager, Contract and driver:

```php
class Storage extends Manager
{
    public function getDefaultDriver(): string {
      return 'file';
    }

    public function createFileDriver(): StorageContract {
      return app(FileDriver::class);
    }
}

interface StorageContract
{
    public function type(): string;
}

class FileDriver implements StorageContract
{
    public function type(): string {
      return 'file';
    }

    public function helloWorld(): string {
       return 'Hello World';
     }
}
```
Then I expect the following code to fail with larastan:

```php
app(Storage::class)->helloWorld();
```

Instead, this code pass the checks.

## Solution
I believe the following line of code can be improved : https://github.com/larastan/larastan/blob/2.x/src/Methods/Pipes/Managers.php#L39

Instead of simply using the driver's class, we should get the return type of the Manager's `'create'.Str::studly($concrete->getDefaultDriver()).'Driver';` method.

I can submit a PR with a fix.

Thank you very much for maintaining this awesome project :+1:

jf-m:
- [x] Added or updated tests

This PR resolves https://github.com/larastan/larastan/issues/2046

**Changes**

The return type of Manager's `createXXXDriver()` methods is now used instead of the actual returned class type. If not possible or unavailable, the actual returned class type is used as a fallback.

This change gives the option for `createXXXDriver()` to have a contract as a return type. If so, only the public methods of the contract will be considered valid when using the Manager's `__call` magic methods.

**Breaking changes**

People who have a Manager that returns a contract, and who were using public methods that are not part of the contract but part of the runtime implementation, will have errors showing up after this release.

I've added all the details to `UPGRADE.md`.

**Note**

This is my first PR on this project, I've followed the contributing guide, and I've tried to write code that matches the codebase. But of course, I might have missed stuff or things I've done could have been done in a better way. If so, don't hesitate to tell me what could be improved, I'll amend.


src/Methods/Pipes/Managers.php
```diff
@@ -23,25 +26,38 @@ public function handle(PassableContract $passable, Closure $next): void
...
+
+            $createDefaultDriverMethod = 'create' . Str::studly($concrete->getDefaultDriver()) . 'Driver';
+            if ($classReflection->hasNativeMethod($createDefaultDriverMethod)) {
+                $createDefaultDriverMethod = $classReflection->getNativeMethod($createDefaultDriverMethod);
+                try {
+                    $methodReturnType = ParametersAcceptorSelector::selectSingle($createDefaultDriverMethod->getVariants())->getReturnType();
+                    $class            = $methodReturnType->getObjectClassNames()[0] ?? null;
+                } catch (ShouldNotHappenException) {
+                    // ..
+                }
```
calebdw: You shouldn't need to catch the ShouldNotHappenException here:

```suggestion
                $class = ParametersAcceptorSelector::selectSingle($createDefaultDriverMethod->getVariants())
                    ->getReturnType()
                    ->getObjectClassNames()[0] ?? null;
```

tests/application/app/DummyManagerReturnsImpl.php
calebdw, line 1: You can move all these classes inside the `tests/Integration/data/managers.php` file

tests/application/app/DummyManagerReturnsImpl.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace App;
+
+use Illuminate\Support\Manager;
+
+/**
+ * @mixin DummyDriverContract
+ */
```
calebdw: You shouldn't need this mixin, this is the whole reason why we have the extension

tests/application/app/DummyManagerReturnsContract.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace App;
+
+use Illuminate\Support\Manager;
+
+/**
+ * @mixin DummyDriverContract
+ */
```
calebdw: Don't need the mixin

tests/application/app/DummyManagerReturnsImpl.php
jf-m: Ah, yes, I must have added it by automatism =_=

calebdw (approved):
Looks good to me, thanks!

jf-m:
Some PHPUnit errors are not related to this PR I think. I was having them before even starting the changes. 
It's the ones about `App\TransactionCollection`.
Same for the PHPStan issues

calebdw:
Yes there's open PRs to fix the failing tests

jf-m:
Hi @calebdw

Do you think we can merge this PR in the 3.x?

calebdw:
That's up to @canvural , I'm just a contributor

jf-m:
Thanks! @canvural what's your thoughts on this PR?

canvural:

PR #2048, opened 2024-09-23
[3.x] feat: support dynamic relation closures

calebdw:
**Changes**

Hello!

This PR adds support for dynamic relation closures as shown in @canvural's tweet: https://x.com/can__vural/status/1793999106022711748

Below are some examples from the test suite:

```php
    User::has('accounts', callback: function ($query) {
        assertType('Illuminate\Database\Eloquent\Builder<App\Account>', $query);
    });

    User::has('accounts.posts', callback: function ($query) {
        assertType('App\PostBuilder<App\Post>', $query);
    });

    Address::hasMorph('addressable', [User::class, Team::class], callable: function ($query) {
        assertType('App\ChildTeamBuilder|Illuminate\Database\Eloquent\Builder<App\User>', $query);
    });
```

Thanks!

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -534,6 +534,19 @@ services:
             - phpstan.broker.dynamicFunctionReturnTypeExtension
 
     -
+        class: Larastan\Larastan\Parameters\EloquentBuilderRelationParameterExtension
+        tags:
+            - phpstan.methodParameterClosureTypeExtension
+
+    -
+        class: Larastan\Larastan\Parameters\ModelRelationParameterExtension
+        tags:
+            - phpstan.staticMethodParameterClosureTypeExtension
+
+    -
+        class: Larastan\Larastan\Parameters\RelationClosureHelper
+
+    -
         class: Larastan\Larastan\ReturnTypes\AppMakeHelper
 
     -
--- a/src/Methods/BuilderHelper.php
+++ b/src/Methods/BuilderHelper.php
@@ -17,13 +17,18 @@ use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Generic\GenericObjectType;
+use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+use PHPStan\Type\TypeWithClassName;
 use PHPStan\Type\VerbosityLevel;
 
 use function array_key_exists;
 use function array_shift;
+use function collect;
 use function count;
 use function in_array;
+use function is_string;
 use function preg_split;
 use function substr;
 use function ucfirst;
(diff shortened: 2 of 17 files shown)


calebdw:
@canvural, can we please try to get this merged? This would add a ton of value and remove the need to manually override the query types inside the closures

canvural:
> This would add a ton of value and remove the need to manually override the query types inside the closures

I know. That's why I implemented those extensions in PHPStan 😊

I also had the implementation ready here for a long time, but unfortunately there is still a bug in PHPStan that I couldn't fix which would add lots of false positive errors with the extension.

Also looking at the code I have a simpler solution. But that's not the main point we could discuss those parts. But I cannot merge this until the bug in PHPStan is fixed.

I was under the impression that since I implemented the extensions in PHPStan, it'd be logical to expect I was gonna implement them here too. This is a perfect example of how a prior discussion would save us time. But now unfortunately you spent the time to implement it but I think I'm gonna have to close it. I really don't like dismissing other people's efforts like I said in that other comment, so I'm sorry.

calebdw:
@canvural, thanks for responding.

> there is still a bug in PHPStan that I couldn't fix which would add lots of false positive errors with the extension.

Could you please give some detail on the PHPStan bug (and maybe even point to an issue)?

> Also looking at the code I have a simpler solution. But that's not the main point we could discuss those parts.

I would be interested in seeing your simpler solution---but does your solution pass all my tests? Besides all the obvious simple cases, I also correctly handle:
- nested relationships (`->whereHas('accounts.posts', ...)`)
- unions between both strings and relationship objects (`->whereHas('accounts.posts'|BelongsTo<...>, ...)`)
- methods called on a builder union (e.g., `$userOrTeamBuilder->has('transactions', ...)` which means that the `transactions` relationship has to be checked against both the `User` and the `Team` models---and those resultant relation models might not be the same which means the inner `$query` could be union)
- properly resolving custom builders inside the closure (and multiple models are handled correctly)
- the `withWhereHas` method requiring the inner `$query` type to be a `Builder|Relation` union
  ```php
  User::query()->withWhereHas('accounts.posts', function (Builder|Relation $query) {
    assertType('App\PostBuilder<App\Post>|Illuminate\Database\Eloquent\Relations\BelongsToMany<App\Post, App\Account>', $query);
  });
  ```

I have a feeling that many of the cases haven't been accounted for and your solution would grow to be equally complex as mine should you support them all (but I could be wrong).

> I was under the impression that since I implemented the extensions in PHPStan, it'd be logical to expect I was gonna implement them here too.

Sure, but your tweet is from May 2024 and I submitted this PR in Sept. 2024. I figured you just got busy with more important things and so decided to try and help you out by taking a crack at it.

> But now unfortunately you spent the time to implement it but I think I'm gonna have to close it.

You closed this PR, but you can't take away the knowledge and experience I gained along the way :upside_down_face: 

Best,

PR #2052, opened 2024-10-02
Add failedValidation() to FormRequest.stub

shaedrich:
- [ ] Added or updated tests

**Changes**

* Added [`failedValidation()`](/laravel/framework/blob/11.x/src/Illuminate/Foundation/Http/FormRequest.php#L218-L221) to FormRequest.stub
* Changed type of `$validator` argument changed to [`\Illuminate\Validation\Validator`](/laravel/framework/blob/11.x/src/Illuminate/Validation/Validator.php)
* Therefore, calling [`$validator->getException()`](/laravel/framework/blob/11.x/src/Illuminate/Validation/Validator.php#L1518-L1521 inside the method doesn't result in an error anymore, that `getException()`) is not defined in [`\Illuminate\Contracts\Validation\Validator`](/laravel/framework/blob/11.x/src/Illuminate/Contracts/Validation/Validator.php)

**Breaking changes**

_n/a_

### DIFF
--- a/stubs/common/Foundation/Http/FormRequest.stub
+++ b/stubs/common/Foundation/Http/FormRequest.stub
@@ -21,4 +21,14 @@ class FormRequest
      * @return ($key is null ? array<string, mixed> : mixed)
      */
     public function validated($key = null, $default = null);
+
+    /**
+     * Handle a failed validation attempt.
+     *
+     * @param  \Illuminate\Validation\Validator  $validator
+     * @return void
+     *
+     * @throws \Illuminate\Validation\ValidationException
+     */
+    protected function failedValidation(Validator $validator);
 }


calebdw:
There should be no need to add a stub for a `protected` method---this will never be called by the user.


If you are getting this error then you are trying to analyze the framework code **which you should not be doing**.

shaedrich:
This concerns classes extended from `FormRequest`. In these extended classes, you sometimes need to overwrite `failedValidation()`, which then results in said error.

calebdw:
The framework is violating their own contract---you should open a PR at the framework instead of here. Either the parameter needs to change to `\Illuminate\Validation\Validator` or the `getException` method needs to be added to the contract.

Alternatively, you can just narrow the type yourself in the method:

```php
    protected function failedValidation(ValidatorContract $validator)
    {
        assert($validator instanceof Validator);
        $exception = $validator->getException();
        // ...
    }
```

I still see no reason for a protected method to be added to our stubs

shaedrich:
> you should open a PR at the framework instead of here

Too bad, they barely take PRs from strangers :/ But of course, that's not your fault. Guess, I have to go with the assert solution, then. Thanks, nonetheless.

PR #2054, opened 2024-10-03
remove unnecessary use of `ParametersAcceptorSelector::selectSingle`

staabm:
- [ ] Added or updated tests

**Changes**

Simplify the migration path to PHPStan 2.x in which `ParametersAcceptorSelector::selectSingle` is deprecated.

see https://phpstan.org/developing-extensions/dynamic-return-type-extensions

grafik

**Breaking changes**

none


staabm:
fixed cs

szepeviktor:
Thank you, Markus.

PR #2060, opened 2024-10-08
feat: useContainerCommandLoader parameter

wslawski-printify:
- [x] Added or updated tests

**Changes**

- Adds option to use `ContainerCommandLoader` when resolving command options/arguments etc, so when using `#[AsCommand]` attribute on commands it will work as expected and resolve options/arguments properly

**Breaking changes**

No breaking changes, added new parameter with false value

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -26,7 +26,6 @@ parameters:
     checkPhpDocMissingReturn: false
     checkUnusedViews: false
     checkModelAppends: false
-    useContainerCommandLoader: false
 
 parametersSchema:
     checkOctaneCompatibility: bool()
@@ -43,7 +42,6 @@ parametersSchema:
     checkModelProperties: bool()
     checkUnusedViews: bool()
     checkModelAppends: bool()
-    useContainerCommandLoader: bool()
 
 conditionalTags:
     Larastan\Larastan\Rules\NoEnvCallsOutsideOfConfigRule:
--- a/extension.neon
+++ b/extension.neon
@@ -26,6 +26,7 @@ parameters:
     checkPhpDocMissingReturn: false
     checkUnusedViews: false
     checkModelAppends: false
+    useContainerCommandLoader: false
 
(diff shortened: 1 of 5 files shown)


szepeviktor (approved):

wslawski-printify:
Changed definition to one line to fix phpcs @szepeviktor all other failing checks seems not that change related to me.

src/Internal/ConsoleApplicationResolver.php
```diff
@@ -31,6 +35,10 @@ public function findCommands(ClassReflection $classReflection): array
 
         $commands = [];
 
+        if ($this->useContainerCommandLoader) {
+            $consoleApplication->setContainerCommandLoader();
+        }
```
calebdw: Why do we really need another config option? I think this can just be appended on the `getApplication()` method below:
```php
        if ($this->application === null) {
            $this->application = (new Application(app(Container::class), app(Dispatcher::class), app()->version()))
                ->setContainerCommandLoader();
        }
```

This is essentially what the framework does, and I see no reason why anyone would want to disable this:

https://github.com/laravel/framework/blob/bc5bd660b0a3079e3fb7f392bfac06e41d80b64b/src/Illuminate/Foundation/Console/Kernel.php#L531-L541

Jurigag: Well probably it makes sense, wanted to keep it without it to make sure we don't break anything but I will change it.

extension.neon
calebdw, line 1: Make sure to clean up these changes as well (same as the laravel-test.sh file)

Jurigag: Done :)

Jurigag: Oh sorry, othter github account :D

wslawski-printify:
Hey @szepeviktor @calebdw when it can be merged?

szepeviktor:
Not by me!

PR #2064, opened 2024-10-10
Support MailMessage in UsedEmailViewCollector

### LINKED ISSUE #2063: CheckUnusedViews don't check method view in MailMessage class
- Larastan Version: 2.9.8
- Laravel Version: 11.26.0

### Description
I am using views in my notifications with the MailMessage class, but phpstan doesn't recognize that these views are being used there and throws an error with the rule checkUnusedViews: true.

```php

class TournamentConfirmationRegistration extends Notification
{
    /**
     * @return non-empty-list<string>
     */
    public function via(): array
    {
        return [MailChannel::class];
    }

    public function toMail(): MailMessage
    {
        return (new MailMessage())->view('mails.tournament-confirmation-registration');
    }
}

```

error: 
```
  Line   resources/views/mails/tournament-confirmation-registration.blade.php  
  0      This view is not used in the project.                                 
```

koraga:
- [x] Added or updated tests

Resolves #2063

**Changes**

Add check MailMessage is super class in UsedEmailViewCollector

**Breaking changes**

No breaking changes

### DIFF
--- a/tests/Rules/data/FooController.php
+++ b/tests/Rules/data/FooController.php
@@ -4,6 +4,7 @@ namespace Tests\Rules\Data;
 
 use Illuminate\Contracts\View\View;
 use Illuminate\Mail\Mailable;
+use Illuminate\Notifications\Messages\MailMessage;
 use Illuminate\Support\Facades\Route;
 use Illuminate\View\Factory;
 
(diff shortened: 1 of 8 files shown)


tests/Rules/data/FooController.php
```diff
@@ -30,7 +31,20 @@ public function notExisting(): View
     }
 }
 
-class FooEmail extends Mailable
+class FooMailable extends Mailable
```
calebdw, line 34: The mailable and the mail message should use different views or the test is not going to test anything

calebdw (approved):
Looks great, thanks!

canvural:
Thank you!

PR #2068, opened 2024-10-13
feat: add support for chunkById, chunkByIdDesc, and chunkMap methods

### LINKED ISSUE #2065: Access to an undefined property TModel of Illuminate\Database\Eloquent\Model
- Larastan Version: 2.9.8
- Laravel Version: 11.27.2

### Description
After updating both laravel and larastant I'm getting the following error. (This was not happening before the update)
Screenshot from 2024-10-11 12-07-51

I think the error is related to the fact that the checks thinks I'm using a different class I'm using App\Model\Exports, but larastan says undefined property on Illuminate\Database\Eloquent\Model, that is the parent class.

```php
<?php

namespace App\Console\Commands;

use App\Models\Export;
use Carbon\Carbon;
use Illuminate\Console\Command;

class ExportsMaintenanceCommand extends Command
{
    /**
     * The name and signature of the console command.
     *
     * @var string
     */
    protected $signature = 'exports:maintenance';

    /**
     * The console command description.
     *
     * @var string
     */
    protected $description = 'Remove export files stored for more than 7 days.';

    /**
     * Create a new command instance.
     *
     * @return void
     */
    public function __construct()
    {
        parent::__construct();
    }

    /**
     * Execute the console command.
     *
     * @return int
     */
    public function handle()
    {
        // Delete files
        Export::whereDate('created_at', '<', Carbon::now()->subDays(7))
            ->where('status', 'finished')
            ->whereNotNull('filepath')
            ->chunkById(100, function ($exports) {
                foreach ($exports as $export) {
                    if (file_exists($export->filepath)) {
                        unlink($export->filepath);
                    }

                    $export->update([
                        'filepath' => null,
                        'filename' => null,
                        'filemime' => null,
                    ]);
                }
            });

        return 1;
    }
}
```

calebdw:
Hello!

This closes #2065.

Thanks!

### DIFF
--- a/stubs/10.0.0/EloquentBuilder.stub
+++ b/stubs/10.0.0/EloquentBuilder.stub
@@ -536,6 +536,39 @@ class Builder
      * @return bool
      */
     public function chunk($count, $callback);
+
+    /**
+     * Chunk the results of a query by comparing IDs.
+     *
+     * @param  int  $count
+     * @param  callable(\Illuminate\Database\Eloquent\Collection<int, TModelClass>, int): mixed  $callback
+     * @param  string|null  $column
+     * @param  string|null  $alias
+     * @return bool
+     */
+    public function chunkById($count, callable $callback, $column = null, $alias = null);
+
+    /**
+     * Chunk the results of a query by comparing IDs in descending order.
+     *
+     * @param  int  $count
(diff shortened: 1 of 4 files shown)


canvural:
Thank you!

AJenbo:
Thanks

PR #2077, opened 2024-10-19
fix: do not throw if application could not be bootstrapped

NickSdot:
Since https://github.com/larastan/larastan/pull/1983 wasn't merged yet, and someone else [ran in the same issue](https://github.com/larastan/larastan/pull/1983#issuecomment-2423574138) today, I submit this one as an alternative attempt. 

Instead of using env, this PR seeks to not throw and silcently move on to bootstrappers defined in the PhpStan config.

**Summary of the problem:**

- PhpStan loads extension configs first.
- The Larastan bootstrap.php (not PhpStan) file has hard-coded paths. If Larastan cannot bootstrap the Laravel application it throws `Could not find Laravel bootstrap file nor Testbench is installed. Install orchestra/testbench if analyzing a package`.
- Since the PhpStan config would have been loaded after this step, we never reach this stage to handle our own bootstrapping.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -18,16 +18,16 @@ if (file_exists($applicationPath = getcwd().'/bootstrap/app.php')) { // Applicat
     $app = require $applicationPath;
 } elseif (trait_exists(CreatesApplication::class)) { // Packages
     $app = ApplicationResolver::resolve();
-} else {
-    throw new Exception('Could not find Laravel bootstrap file nor Testbench is installed. Install orchestra/testbench if analyzing a package.');
 }
 
-if ($app instanceof Application) {
-    $app->make(Kernel::class)->bootstrap();
-} elseif ($app instanceof LumenApplication) {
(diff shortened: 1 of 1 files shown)


szepeviktor:
`2.x` is in bad condition. Sorry.

bootstrap.php
```diff
@@ -18,16 +18,16 @@
...
-} else {
-    throw new Exception('Could not find Laravel bootstrap file nor Testbench is installed. Install orchestra/testbench if analyzing a package.');
 }
 
-if ($app instanceof Application) {
-    $app->make(Kernel::class)->bootstrap();
-} elseif ($app instanceof LumenApplication) {
-    $app->boot();
-}
+if(isset($app)) {
```
szepeviktor:
```suggestion
if (isset($app)) {
```

canvural:
I think I like this approach better. Thank you!

PR #2081, opened 2024-10-21
fix: improve PHPDoc for find method of QueryBuilder to support FETCH_ASSOC

kayw-geek:
The current PHPDoc for the find() method declares the return type as object|null. 

This likely stems from Laravel’s default fetchMode for the Database\Connection property being set to `PDO::FETCH_OBJ.` However, when users customize the `Connection` to use `PDO::FETCH_ASSOC`, this can result in numerous errors, as the return type differs. 

My attempt to override Larastan’s stub with a custom file was unsuccessful because PHPStan does not support overriding stub files. 

This limitation highlights the need for a more flexible solution to accommodate different fetch modes without causing type-related errors.

### DIFF
--- a/stubs/10.0.0/QueryBuilder.stub
+++ b/stubs/10.0.0/QueryBuilder.stub
@@ -180,7 +180,7 @@ class Builder
      *
      * @param  int|string  $id
      * @param  array<string|\Illuminate\Contracts\Database\Query\Expression>  $columns
-     * @return object|null
+     * @return array<array-key, mixed>|object|null
      */
     public function find($id, $columns = ['*']);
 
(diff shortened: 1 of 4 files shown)


src/Properties/ModelRelationsExtension.php
```diff
@@ -127,6 +127,7 @@ private function getRelationTypeBeforeL11(ExtendedMethodReflection $method): Typ
     {
         $returnType = $method->getVariants()[0]->getReturnType();
 
+        \PHPStan\dumpType('12333333333333333');
```
szepeviktor: @kayw-geek It'll be a literal string.

kayw-geek: Sorry, I forgot to delete it.

canvural:
Thank you! It was mixed before actually 😅  And was recently changed to `object` and I knew we had a reason to have mixed but forgot it. Makes sense now 👍🏽

PR #2092, opened 2024-10-29
Illuminate\Database\Query\Builder::get() stub

rudiedirkx:
- [ ] Added or updated tests
- <s>Documented user facing changes</s> none

**Changes**

Adds stub function for `Illuminate\Database\Query\Builder::get()` to make the return type better: 'Collection of objects' instead of just 'Collection (of mixed)'.

**Breaking changes**

No breaking changes.

----

Tests?

### DIFF
--- a/stubs/11.0.0/QueryBuilder.stub
+++ b/stubs/11.0.0/QueryBuilder.stub
@@ -7,6 +7,14 @@ class Expression {}
 class Builder
 {
     /**
+     * Execute the query as a "select" statement.
+     *
+     * @param  list<string>|string  $columns
+     * @return \Illuminate\Support\Collection<int, object>
+     */
+    public function get($columns = ['*']);
+
+    /**
      * Add a basic where clause to the query.
      *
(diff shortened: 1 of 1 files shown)


calebdw:
This kind of fix should really be PR'd to the framework first to fix it at the source

rudiedirkx:
Does Laravel accept `list<string>` specificity (instead of just `array`)? Is that general phpdoc enough? Or psalm-like enough?

(Wow, that's a lot of tests! 😆)

calebdw:
Should be fine, but you could also just use `array<int, string>`

canvural:
Hi,

This looks good but it needs to have tests. You can add it in `query-builder.php` Also the stub needs to be added to other versions.

rudiedirkx:
In Laravel soon 😮 https://github.com/laravel/framework/pull/53567 Good advice @calebdw

canvural:
For this kind of changes we can use stub files. That way we can have the changes faster and also in other Laravel versions too (where it makes sense) I added it to the stubs. Thank you for the PR for the Laravel.

PR #2099, opened 2024-11-03
fix: don't report error when using Builder<*>

### LINKED ISSUE #2074: Regression of Eloquent Query Builder where method
- Larastan Version: 2.9.9
- Laravel Version: 11.28.1

### Description
I'm getting the following error with my static analysis today:
```
  :51    Parameter #1 $column of method Illuminate\Database\Eloquent\Builder<*>::where() expects array<int|model property of never, mixed>|(Closure(Illuminate\Database\Eloquent\Builder<*>):  
         mixed)|Illuminate\Contracts\Database\Query\Expression|model property of never, 'record_status' given.   
```

Indicating that I'm sending an incorrect value to the where method. However I'm sending a simple string for the column name. See code below.

```php
        $query = $this->getBasePinsQuery()
            ->where(static function (Builder $query) {
                $query->where('record_status', '!=', Pin::RECORD_STATUS_ACTIVE)
                    ->orWhereNull('record_status');
            })
            ->whereNull('deleted_at')
            ->whereRaw("to_tsvector('simple', public.f_concat_ws(' ', address, city, state, zip))
                @@ plainto_tsquery('simple', ?)", $search);

        return $query->limit(20)->get();
```
The error refers to line 51, which is `$query->where('record_status', '!=', Pin::RECORD_STATUS_ACTIVE)`

However, if I replace the where that passes in the closure with a simple where as such:

```php
        $query = $this->getBasePinsQuery()
            ->where('record_status', '')
            ->whereNull('deleted_at')
            ->whereRaw("to_tsvector('simple', public.f_concat_ws(' ', address, city, state, zip))
                @@ plainto_tsquery('simple', ?)", $search);

        return $query->limit(20)->get();
```

I get the same error:

```
  :50    Parameter #1 $column of method Illuminate\Database\Eloquent\Builder<*>::where() expects array<int|model property of never, mixed>|(Closure(Illuminate\Database\Eloquent\Builder<*>):  
         mixed)|Illuminate\Contracts\Database\Query\Expression|model property of never, 'record_status' given. 
```

Line 50, which would be `->where('record_status', '')`

I downgraded larastan to 2.9.8 and it started working again, so the regression must be with this morning's release of 2.9.9

calebdw:
Hello!

This closes #2074

Thanks!

### DIFF
--- a/src/Types/ModelProperty/GenericModelPropertyType.php
+++ b/src/Types/ModelProperty/GenericModelPropertyType.php
@@ -65,7 +65,7 @@ class GenericModelPropertyType extends StringType
             }
 
             if (count($genericType->getObjectClassNames()) < 1) {
-                return AcceptsResult::createNo();
+                return AcceptsResult::createYes();
             }
 
             if (str_contains($givenString, ' as ')) {
--- a/tests/Integration/IntegrationTest.php
+++ b/tests/Integration/IntegrationTest.php
@@ -21,6 +21,7 @@ class IntegrationTest extends PHPStanTestCase
     {
         self::getContainer();
 
+        yield [__DIR__ . '/data/bug-2074.php'];
         yield [__DIR__ . '/data/test-case-extension.php'];
         yield [__DIR__ . '/data/model-builder.php'];
         yield [__DIR__ . '/data/model-properties.php'];
--- a/tests/Integration/data/bug-2074.php
+++ b/tests/Integration/data/bug-2074.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Bug2074;
+
+use Illuminate\Database\Eloquent\Builder;
+
+use function PHPStan\Testing\assertType;
+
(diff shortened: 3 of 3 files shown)


szepeviktor (approved):

src/Types/ModelProperty/GenericModelPropertyType.php
```diff
@@ -65,7 +65,7 @@ public function acceptsWithReason(Type $type, bool $strictTypes): AcceptsResult
             }
 
             if (count($genericType->getObjectClassNames()) < 1) {
-                return AcceptsResult::createNo();
+                return AcceptsResult::createYes();
```
canvural, line 68: Can we do a `maybe` here? Are the any negative impacts?

calebdw: `maybe` results in the same error

canvural:
Thank you!

PR #2102, opened 2024-11-11
Use PHPStan 2.0

szepeviktor:
👉🏻 3.x

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "illuminate/pipeline": "^11.15.0",
         "illuminate/support": "^11.15.0",
         "phpmyadmin/sql-parser": "^5.9.0",
-        "phpstan/phpstan": "2.0.x-dev as 1.12.5"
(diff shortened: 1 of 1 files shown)


canvural:
Oh wow, this is dependency hell 😢 Can we workaround somehow? 🤔

szepeviktor:
This may be enough: https://github.com/rectorphp/rector/issues/8815

canvural:
> This may be enough: [rectorphp/rector#8815](https://github.com/rectorphp/rector/issues/8815)

I don't think they will do it soon. It's a huge change. Also they'll need to release Rector 2.0, then `spatie/laravel-ray` needs to update their code and then `workbench` needs to update their code.. So we need another solution for now. I'll think about it.

canvural:
We can add
```json
"replace": {
  "rector/rector": "*"
},
```

just before installing the dependencies 🤷🏽 Very ugly hack

canvural:
Fixed in the branch.

PR #2105, opened 2024-11-11
Uses `orchestra/testbench-core` instead of `orchestra/testbench`

crynobone:
- ~Added or updated tests~
- ~Documented user facing changes~

**Changes**

For "Development" packages it is best to require `orchestra/testbench-core` due to the minimum amount of dependencies introduced.

**Breaking changes**

No breaking changes since it only changes `require-dev`


canvural:
Thank you.

PR #2107, opened 2024-11-12
[2.x] Dynamically determine return type for `Collection::reject`

spawnia:
- [x] Added or updated tests

**Changes**

Adds support for dynamically determining the return type for `Collection::reject`, analogous to `Collection::filter`.

**Breaking changes**

None.


src/ReturnTypes/CollectionFilterRejectDynamicReturnTypeExtension.php
```diff
@@ -15,13 +15,15 @@
...
+use PHPStan\Type\StaticTypeFactory;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 
 use function count;
+use function in_array;
 use function is_string;
 
-class CollectionFilterDynamicReturnTypeExtension implements DynamicMethodReturnTypeExtension
+class CollectionFilterRejectDynamicReturnTypeExtension implements DynamicMethodReturnTypeExtension
```
spawnia, line 26: I have considered making the class abstract instead and introducing a bunch of abstract methods that handle the different code paths, but found that using `match` is a simpler solution here. It keeps the amount of boilerplate to a minimum and keeps the code readable.

src/ReturnTypes/CollectionFilterRejectDynamicReturnTypeExtension.php
```diff
@@ -51,10 +53,16 @@ public function getTypeFromMethodCall(
             return null;
         }
 
+        $methodName = $methodReflection->getName();
+        assert($methodName === 'filter' || $methodName === 'reject', 'proven in isMethodSupported');
```
spawnia, line 57: Otherwise PHPStan complains about unhandled cases in `match`.

src/ReturnTypes/CollectionFilterRejectDynamicReturnTypeExtension.php
```diff
@@ -51,10 +53,16 @@ public function getTypeFromMethodCall(
...
+
         if (count($methodCall->getArgs()) < 1) {
-            $nonFalseyTypes = TypeCombinator::removeFalsey($valueType);
+            $modifiedType = match ($methodName) {
+                'filter' => TypeCombinator::removeFalsey($valueType),
+                'reject' => TypeCombinator::intersect($valueType, StaticTypeFactory::falsey())
```
spawnia: Fails with the following error:

```
 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   ReturnTypes/CollectionFilterRejectDynamicReturnTypeExtension.php
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------------
  62     Calling PHPStan\Type\StaticTypeFactory::falsey() is not covered by backward compatibility promise. The method might change in a minor PHPStan version.
         💡 If you think it should be covered by backward compatibility promise, open a discussion:
         💡 https://github.com/phpstan/phpstan/discussions
         💡
         💡 See also:
         💡 https://phpstan.org/developing-extensions/backward-compatibility-promise

 ------ --------------------------------------------------------------------------------------------------------------------------------------------------------
```

How should we handle this? I found no other way to do this, `TypeCombinator` does not offer something like `removeTruthy`.

tests/Type/CollectionDynamicReturnTypeExtensionTest.php
```diff
@@ -6,21 +6,14 @@
...
     {
         yield from self::gatherAssertTypes(__DIR__ . '/data/collection-filter.php');
+        yield from self::gatherAssertTypes(__DIR__ . '/data/collection-reject.php');
         yield from self::gatherAssertTypes(__DIR__ . '/data/collection-where-not-null.php');
-
-        if (PHP_VERSION_ID < 70400) {
-            return;
-        }
-
-        yield from self::gatherAssertTypes(__DIR__ . '/data/collection-filter-arrow-function.php');
```
spawnia, line 23: No longer necessary to be kept separate because the project now requires PHP 8.

tests/Type/data/collection-reject.php
spawnia, line 1: Essentially a copy of `collection-filter.php`. I found it easy to compare the two by diffing them, rather than having a single file that handles both methods. Let me know if we I should combine them instead, or perhaps add a comment that hints to keep them in sync.

spawnia:
I just installed this in a large project (over 5000 files), it works great and fixes the specific issue with `reject` we were having.

src/ReturnTypes/CollectionFilterRejectDynamicReturnTypeExtension.php
canvural: Can you open a discussion in the PHPStan repo, like the error message says? I think it is a use case and we can add `@api` annotation to the method. Or like you said maybe we can introduce `removeTruthy` method.

src/ReturnTypes/CollectionFilterRejectDynamicReturnTypeExtension.php
```diff
@@ -30,7 +33,7 @@ public function getClass(): string
 
     public function isMethodSupported(MethodReflection $methodReflection): bool
     {
-        return $methodReflection->getName() === 'filter';
+        return in_array($methodReflection->getName(), ['filter', 'reject']);
```
canvural:
```suggestion
        return in_array($methodReflection->getName(), ['filter', 'reject'], true);
```

Let's be more strict.

src/ReturnTypes/CollectionFilterRejectDynamicReturnTypeExtension.php
spawnia: I opened https://github.com/phpstan/phpstan/discussions/12039. Are you fine with ignoring the error for now? Since tests run on both the highest and lowest allowed dependencies, we should notice when the API breaks.

canvural: We can now use `TypeCombinator::removeTruthy` We also need to update PHPStan version in composer.json to latest one

spawnia: Thanks for helping to land this feature in PHPStan @canvural, I updated to use it.

spawnia:
~Regarding https://github.com/larastan/larastan/actions/runs/11888502482/job/33123165616?pr=2107 - can you just squash and merge, or do you need me to squash my commits?~ Never mind, just squashed.

canvural:
Thank you!

PR #2118, opened 2024-11-18
Fixed the type of the make method of the Attribute class

naotake51:
- [ ] Added or updated tests

**Changes**

The argument type of the make method of the Attribute class has been modified to match the Laravel documentation.

https://laravel.com/docs/11.x/eloquent-mutators#building-value-objects-from-multiple-attributes

I wrote the following code according to the Laravel documentation, but when I ran it on Larastan (Level 9), I noticed the following error, which led me to create a PR:
If there doesn't seem to be any problems, I'd like to ask you to merge it.

```php
    /**
     * @return Attribute<Period, Period>
     */
    protected function publishPeriod(): Attribute
    {
        return Attribute::make(
            get: fn(mixed $value, array $attributes) => new Period(
                $attributes['publish_from'],
                $attributes['publish_to'],
            ),
            set: fn(Period $value) => [
                'publish_from' => $value->from,
                'publish_to' => $value->to,
            ],
        );
    }
```

```
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Models/Recruitment.php                                                                                                                                                            
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  24     Parameter $get of static method Illuminate\Database\Eloquent\Casts\Attribute<mixed,mixed>::make() expects (callable(mixed, mixed): App\ValueObjects\Period)|null, Closure(mixed,  
         array): App\ValueObjects\Period given.                                                                                                                                            
         🪪  argument.type                                                                                                                                                                 
         💡 Type #1 from the union: Type array of parameter #2 $attributes of passed callable needs to be same or wider than parameter type mixed of accepting callable.                   
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
```

**Breaking changes**

### DIFF
--- a/stubs/common/Attribute.stub
+++ b/stubs/common/Attribute.stub
@@ -11,14 +11,14 @@ class Attribute
     /**
      * The attribute accessor.
      *
-     * @var callable(): TGet
+     * @var callable(mixed, array<string, mixed>): TGet
      */
     public $get;
 
     /**
      * The attribute mutator.
      *
-     * @var callable(TSet): mixed
+     * @var callable(TSet, array<string, mixed>): mixed
      */
     public $set;
 
@@ -27,8 +27,8 @@ class Attribute
      *
      * @template TMakeGet
      * @template TMakeSet
-     * @param  (callable(mixed, mixed): TMakeGet)|null  $get
-     * @param  (callable(TMakeSet, mixed=): mixed)|null  $set
+     * @param  (callable(mixed, array<string, mixed>): TMakeGet)|null  $get
+     * @param  (callable(TMakeSet, array<string, mixed>): mixed)|null  $set
      * @return Attribute<TMakeGet, TMakeSet>
      */
     public static function make(callable $get = null, callable $set = null);
@@ -37,7 +37,7 @@ class Attribute
      * Create a new attribute accessor.
      *
      * @template T
-     * @param  callable(mixed, mixed=): T  $get
+     * @param  callable(mixed, array<string, mixed>): T  $get
      * @return Attribute<T, never>
      */
     public static function get(callable $get);
@@ -46,7 +46,7 @@ class Attribute
      * Create a new attribute mutator.
      *
      * @template T
-     * @param  callable(T, mixed=): mixed $set
+     * @param  callable(T, array<string, mixed>): mixed $set
      * @return Attribute<never, T>
      */
     public static function set(callable $set);
--- a/tests/Integration/IntegrationTest.php
+++ b/tests/Integration/IntegrationTest.php
@@ -89,6 +89,14 @@ class IntegrationTest extends PHPStanTestCase
                 34 => ['Parameter #1 $attributes of static method Illuminate\Database\Eloquent\Builder<ModelPropertyStaticCall\ModelPropertyStaticCallsInClass>::create() expects array<model property of ModelPropertyStaticCall\ModelPropertyStaticCallsInClass, mixed>, array<string, string> given.'],
             ],
         ];
+
+        yield [
+            __DIR__ . '/data/model-property-mutator-and-casting.php',
+            [
+                24 => ['Parameter #1 $lineOne of class ModelPropertyMutatorAndCasting\Address constructor expects string, mixed given.'],
+                25 => ['Parameter #2 $lineTwo of class ModelPropertyMutatorAndCasting\Address constructor expects string, mixed given.'],
+            ],
+        ];
     }
 
     /**
--- a/tests/Integration/data/model-property-mutator-and-casting.php
+++ b/tests/Integration/data/model-property-mutator-and-casting.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace ModelPropertyMutatorAndCasting;
+
+use Illuminate\Database\Eloquent\Casts\Attribute;
+
+class Address
+{
+    public function __construct(
+        public string $lineOne,
+        public string $lineTwo,
+    ) {}
+}
+
+class ModelPropertyMutatorAndCastingInClass extends \Illuminate\Database\Eloquent\Model
+{
+    /**
+     * @return Attribute<Address,Address>
+     */
+    protected function address(): Attribute
+    {
+        return Attribute::make(
+            get: fn(mixed $value, array $attributes) => new Address(
+                $attributes['address_line_one'],
+                $attributes['address_line_two'],
+            ),
+            set: fn(Address $value) => [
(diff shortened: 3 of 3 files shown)


stubs/common/Attribute.stub
```diff
@@ -27,7 +27,7 @@ class Attribute
      *
      * @template TMakeGet
      * @template TMakeSet
-     * @param  (callable(mixed, mixed): TMakeGet)|null  $get
+     * @param  (callable(mixed, array): TMakeGet)|null  $get
```
calebdw: this should be:

```suggestion
     * @param  (callable(mixed, array<string, mixed>=): TMakeGet)|null  $get
```

stubs/common/Attribute.stub
```diff
@@ -27,7 +27,7 @@ class Attribute
...
      * @param  (callable(TMakeSet, mixed=): mixed)|null  $set
```
calebdw: You should also adjust the other types (including the `get` and `set` methods below):

```suggestion
     * @param  (callable(TMakeSet, array<string, mixed>=): mixed)|null  $set
```

calebdw (changes requested):

calebdw (changes requested):
The other methods can also be updated:

image

as can these `@vars`:

image

stubs/common/Attribute.stub
naotake51: This may just be due to a lack of knowledge on my part, but if you specify the array type in detail, problems may arise when you actually use it.
I don't know how to get around this easily.

```php
    /**
     * @return Attribute<Period, Period>
     */
    protected function publishPeriod(): Attribute
    {
        return Attribute::make(
            get: fn(mixed $value, array $attributes) => new Period(
                $attributes['publish_from'],
                $attributes['publish_to'],
            ),
            set: fn(Period $value) => [
                'publish_from' => $value->from,
                'publish_to' => $value->to,
            ],
        );
```
```
 ------ -------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Models/Recruitment.php                                                                                                                      
 ------ -------------------------------------------------------------------------------------------------------------------------------------------- 
  24     Parameter $get of static method Illuminate\Database\Eloquent\Casts\Attribute<mixed,mixed>::make() expects (callable(mixed, array<string,    
         mixed>=): App\ValueObjects\Period)|null, Closure(mixed, array): App\ValueObjects\Period given.                                              
         🪪  argument.type                                                                                                                           
         💡 Type #1 from the union: Parameter #2 $attributes of passed callable is required but the parameter of accepting callable is optional. It  
            might be called without it.                                                                                                              
  25     Parameter #1 $from of class App\ValueObjects\Period constructor expects string, mixed given.                                                
         🪪  argument.type                                                                                                                           
  26     Parameter #2 $to of class App\ValueObjects\Period constructor expects string, mixed given.                                                  
         🪪  argument.type                                                                                                                           
 ------ -------------------------------------------------------------------------------------------------------------------------------------------- 
````

Comment (calebdw) [reply stubs/common/Attribute.stub]: Ah, this is due to the `=` on the end of the array---this is to designate that the calling code might not pass it not not that the called code might not use it. You can just remove the `=` so that you get `array<string, mixed>`

Comment (naotake51) [reply stubs/common/Attribute.stub]: Looking at the get and set calls, it appears that the second argument is always specified.
I'll remove the =. Thank you.

get: https://github.com/laravel/framework/blob/11.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L706-L708

set:
https://github.com/laravel/framework/blob/11.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1131
https://github.com/laravel/framework/blob/11.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1854

**Review by calebdw — APPROVED** [review]
Looks good, thanks!

File: stubs/common/Attribute.stub (hunk 2 above)
Comment (naotake51) [reply stubs/common/Attribute.stub]: I still haven't figured out a way to avoid this.

I'm very sleepy, so I'll get back to it another day. (I have plans tomorrow, so it will be after the day after tomorrow).

```
 25     Parameter #1 $from of class App\ValueObjects\Period constructor expects string, mixed given.                                                
         🪪  argument.type                                                                                                                           
  26     Parameter #2 $to of class App\ValueObjects\Period constructor expects string, mixed given.                                                  
         🪪  argument.type  
```

Comment (calebdw) [reply stubs/common/Attribute.stub]: That's because PHPStan knows nothing about the array shape and contents. You can cast to a string for now


```php
            get: fn(mixed $value, array $attributes) => new Period(
                (string) $attributes['publish_from'],
                (string) $attributes['publish_to'],
            ),
```

naotake51: Yeah, I tried using Cast, but I still got an error for that.😭

```
 ------ ------------------------------ 
  Line   Models/Recruitment.php        
 ------ ------------------------------ 
  26     Cannot cast mixed to string.  
         🪪  cast.string               
  27     Cannot cast mixed to string.  
         🪪  cast.string               
 ------ ------------------------------ 
 ```

Comment (naotake51) [reply stubs/common/Attribute.stub]: I wish I could do 'as ~' like in TypeScript...

```
            get: fn(mixed $value, array $attributes) => new Period(
                $attributes['publish_from'] as string,
                $attributes['publish_to'] as string,
            ),
```

Comment (calebdw) [reply stubs/common/Attribute.stub]: that wouldn't change anything and ain't how casting works in php

Comment (naotake51) [reply stubs/common/Attribute.stub]: I can't think of a good way to do this, so I'm going to ignore it in my project.
```
                $attributes['publish_from'], // @phpstan-ignore-line
                $attributes['publish_to'], // @phpstan-ignore-line
```

Thank you for your cooperation.

calebdw: or just add a global `Cannot cast mixed` phpstan ignore

canvural:
Hi,

Can you add some integration test for this? The code snippet you shared should be fine.

naotake51:
@canvural 

> Hi,
Can you add some integration test for this? The code snippet you shared should be fine.

I added tests in 34bfdc47319401e2fd704cf0d992f0c43c5a62cc. Please let me know if it doesn't match your intention.

The code I referred to for the tests: https://laravel.com/docs/11.x/eloquent-mutators#mutating-multiple-attributes

canvural:
Can you remove the merge commit and rebase instead?

canvural:
Thank you!

PR #2121, opened 2024-11-18
Query chunk() custom collection class

### LINKED ISSUE #2079: EloquentBuilder chunk() + custom eloquent collection
- Larastan Version: v2.9.9
- Laravel Version: v11.23.5

### Description

I have a custom eloquent/model collection `ModelCollection` for all my models. `newCollection()` and genericness etc all work. Larastan knows which collection to load for my models.

Except in the `chunk()` Closure in the eloquent builder. Then it doesn't accept my `ModelCollection`, but only the base `EloquentCollection`.

This breaks:

```php
Site::select(['id'])->chunk(100, function(ModelCollection $sites) {
```

with this beauty:

    Parameter #2 $callback of method Illuminate\Database\Eloquent\Builder<App\Models\Site>::chunk() expects callable(Illuminate\Database\Eloquent\Collection<int, App\Models\Site>, int): mixed, Closure(App\Base\ModelCollection): void given.

Larastan does not complain about `EloquentCollection`:

```php
Site::select(['id'])->chunk(100, function(EloquentCollection $sites) {
	\PHPStan\dumpType($sites); // Illuminate\Database\Eloquent\Collection<int, App\Models\Site>
```

but that's not quite good enough, because `$sites` really is a `ModelCollection`!

----

I assumed a cool Phpstan extension to make `chunk()` even work at all, but that's all phpdoc 🤩 Buuut that's probably not good enough anymore.

rudiedirkx:
- [ ] Added or updated tests
- <s>Documented user facing changes</s>

Fixes #2079 

**Changes**

Phpdoc didn't support custom model collection, only `EloquentCollection`:

```
@param  callable(\Illuminate\Database\Eloquent\Collection<int, TModelClass>, int): mixed  $callback
```

This new extension (first of `MethodParameterClosureTypeExtension` 😮) uses `CollectionHelper` to find the specific collection.

**Breaking changes**

None.


rudiedirkx:
`chunk()` unit tests fail with

```diff
--- Expected
+++ Actual
@@ @@
-'Illuminate\Database\Eloquent\Collection<int, App\User>'
+'mixed'
```

maybe because the detailed phpdoc is gone, but the extension isn't used? Why isn't the extension used?

src/Methods/CollectionChunkClosureTypeExtension.php
```diff
@@ -0,0 +1,84 @@
...
+
+final class CollectionChunkClosureTypeExtension implements MethodParameterClosureTypeExtension
+{
+    public function __construct(private CollectionHelper $collectionHelper)
+    {
+    }
+
+    public function isMethodSupported(MethodReflection $methodReflection, ParameterReflection $parameter): bool
+    {
+        if (!in_array($methodReflection->getName(), ['chunk', 'chunkById', 'chunkByIdDesc'])) {
```
calebdw: prefer strict mode:

```suggestion
        if (!in_array($methodReflection->getName(), ['chunk', 'chunkById', 'chunkByIdDesc'], strict: true)) {
```

src/Methods/CollectionChunkClosureTypeExtension.php
```diff
@@ -0,0 +1,84 @@
...
+            return false;
+        }
+
+        if ($parameter->getName() != 'callback') {
+            return false;
+        }
+
+        return (new ObjectType(EloquentBuilder::class))->isSuperTypeOf(new ObjectType($methodReflection->getDeclaringClass()->getName()))->yes();
```
calebdw: Use this instead:

```suggestion
        return $methodReflection->getDeclaringClass()->is(EloquentBuilder::class);
```

stubs/11.0.0/EloquentBuilder.stub
calebdw, line 1: stubs should not be changed

calebdw (changes requested):
This should target 3.x as that's the newest release. Laravel 10 has reached EOL and Larastan 2.x should be sunsetting

rudiedirkx:
> stubs should not be changed

It won't work then. Apparently an extension can't undo or overwrite the phpdoc `callback` details.

And some more failing tests that I don't understand. I'm not willing to put more hours in for probably very simple issues that I don't understand. I'll take a look after the professionals do their thing.

calebdw:
> Apparently an extension can't undo or overwrite the phpdoc callback details.

You're mistaken, extensions override phpdocs all the time

rudiedirkx:
I couldn't get it to work. Only like this. I tried all combinations of Laravel's phpdoc + Larastan's phpdoc.

rudiedirkx:
@calebdw It really doesn't work without messing up the phpdoc. (Phpstan 1.x still.) The extension works perfectly, and it returns the perfect `ClosureType`, but Phpstan doesn't use it. Could you look at it and tell me what I'm doing wrong/how Phpstan works?

canvural:
Hi,

Thank you for the PR. But currently there are some issues with the `*ParameterClosureTypeExtension` extensions. I already had changes prepared to use the new extensions, but the bugs are keeping me from releasing them. 

I'll close this for now, and when the bugs related to extensions are fixed I'll implement them here.

PR #2122, opened 2024-11-20
fix: parameter type for HasOneOrMany::createMany

sebdesign:
- [ ] Added or updated tests

**Changes**

The `$records` parameter type for the `HasOneOrMany::createMany()` method was wrong.
It should be an iterable of arrays with model properties as keys and mixed values.

**Breaking changes**

### DIFF
--- a/stubs/common/HasOneOrMany.stub
+++ b/stubs/common/HasOneOrMany.stub
@@ -71,7 +71,7 @@ abstract class HasOneOrMany extends Relation
     /**
      * Create a Collection of new instances of the related model.
      *
-     * @param  iterable<model-property<TRelatedModel>, mixed>  $records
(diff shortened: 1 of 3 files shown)


zepfietje:
I'm also running into this issue with the latest version of Larastan.
Seems to have been introduced in https://github.com/larastan/larastan/commit/61d236a1062113a9150c2786e13d534046d22530.

CC @calebdw

calebdw:
@zepfietje, 

This method does not exist in my stubs: https://github.com/calebdw/larastan/blob/2.x/stubs/11.0.0/HasOneOrMany.stub

This must've been added by @canvural when he was cherry-picking my commits from #1990

zepfietje:
Ah I was already afraid of that since I couldn't find the reference to your PR from the commit. Sorry for notifying you :)

canvural:
I added some tests. Thank you!

PR #2131, opened 2024-11-27
Improve parameter of QueryBuilder::get

markwalet:
Allow `Expression` in querybuilder `->get()` method.

- [ ] Added or updated tests

**Changes**

### DIFF
--- a/stubs/common/QueryBuilder.stub
+++ b/stubs/common/QueryBuilder.stub
@@ -178,7 +178,7 @@ class Builder
     /**
      * Execute the query as a "select" statement.
      *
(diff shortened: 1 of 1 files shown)


canvural:
Thank you! Could you send the same one to 2.x? If not I'll do it later.

markwalet:
@canvural Is it possible to get this in a patch release perhaps?

Not a big deal, but my Renovatebot is complaining about QA when automatically upgrading to 3.0.2 :joy:. I could also disable the rule temporarily ofcourse but just applying a newer (fixed) version would be better

markwalet:
@canvural Do you think we'll see a new release soon?

Thanks for the great work! 🚀

PR #2132, opened 2024-11-27
[2.x] Improve parameter of QueryBuilder::get

markwalet:
Allow `Expression` in querybuilder `->get()` method.

Asked to put a PR in for V2 as well in https://github.com/larastan/larastan/pull/2131

### DIFF
--- a/stubs/10.0.0/QueryBuilder.stub
+++ b/stubs/10.0.0/QueryBuilder.stub
@@ -346,7 +346,7 @@ class Builder
     /**
      * Execute the query as a "select" statement.
      *
-     * @param  string[]|string  $columns
+     * @param  array<string|\Illuminate\Contracts\Database\Query\Expression>|string  $columns
      * @return \Illuminate\Support\Collection<int, \stdClass>
      */
     public function get($columns = ['*']);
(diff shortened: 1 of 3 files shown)


szepeviktor:
```
Parameter $columns of method Illuminate\Database\Query\Builder::get() has
 invalid type Illuminate\Contracts\Database\Query\Expression.
```
Stubs need all used classes.

markwalet:
@szepeviktor I'm not entirely sure what you mean. I was just asked to apply the same changes from the 3.x branch to the 2.x branch as well. So I cherry-picked the commits.

szepeviktor:
`Expression` is not part of the stubs in this branch.

markwalet:
Ah, it looks like it is only a problem in Laravel 9 because the Expression class didn't exist back then.

@canvural So that means we can close this one because it is not compatible with all supported Laravel versions?

calebdw:
You can apply the stub and the method fix under the 10.x directory
https://github.com/larastan/larastan/blob/2.x/stubs/10.0.0/QueryBuilder.stub

szepeviktor:
and maybe 11 too

markwalet:
@calebdw Ahh I see, no clue how I missed that option completely 😂 

Should be fixed now. Thanks for the feedback!

szepeviktor:
Thank you.

PR #2133, opened 2024-11-28
Collection::groupBy() return type

lorenzolosa:
- [ ] Added or updated tests

**Changes**

Updates the return type definition of `Collection::groupBy()` in line with https://github.com/laravel/framework/pull/53684.

**Breaking changes**

None.


calebdw:
Or just remove this method since it's no longer needed---given that the framework has the proper docs

canvural:
Thank you!

PR #2143, opened 2024-12-04
[3.x] fix: macros on Cache facade

### LINKED ISSUE #2141: Macro for \Cache - Call to an undefined static method
- Larastan Version: 2.9.12
- Laravel Version: 11.31.0

### Description
I get a phpstan error when I try to extend the Cache facade
```
index.php:10:Call to an undefined static method Illuminate\Support\Facades\Cache::rememberIf(). [identifier=staticMethod.notFound]
```

```php
<?php

declare(strict_types=1);

namespace App\Providers;

use Illuminate\Support\ServiceProvider;

class AppServiceProvider extends ServiceProvider
{
    public function boot(): void
    {
        \Cache::macro(
            'rememberIf',
            function (bool $condition, string $key, \DateTimeInterface|\DateInterval|int $ttl, \Closure $callback) {
                if ($condition) {
                    return \Cache::remember($key, $ttl, $callback);
                } else {
                    return $callback();
                }
            }
        );
    }
}
```

```php
<?php

$cache = \Cache::rememberIf($condition), 'test', 999, function () {
    return rand();
});
```

calebdw:
**Changes**

Hello!

This fixes Cache macros and closes #2141.

Thanks!

**Breaking changes**

None

### DIFF
--- a/src/Methods/MacroMethodsClassReflectionExtension.php
+++ b/src/Methods/MacroMethodsClassReflectionExtension.php
@@ -6,9 +6,13 @@ namespace Larastan\Larastan\Methods;
 
 use Closure;
 use Illuminate\Auth\RequestGuard;
+use Illuminate\Auth\SessionGuard;
+use Illuminate\Cache\CacheManager;
+use Illuminate\Cache\Repository as CacheRepository;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Query\Builder as QueryBuilder;
 use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\Facade;
 use Illuminate\Support\Str;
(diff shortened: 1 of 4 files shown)


src/Methods/MacroMethodsClassReflectionExtension.php
```diff
@@ -111,7 +118,6 @@ public function hasMethod(ClassReflection $classReflection, string $methodName):
                 }
 
                 $refProperty = $macroClassReflection->getNativeReflection()->getProperty($macroTraitProperty);
-                $refProperty->setAccessible(true);
```
calebdw, line 114: From the [docs](https://www.php.net/manual/en/reflectionproperty.setaccessible.php):

> Note: As of PHP 8.1.0, calling this method has no effect; all properties are accessible by default.

canvural:
Thank you!

PR #2157, opened 2024-12-27
Add Fallback for Types That Can't Be Resolved Dynamically

AJenbo:
**Changes**
When resolving types for calls like `resolve(MyClass::class)`, Larastan may fail to resolve an object dynamically in certain environments. In such cases, the resolved type is marked as `*ERROR*`, leading to a fallback to `mixed` without explicit warnings, even if the type could be derived from the input string.

This patch introduces a fallback mechanism that uses the provided string (e.g., `MyClass::class`) as the return type if dynamic resolution fails. This ensures that type information is preserved whenever possible, while maintaining compatibility with existing behavior.

**Breaking changes**
Cases where the `*ERROR*` type was previously returned (causing subsequent code to operate on an implicit `mixed`) will now see the object type. This may result in the detection of errors that were previously missed.

Anyone using PHPStan at level 10 would already have addressed such cases in one way or another, so they should not see any new issues. However, they may now be able to remove some workarounds.

### DIFF
--- a/src/ReturnTypes/AppMakeDynamicReturnTypeExtension.php
+++ b/src/ReturnTypes/AppMakeDynamicReturnTypeExtension.php
@@ -28,7 +28,7 @@ final class AppMakeDynamicReturnTypeExtension implements DynamicStaticMethodRetu
         return $methodReflection->getName() === 'make' || $methodReflection->getName() === 'makeWith';
     }
 
-    public function getTypeFromStaticMethodCall(MethodReflection $methodReflection, StaticCall $methodCall, Scope $scope): Type
+    public function getTypeFromStaticMethodCall(MethodReflection $methodReflection, StaticCall $methodCall, Scope $scope): Type|null
     {
         return $this->appMakeHelper->resolveTypeFromCall($methodCall, $scope);
     }
--- a/src/ReturnTypes/AppMakeHelper.php
+++ b/src/ReturnTypes/AppMakeHelper.php
@@ -10,7 +10,6 @@ use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\StaticCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Type\ErrorType;
-use PHPStan\Type\MixedType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
@@ -58,6 +57,6 @@ final class AppMakeHelper
             return TypeCombinator::union(...$types);
         }
 
-        return new MixedType();
+        return null;
     }
 }
--- a/src/ReturnTypes/AppMakeHelper.php
+++ b/src/ReturnTypes/AppMakeHelper.php
@@ -16,13 +16,15 @@ use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use Throwable;
 
+use function class_exists;
 use function count;
+use function interface_exists;
 
 final class AppMakeHelper
 {
     use HasContainer;
 
-    public function resolveTypeFromCall(FuncCall|MethodCall|StaticCall $call, Scope $scope): Type
+    public function resolveTypeFromCall(FuncCall|MethodCall|StaticCall $call, Scope $scope): Type|null
     {
         $args = $call->getArgs();
         if (count($args) === 0) {
@@ -37,14 +39,17 @@ final class AppMakeHelper
             $types = [];
             foreach ($constantStrings as $constantString) {
                 try {
+                    $class = $constantString->getValue();
                     /** @var object|null $resolved */
-                    $resolved = $this->resolve($constantString->getValue());
+                    $resolved = $this->resolve($class);
 
-                    if ($resolved === null) {
+                    if ($resolved !== null) {
+                        $class = $resolved::class;
+                    } elseif (! class_exists($class) && ! interface_exists($class)) {
                         return new ErrorType();
                     }
 
-                    $types[] = new ObjectType($resolved::class);
+                    $types[] = new ObjectType($class);
                 } catch (Throwable) {
                     return new ErrorType();
                 }
--- a/src/ReturnTypes/ApplicationMakeDynamicReturnTypeExtension.php
+++ b/src/ReturnTypes/ApplicationMakeDynamicReturnTypeExtension.php
@@ -30,7 +30,7 @@ final class ApplicationMakeDynamicReturnTypeExtension implements DynamicMethodRe
         return in_array($methodReflection->getName(), ['make', 'makeWith', 'resolve'], true);
     }
 
-    public function getTypeFromMethodCall(MethodReflection $methodReflection, MethodCall $methodCall, Scope $scope): Type
+    public function getTypeFromMethodCall(MethodReflection $methodReflection, MethodCall $methodCall, Scope $scope): Type|null
     {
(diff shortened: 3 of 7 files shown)


calebdw:
This is already the default behaviour provided by the framework generics:

https://github.com/laravel/framework/blob/8f6a9b03b8e41eedca5f6182d1af2b81ad4be3e6/src/Illuminate/Foundation/helpers.php#L793-L802

Returning `null` from the extension in the case of a throwable should fallback to the default behaviour given by the method docs

AJenbo:
> This is already the default behaviour provided by the framework generics:
>
> https://github.com/laravel/framework/blob/8f6a9b03b8e41eedca5f6182d1af2b81ad4be3e6/src/Illuminate/Foundation/helpers.php#L793-L802

The test case I have added fails on the current 3.x branch so it must be overwriting the behavior you mention.

> Returning null from the extension in the case of a throwable should fallback to the default behaviour given by the method docs

It's not the catch that is in question here but the return on line 44: https://github.com/larastan/larastan/blob/3.x/src/ReturnTypes/AppMakeHelper.php#L44

Returning `null` results in the type being `mixed` rather then `*ERROR*` or the given class name.

calebdw:
So return `null` instead of the ErrorType or MixedType returns

AJenbo:
> So return `null` instead of the ErrorType or MixedType returns

Like I said I tried that, it results in `mixed` not the given class. If you want to be sure I'm not simply doing it wrong then please copy my test case and try it out locally.

calebdw:
It does work for the `app()` and `resolve()` helper functions: https://github.com/calebdw/larastan/commit/e98650f6a420f3cd7f7b9ec661a9579d9e88a7f7

The reason why your tests are failing are because the `make`, `makeWith`,and  `resolve` **methods** still have the return type of `mixed`---however, they just need to be fixed in the framework:

https://github.com/laravel/framework/blob/8f6a9b03b8e41eedca5f6182d1af2b81ad4be3e6/src/Illuminate/Foundation/Application.php#L1017-L1026

AJenbo:
In our application we use App::make() which I don't think can be hinted in a way that would work with you suggestion.

calebdw:
Ah, that's a good point... however, you're not actually testing the `App::make` facade

AJenbo:
True enough, I was thinking a general test would be enough since it covered the changed code :)

AJenbo:
Test should pass now.

@calebdw do you think it makes sens to also have it return `null` instead of ErrorType/MixedType so that we leave it up to the framework what to do when we are unable to determine the type in the extension?

calebdw:
I would return `null` instead of Mixed, the error type is probably fine if it's a junk value

AJenbo:
Done, also fixed the code style and cleaned up the commits. Should be ready to merge now I hope :)
(the 3 failing tests are also failing on the main 3.x branch)

calebdw (approved):
Looks good to me

AJenbo:
Rebased since @szepeviktor fixed the failing tests :+1:

canvural:
Hi,

I do not really understand the logic here. If we can't resolve a class/interface we should not fallback to it anyway. It means class does not exist or like in your example the constructor throws an exception etc. This would swallow even more potential errors. 

The logic kinda overlaps with the discussion in this [issue](https://github.com/larastan/larastan/issues/2128). And I'd pretty much prefer a new rule that'd warn the user if a class/interface could not be resolved.

AJenbo:
@canvural 

> I do not really understand the logic here.

I wish you had taken the time to let me explain rather than just closing the PR. It’s not that it can’t be resolved—it’s that it can’t be instantiated during static analysis.

The current approach is causing Larastan to fail for several applications because it tries to resolve classes that attempt to call a database or an API that isn’t available in that environment when instantiating the class.

In these cases, application developers have to manually hint the class rather than relying on Larastan to resolve it. What’s worse, Larastan returns an invalid type, causing PHPStan to overlook operations on an unknown class. This results in portions of code not being analyzed, even when running PHPStan at high levels.

Larastan trying to instantiate these also violates the PHPStan promise of not executing code, and can potentially lead to unexpected side effects. I understand that it's sometimes the only viable way for it to understand what type the resolver would return like in cases of `resolve('cache')`, but in my view it would be better that it only do so in cases where the given string is not an existing class/interface (`resolve('CacheManager::class')`), doing so should be less prone to fail, faster, less risk of side effects.

calebdw:
@AJenbo,

> Larastan trying to instantiate these also violates the PHPStan promise of not executing code, and can potentially lead to unexpected side effects.

While I mostly agree with you, Larastan has been up-front about this being "code analysis" rather than purely "static analysis":

image

In order to understand the magic of Laravel it is necessary to execute some code. Generally speaking, it's not wise to put heavy initialization logic (like database queries or api calls) in the constructor of objects. If you absolutely must do this, then I would encourage you to look into the new PHP 8.4 [Lazy Objects](https://www.php.net/manual/en/language.oop5.lazy-objects.php).

Also, the Laravel community has a very bad habit (one of my biggest pet peeves) of typing interfaces, but then proceeded to call methods that only exist on the runtime implementation---this violates the contract but works at runtime so people just go with it. Therefore it is necessary to use the container to resolve the runtime implementation for a contract or you will have a lot of unhappy people

AJenbo:
> In order to understand the magic of Laravel it is necessary to execute some code.

I understand this and even mentioned it in my comment. My appeal is to minimize it by avoiding execution when the type can be derived directly from the input.

Where I most often see this issue arise is with API SDKs that require a password/token in their constructor, or even worse, make external calls. This is often not code that the developer controls, as it comes from a third-party package they use but do not maintain. Having to wrap these classes is just extra work for something Larastan could handle more efficiently.

In fact, if I uninstall Larastan, PHPStan resolves these correctly, but then I lose the context for all the other Laravel-specific magic that Larastan usually provides.

> Also, the Laravel community has a very bad habit (one of my biggest pet peeves) of typing interfaces, but then proceeded to call methods that only exist on the runtime implementation---this violates the contract but works at runtime so people just go with it.

100% agree. Maybe they should be made aware of these mistakes instead of Larastan making potentially incorrect assumptions about return types across all environments. Just today, I encountered a case where the Microsoft auth provider was assumed because it was returned during analysis, but depending on the application configuration, a different implementation could be returned at runtime which did not have the used method.

> Therefore it is necessary to use the container to resolve the runtime implementation for a contract or you will have a lot of unhappy people

I disagree. I'm unhappy with how the current approach encourages developers to make incorrect assumptions when building applications.

For this PR, I kept it compatible with the existing behavior and only used the given string when instation wasn’t possible during analysis. So that we could look at that separately.

calebdw:
> I'm unhappy with how the current approach encourages developers to make incorrect assumptions when building applications.

I would prefer being strict about interfaces, and if you type an interface then you're required to adhere to the contract instead of Larastan resolving the runtime implementation (which could be different in prod as opposed to the test pipeline)

AJenbo:
image

I could argue that the statement is misleading since it also applies when the type could be derived statically, but I suppose that would be splitting hairs.

> I would prefer being strict about interfaces, and if you type an interface then you're required to adhere to the contract instead of Larastan resolving the runtime implementation (which could be different in prod as opposed to the test pipeline)

I think the optimal solution would be to make this a setting, allowing developers to choose a "strict-contract" mode, while keeping the current behavior as the default.

AJenbo:
@calebdw @canvural please see https://github.com/larastan/larastan/issues/2216

PR #2162, opened 2025-01-04
fix typo in ModelAppendsRule example

jamessa:
Found a typo in ModelAppendsRule example.

- [ ] Added or updated tests

**Changes**

Variable name in ModelAppendsRule example document.

**Breaking changes**

None.

**Note**

I just happened to bump into this rule and found the example is not correct. Thank you all for the great effort on build this wonderful tool.


szepeviktor:
What a catch! 🎣

szepeviktor:
Thank you.

PR #2164, opened 2025-01-05
Update param types in QueryBuilder

bskl:
- [x] Added or updated tests

**Changes**

Updates the param type definition of QueryBuilder::union() and QueryBuilder::unionAll() in line with [https://github.com/laravel/framework/pull/51851](https://github.com/laravel/framework/pull/51851)
and from [Builder.php#L2885](https://github.com/laravel/framework/blob/8b93d39c4721a0c170b92bfbf7bfb5090b5407cc/src/Illuminate/Database/Query/Builder.php#L2885), [Builder.php#L2905](https://github.com/laravel/framework/blob/8b93d39c4721a0c170b92bfbf7bfb5090b5407cc/src/Illuminate/Database/Query/Builder.php#L2905)

**Breaking changes**

None

### DIFF
--- a/stubs/common/QueryBuilder.stub
+++ b/stubs/common/QueryBuilder.stub
@@ -143,8 +143,7 @@ class Builder
     /**
      * Add a union statement to the query.
      *
-     * @template TModel of \Illuminate\Database\Eloquent\Model
-     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder<TModel>|\Closure  $query
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder<*>|\Closure  $query
      * @param  bool  $all
      * @return $this
      */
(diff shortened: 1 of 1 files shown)


szepeviktor (approved):

stubs/common/QueryBuilder.stub
```diff
@@ -143,8 +143,7 @@ class Builder
     /**
      * Add a union statement to the query.
      *
-     * @template TModel of \Illuminate\Database\Eloquent\Model
-     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder<TModel>|\Closure  $query
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder<*>|\Closure  $query
```
calebdw: The stubs for both these methods provide no additional information over what's already given by the method itself. There's no need to duplicate the information---they can simply be removed

calebdw (changes requested):

stubs/common/QueryBuilder.stub
bskl: Thanks, I think so too. If no need i can removed.

calebdw: Yes they can be removed as there's no need for them with your changes :)

calebdw (approved):
Looks good, thanks!

canvural:
Thank you!

PR #2166, opened 2025-01-06
Fix this mess

szepeviktor:
```
Call to undefined function Illuminate\Foundation\Bootstrap\laravel_cloud()
```

Missing in our bootstrap process ...
https://github.com/laravel/framework/blob/8b93d39c4721a0c170b92bfbf7bfb5090b5407cc/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php#L59

defined in `src/Illuminate/Support/helpers.php`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,7 +25,7 @@ jobs:
       fail-fast: false
       matrix:
         php: ["8.2", "8.3", "8.4"]
-        laravel: ["^11.16.0"]
+        laravel: ["^11.34.0"]
         dependencies:
             - "lowest"
             - "highest"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,7 +25,7 @@ jobs:
       fail-fast: false
       matrix:
         php: ["8.2", "8.3", "8.4"]
-        laravel: ["^11.15.0"]
+        laravel: ["^11.16.0"]
         dependencies:
             - "lowest"
             - "highest"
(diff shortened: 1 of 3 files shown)


szepeviktor:
💡 In the `test` workflow we mix framework version with sample app version!

calebdw:
The composer json should probably be updated

szepeviktor:
> The composer json should probably be updated

Which line? Could you help me?

calebdw:
You're trying to fix the `laravel_cloud` right? So maybe just increase the minimum version to when that was released:

- https://github.com/larastan/larastan/blob/83278e567c244f83fbe0c879b8f91b65e0646284/composer.json#L20
- https://github.com/larastan/larastan/blob/83278e567c244f83fbe0c879b8f91b65e0646284/composer.json#L33

szepeviktor:
Thank you.
In the separate directory for the Laravel sample app v11.37 is installed.
https://github.com/larastan/larastan/actions/runs/12624947660/job/35175908140?pr=2166#step:10:12

I think PHPStan autoloads the framework from the **other directory**.
So increasing that version would hide the problem.

calebdw:
> I think PHPStan autoloads the framework from the other directory.

Ahhh, part of the problem is that the test file is `cd`ing  **outside** the source code directory---in general, everything should be self contained (create subdirectories instead of going outside). There was a PR to do this: https://github.com/larastan/larastan/pull/1856 (which also made it possible to test locally so you don't have to push every single change to see it if it works)

szepeviktor:
Thank you.

PR #2179, opened 2025-01-24
Update EnumeratesValues.stub

piurafunk:
Teach EnumeratesValues::mapSpread() to expand an array of TValue and TKey into the expected callback signature.

- [ ] Added or updated tests

I attempted to see this fixed via PHPStan directly: https://phpstan.org/r/8ca0712e-3193-4ada-8069-078cda880038

However, PHPStan on its own complains about a syntax error. I believe something Larastan does handles the portion that PHPStan views as a docblock syntax error. I don't know where that is, I'm not familiar with the internals of Larastan.

**Changes**

In `\Illuminate\Support\Traits\EnumeratesValues::mapSpread()`, the `$callback` parameter is document as `@param  callable(mixed...): TMapSpreadValue  $callback`. This prevents proper identification of callbacks that match the spread of the collection. For example:

```php
/** @var Collection<int, array{int, int}> $c */
$c = new Collection();
$c->mapSpread(function (int $i, int $j): int {
    return $i + $j;
});
```

This should work, but in my app it fails, complaining that:

```
Parameter #1 $callback of method Illuminate\Support\Collection<int,array<int, int>>::mapSpread() expects callable(mixed ...): int, Closure(int, int): int given.
```

Something in Larastan allows for the `...TValue` syntax I propose here, allowing `mapSpread()` to correctly identify the types of the argument spread.

**Breaking changes**

I don't know of any breaking changes. My best guess would be that any baselines that have these errors notated could report `Ignored error pattern ... was not matched in reported errors.`, depending on the value of `reportUnmatchedIgnoredErrors`.

### DIFF
--- a/stubs/common/EnumeratesValues.stub
+++ b/stubs/common/EnumeratesValues.stub
@@ -48,4 +48,14 @@ trait EnumeratesValues
      * @return static<TFlatMapKey, TFlatMapValue>
      */
     public function flatMap(callable $callback);
+
+	/**
+     * Run a map over each nested chunk of items.
+     *
+     * @template TMapSpreadValue
+     *
+     * @param  callable(...TValue, ?TKey): TMapSpreadValue  $callback
+     * @return static<TKey, TMapSpreadValue>
+     */
+    public function mapSpread(callable $callback);
 }


calebdw:
A phpdoc fix like this should ideally be fixed at the framework so we don't have to maintain extra stubs. They've generally been pretty open to these types of PRs, provided you write an associative type test

piurafunk:
I was exploring that option first. Unfortunately, I found that this syntax isn't supported directly by PHPStan. If they were to make the change, anyone using PHPStan without using Larastan would get errors regarding docblock syntax errors. I very much agree, ideally this would be fixed in `laravel/framework`, but I don't think that's possible without porting some functionality from Larastan to PHPStan (as mentioned in the description, I don't know exactly which part of Larastan makes it work, I only know the playground shows a syntax error, and using Larastan does not).

Thoughts?

calebdw:
> I found that this syntax isn't supported directly by PHPStan. 

If it's not supported by PHPStan then it's not supported by Larastan---you haven't written any tests or you would see this does not work either.

You had some syntax errors in your templates, I updated the code example: https://phpstan.org/r/815cc323-4cc7-4e23-a062-b82e8e90cd33

The problem is that you can't decompose `TValue` in a PHPDoc, you need the inner types of the `TValue`, not `TValue` itself. However, even if you had the inner types I'm not sure it would work because you can't have a `callable(int, int)` given where a `callable(int...)` is expected (you could have more than two values) as shown here: https://phpstan.org/r/f51dd8dc-f14e-4323-bbf7-165a7acda810

This might be solvable with a custom `MethodParameterClosureTypeExtension` but I'm not sure

piurafunk:
I did some local tinkering, to isolate my scenario from my app, and indeed my PR doesn't fix anything. I believe it breaks parsing, thus PHPStan doesn't know what the parameter should be, and so it doesn't completely check it. Even if I gave it a callback that required wrong aruments (`string` vs `int`, for example), and PHPStan didn't complain.

It's not critical to my work now, so I'll close the PR. Perhaps in the future I'll revisit and see if a customer extension can help or something.

PR #2183, opened 2025-01-27
[3.x] fix: package config(`Package/config/config.php`) can use `env`

### LINKED ISSUE #2173: Feature request/Bug report: possibility to specify config/view directory
# Feature Request/Bug Report

- Larastan Version: 3.0.2
- Laravel Version: 11.38.2

### Description

Hey,

not sure if this is a feature request or bug report?! 😄 

Problem: We upgraded to the newest version of larastan and now getting errors regarding views and configs.
As you can see below, in addition to the normal Laravel structure, we also have individual packages with service providers in which config, views, ... are registered. However, these are not recognised when it comes to config or views. Is it possible to tell larastan to look for config and views in these packages as well?

```
  Line   packages/PackageName/src/Controllers/PackageController.php
  22     Parameter #1 $view of function view expects view-string|null, string
         given.
         🪪  argument.type
```

```
  Line   packages/PackageName/config/config.php
  6      Called 'env' outside of the config directory which returns null when
         the config is cached, use 'config'.
         🪪  larastan.noEnvCallsOutsideOfConfig
```

erikn69:
- Resolves #2173

**Changes**

-  Support packages config file 
> ```
>Line   packages/PackageName/config/config.php
> ------ ----------------------------------------------------------------------
>  6      Called 'env' outside of the config directory which returns null when
>         the config is cached, use 'config'.
>         🪪  larastan.noEnvCallsOutsideOfConfig
>```

### DIFF
--- a/src/Rules/NoEnvCallsOutsideOfConfigRule.php
+++ b/src/Rules/NoEnvCallsOutsideOfConfigRule.php
@@ -67,6 +67,6 @@ class NoEnvCallsOutsideOfConfigRule implements Rule
 
     protected function isCalledOutsideOfPackageConfig(FuncCall $call, Scope $scope): bool
     {
-        return basename(dirname($scope->getFile())) !== 'config';
+        return str_starts_with($scope->getFile(), getcwd() . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR) === false;
     }
 }
--- a/src/Rules/NoEnvCallsOutsideOfConfigRule.php
+++ b/src/Rules/NoEnvCallsOutsideOfConfigRule.php
@@ -9,13 +9,13 @@ use PhpParser\Node;
 use PhpParser\Node\Expr\FuncCall;
 use PhpParser\Node\Name;
 use PHPStan\Analyser\Scope;
+use PHPStan\File\FileHelper;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleError;
 use PHPStan\Rules\RuleErrorBuilder;
 
 use const DIRECTORY_SEPARATOR;
 use function config_path;
-use function getcwd;
 use function str_starts_with;
 
 /**
@@ -26,6 +26,10 @@ use function str_starts_with;
 class NoEnvCallsOutsideOfConfigRule implements Rule
 {
     use HasContainer;
+
+    public function __construct(private FileHelper $fileHelper)
+    {
+    }
 
     public function getNodeType(): string
     {
@@ -69,6 +73,6 @@ class NoEnvCallsOutsideOfConfigRule implements Rule
 
     protected function isCalledOutsideOfPackageConfig(Scope $scope): bool
     {
-        return str_starts_with($scope->getFile(), getcwd() . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR) === false;
+        return str_starts_with($scope->getFile(), $this->fileHelper->absolutizePath('config') . DIRECTORY_SEPARATOR) === false;
     }
 }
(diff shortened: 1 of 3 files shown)


calebdw:
This PR would allow me to use `env()` in `app/Models/config` (or any other directory named `config`)...

erikn69:
> This PR would allow me to use env() in app/Models/config (or any other directory named config)...

@calebdw hi, Yes, I noticed, I was trying to obtain the `base_path`, but within the context of Orchestra it is not the correct, any help is welcome, thanks

erikn69:
Seems fixed

src/Rules/NoEnvCallsOutsideOfConfigRule.php
```diff
@@ -60,4 +64,9 @@ protected function isCalledOutsideOfConfig(FuncCall $call, Scope $scope): bool
     {
         return str_starts_with($scope->getFile(), config_path()) === false;
     }
+
+    protected function isCalledOutsideOfPackageConfig(FuncCall $call, Scope $scope): bool
```
szepeviktor: You do not need `$call`, the first parameter.

erikn69: I copied it one line above, I thought it was for a good reason(_maybe override_), since it is not used there either.
I'll delete it then
https://github.com/larastan/larastan/blob/8d433ed5304abdc437439b5b59fa040d6cf43de2/src/Rules/NoEnvCallsOutsideOfConfigRule.php#L59-L62
https://github.com/larastan/larastan/blob/8d433ed5304abdc437439b5b59fa040d6cf43de2/src/Rules/NoEnvCallsOutsideOfConfigRule.php#L46-L48

erikn69: Fixed

calebdw:
This won't work when calling phpstan from another directory like here: https://github.com/larastan/larastan/blob/8d433ed5304abdc437439b5b59fa040d6cf43de2/tests/laravel-test.sh#L32-L34

erikn69:
@calebdw hi, It currently doesn't work under any circumstances, but any suggestions?

erikn69:
@calebdw, @szepeviktor hi, I did add `PHPStan\File\FileHelper`, that should work, what do you think? Also test added

calebdw:
It's probably better to simply accept configuration paths instead of trying to automate discovery

cesarreyes3 (approved):

fredsal (approved):

PaolaRuby (approved):

canvural:
Hi,

Thanks for the PR. But in this state this is not acceptable. I liked the suggestions from the issue to make it configurable. ([here](https://github.com/larastan/larastan/issues/2173#issuecomment-2616269155) and [here](https://github.com/larastan/larastan/issues/2173#issuecomment-2606337406))

So we should add a new `configDirectories` parameter into config. It should default to an empty list. Then in the rule we should check if the list is empty, and if it is we keep using `config_path` to keep the backwards compatibility. And if not we loop over the list and see if the current file path starts with any of them. You can check how `viewDirectories` is implemented for inspiration.

szepeviktor:
- So we should add a new configDirectories parameter into config.
- It should default to an empty list.
- Then in the rule we should check if the list is empty,
- and if it is we keep using config_path to keep the backwards compatibility.
- And if not we loop over the list and see if the current file path starts with any of them.

erikn69:
>You can check how viewDirectories is implemented for inspiration.

@canvural hi, feel free to open the alternative
I prefer it to be automatic, this PR maintains backwards compatibility,
in view of the refusal I have only managed to add the following:
```yml
ignoreErrors:
  -
    identifier: larastan.noEnvCallsOutsideOfConfig
    path: config/data.php
```

canvural:
Like I said in it's current state the PR is not acceptable. It'll look for `config` path the current working directory. If we make it configurable users can configure it to their liking and we wouldn't need to deal with more issues.

I'm gonna close this PR if you are not willing to work on it but anyone feel free to submit a new one with the requested changes.

PR #2190, opened 2025-02-01
Test GenericStaticType

ondrejmirtes:
I just added support for `static<...>` to PHPStan: https://github.com/phpstan/phpstan-src/commit/0c1f22ad73fea06e7ef3c5ea0b936700057f6062 (to both 1.12.x and 2.1.x)

@canvural has some concerns about compatibility with Larastan so I'm opening PRs to test both 2.x and 3.x branches.

3.x here https://github.com/larastan/larastan/pull/2191


ondrejmirtes:
2.x isn't always green but the failure looks like is unrelated to PHPStan update.

PR #2191, opened 2025-02-01
Test GenericStaticType - 3.x

ondrejmirtes:
I just added support for `static<...>` to PHPStan: https://github.com/phpstan/phpstan-src/commit/0c1f22ad73fea06e7ef3c5ea0b936700057f6062 (to both 1.12.x and 2.1.x)

@canvural has some concerns about compatibility with Larastan so I'm opening PRs to test both 2.x and 3.x branches.

2.x here: https://github.com/larastan/larastan/pull/2190

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "illuminate/pipeline": "^11.15.0",
         "illuminate/support": "^11.15.0",
         "phpmyadmin/sql-parser": "^5.9.0",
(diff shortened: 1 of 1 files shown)


canvural:
Thank you! I'll try to take a look at it tonight

ondrejmirtes:
Well, 3.x is green, that's good news I think :)

PR #2193, opened 2025-02-03
[3.x] fix: collection template types being overwritten

calebdw:
Hello!

This fix was included with #1990, but it looks like it didn't make it into the release. This closes #2044 by preventing the collection template types from being overwritten.

Thanks!


calebdw:
Thanks!

calebdw:
@nunomaduro, could you please explain why this was reverted?

PR #2195, opened 2025-02-03
[3.x] deps: support Laravel 12.x

calebdw:
Closes #2192

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,7 +66,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0-dev"
+            "dev-master": "3.0-dev"
         },
         "phpstan": {
             "includes": [
--- a/composer.json
+++ b/composer.json
@@ -17,25 +17,25 @@
     "require": {
         "php": "^8.2",
         "ext-json": "*",
-        "illuminate/console": "^11.15.0",
-        "illuminate/container": "^11.15.0",
-        "illuminate/contracts": "^11.15.0",
-        "illuminate/database": "^11.15.0",
-        "illuminate/http": "^11.15.0",
-        "illuminate/pipeline": "^11.15.0",
-        "illuminate/support": "^11.15.0",
+        "illuminate/console": "^11.15.0 || ^12.0",
+        "illuminate/container": "^11.15.0 || ^12.0",
+        "illuminate/contracts": "^11.15.0 || ^12.0",
+        "illuminate/database": "^11.15.0 || ^12.0",
+        "illuminate/http": "^11.15.0 || ^12.0",
+        "illuminate/pipeline": "^11.15.0 || ^12.0",
+        "illuminate/support": "^11.15.0 || ^12.0",
         "phpmyadmin/sql-parser": "^5.9.0",
         "phpstan/phpstan": "^2.1.3"
     },
     "require-dev": {
         "doctrine/coding-standard": "^12.0",
         "nikic/php-parser": "^5.3",
-        "laravel/framework": "^11.15.0",
+        "laravel/framework": "^11.15.0 || ^12.0",
         "mockery/mockery": "^1.6",
-        "orchestra/canvas": "^v9.1.3",
-        "orchestra/testbench-core": "^9.5.2",
+        "orchestra/canvas": "^v9.1.3 || ^10.0",
+        "orchestra/testbench-core": "^9.5.2 || ^10.0",
         "phpstan/phpstan-deprecation-rules": "^2.0.0",
-        "phpunit/phpunit": "^10.5.16"
+        "phpunit/phpunit": "^10.5.35 || ^11.3.6"
     },
     "suggest": {
         "orchestra/testbench": "Using Larastan for analysing a package needs Testbench"
--- a/tests/Integration/IntegrationTest.php
+++ b/tests/Integration/IntegrationTest.php
@@ -22,7 +22,7 @@ class IntegrationTest extends PHPStanTestCase
         self::getContainer();
 
         yield [__DIR__ . '/data/bug-2074.php'];
-        yield [__DIR__ . '/data/test-case-extension.php', [39 => ['Call to function method_exists() with $this(TestTestCase) and \'partialMock\' will always evaluate to true.']]];
+        yield [__DIR__ . '/data/test-case-extension.php', [34 => ['Call to function method_exists() with $this(TestTestCase) and \'partialMock\' will always evaluate to true.']]];
         yield [__DIR__ . '/data/model-builder.php'];
         yield [__DIR__ . '/data/model-properties.php'];
         yield [__DIR__ . '/data/blade-view.php'];
--- a/tests/Integration/IntegrationTest.php
+++ b/tests/Integration/IntegrationTest.php
@@ -115,7 +115,7 @@ class IntegrationTest extends PHPStanTestCase
                 $this->assertNotEmpty($errors);
             }
 
-            $this->assertSameErrorMessages($expectedErrors, $errors);
+            $this->assertSameErrorMessages($file, $expectedErrors, $errors);
         }
     }
 
@@ -149,19 +149,23 @@ class IntegrationTest extends PHPStanTestCase
     }
 
     /**
-     *  @param array<int, array<int, string>> $expectedErrors
-     *  @param Error[]                        $errors
+     * @param array<int, array<int, string>> $expectedErrors
+     * @param Error[]                        $errors
      */
-    private function assertSameErrorMessages(array $expectedErrors, array $errors): void
+    private function assertSameErrorMessages(string $file, array $expectedErrors, array $errors): void
     {
         foreach ($errors as $error) {
             $errorLine = $error->getLine() ?? 0;
 
-            $this->assertArrayHasKey($errorLine, $expectedErrors);
+            $this->assertArrayHasKey(
+                $errorLine,
+                $expectedErrors,
+                sprintf('File %s has unexpected error "%s" at line %d.', $file, $error->getMessage(), $errorLine),
+            );
             $this->assertContains(
                 $error->getMessage(),
                 $expectedErrors[$errorLine],
-                sprintf("Unexpected error \"%s\" at line %d.\n\nExpected \"%s\"", $error->getMessage(), $errorLine, implode("\n\t", $expectedErrors[$errorLine])),
+                sprintf("File %s has unexpected error \"%s\" at line %d.\n\nExpected \"%s\"", $file, $error->getMessage(), $errorLine, implode("\n\t", $expectedErrors[$errorLine])),
             );
         }
     }
--- a/tests/Integration/data/test-case-extension.php
+++ b/tests/Integration/data/test-case-extension.php
@@ -7,27 +7,22 @@ class TestCaseExtension
 {
     public function testMockMethod(): void
     {
-        (new TestTestCase())->testMockMethod();
+        (new TestTestCase('foo'))->testMockMethod();
     }
 
     public function testPartialMockMethod(): void
     {
-        (new TestTestCase())->testPartialMockMethod();
+        (new TestTestCase('foo'))->testPartialMockMethod();
     }
 
     public function testSpyMethod(): void
     {
-        (new TestTestCase())->testSpyMethod();
+        (new TestTestCase('foo'))->testSpyMethod();
     }
 }
 
 class TestTestCase extends \Illuminate\Foundation\Testing\TestCase
 {
-    public function __construct()
-    {
-        parent::__construct('testcase');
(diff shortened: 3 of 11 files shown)


szepeviktor:
> Laravel 12 will be launched on February 24th, 2025.

calebdw:
@canvural, @szepeviktor, @ondrejmirtes,

This PR adds support for Laravel 12 as well as the `GenericStaticType` that's going to be added in the next PHPStan version (no need for #2191 anymore).

The addition of the new `GenericStaticType` means that the following extension are no longer needed (and they were in fact interfering with the `GenericStaticType` return types causing all the failures seen in #2191):
- src/ReturnTypes/EnumerableGenericStaticMethodDynamicStaticMethodReturnTypeExtension.php
- src/ReturnTypes/EnumerableGenericStaticMethodDynamicMethodReturnTypeExtension.php

Since these extensions were removed I just went ahead and moved the data files from their respective test cases into the `GeneralTypeTest`.

ondrejmirtes:
Hi, is this in coordination with @canvural? Because I'm talking to him about GenericStaticType and it'd be great to release the fixes for it in both 2.x and 3.x, because the new type is part of both PHPStan 1.12.x and 2.1.x.

I'd like to release the new PHPStan versions as soon as possible, I certainly don't want to wait until Laravel 12 is released.

calebdw:
This PR can be released as soon as the new PHPStan version is released, it doesn't have to wait until Laravel 12 is released. It just adds support for it so folks can start preparing for it (like mentioned in #2192)

ondrejmirtes:
Alright! Can you please test the combo of this PR and dev PHPStan version with GenericStaticType on a real-world project, to see if it causes new issues or not?

ondrejmirtes:
So that I can still squeeze some bugfixes into PHPStan before release.

tests/Type/GeneralTypeTest.php
```diff
@@ -63,13 +72,20 @@ public static function dataFileAsserts(): iterable
...
         yield from self::gatherAssertTypes(__DIR__ . '/data/where-relation.php');
-        yield from self::gatherAssertTypes(__DIR__ . '/data/bug-1997.php');
-        yield from self::gatherAssertTypes(__DIR__ . '/data/bug-1819.php');
 
         if (version_compare(LARAVEL_VERSION, '11.28.0', '>=')) {
             yield from self::gatherAssertTypes(__DIR__ . '/data/model-collections-l11-28.php');
         }
 
+        // 11.100 is an artificial constraint until 12.0.0 is released
+        if (version_compare(LARAVEL_VERSION, '11.0.0', '>=') && version_compare(LARAVEL_VERSION, '11.100.0', '<')) {
```
crynobone:
```suggestion
        if (laravel_version_compare('11.0.0', '>=') && laravel_version_compare('12.0.0', '<')) {
```

Don't forget to import `use function Orchestra\Testbench\laravel_version_compare;`

tests/Type/GeneralTypeTest.php
```diff
@@ -63,13 +72,20 @@ public static function dataFileAsserts(): iterable
...
+            yield from self::gatherAssertTypes(__DIR__ . '/data/collection-generic-static-methods-l11.php');
+        }
+
+        if (version_compare(LARAVEL_VERSION, '12.0.0', '>=') || LARAVEL_VERSION === '12.x-dev') {
```
crynobone:
```suggestion
        if (laravel_version_compare('12.0.0', '>=')) {
```

calebdw:
> Can you please test the combo of this PR and dev PHPStan version with GenericStaticType on a real-world project, to see if it causes new issues or not?

@ondrejmirtes, I just executed phpstan in our ERP codebase and everything is still green!

tests/Type/GeneralTypeTest.php
calebdw: @crynobone, likely need to add 12.x here as well:


image

tests/Type/GeneralTypeTest.php
```diff
@@ -63,13 +72,20 @@ public static function dataFileAsserts(): iterable
...
+        if (laravel_version_compare('11.28.0', '>=')) {
             yield from self::gatherAssertTypes(__DIR__ . '/data/model-collections-l11-28.php');
         }
 
+        // 11.100 is an artificial constraint until 12.0.0 is released
+        if (laravel_version_compare('11.0.0', '>=') && laravel_version_compare('11.100.0', '<')) {
+            yield from self::gatherAssertTypes(__DIR__ . '/data/collection-generic-static-methods-l11.php');
+        }
+
+        if (laravel_version_compare('12.0.0', '>=') || LARAVEL_VERSION === '12.x-dev') {
```
crynobone:
```suggestion
        if (laravel_version_compare('12.0.0', '>=')) {
```

Testbench Core already check for `12.x-dev`: https://github.com/orchestral/testbench-core/blob/c3d080c533cff1960aa0468dd930cce76fdfea5e/src/functions.php#L417-L430

calebdw: Ah sweet! must've been looking at the wrong version

canvural:
Hi,

Can you just keep the Laravel 12 support and drop the rest? Thank you.

calebdw:
Yes, I'll rebase this evening when I get the chance

calebdw:
@canvural, this has been rebased :+1: 


I went ahead and moved the data files from the deleted extension test cases to the `GeneralTypeTest` and removed the test cases. In the future we could presumably remove those data files as it's now testing PHPUnit and the framework and doesn't rely on any of our code---but it's good to leave in for now to prove that there's no breaking changes

composer.json
```diff
@@ -17,25 +17,25 @@
...
-        "illuminate/http": "^11.15.0",
-        "illuminate/pipeline": "^11.15.0",
-        "illuminate/support": "^11.15.0",
+        "illuminate/console": "^11.41.3 || ^12.0",
+        "illuminate/container": "^11.41.3 || ^12.0",
+        "illuminate/contracts": "^11.41.3 || ^12.0",
+        "illuminate/database": "^11.41.3 || ^12.0",
+        "illuminate/http": "^11.41.3 || ^12.0",
+        "illuminate/pipeline": "^11.41.3 || ^12.0",
+        "illuminate/support": "^11.41.3 || ^12.0",
```
canvural: Why are we also increasing the minimum 11.x version?

composer.json
```diff
@@ -17,25 +17,25 @@
...
         "phpmyadmin/sql-parser": "^5.9.0",
         "phpstan/phpstan": "^2.1.3"
     },
     "require-dev": {
         "doctrine/coding-standard": "^12.0",
         "nikic/php-parser": "^5.3",
-        "laravel/framework": "^11.15.0",
+        "laravel/framework": "^11.41.3 || ^12.0",
```
canvural: ditto

tests/Integration/IntegrationTest.php
canvural, line 1: Why the changes in this file?

tests/Integration/data/test-case-extension.php
canvural, line 1: ditto

tests/Rules/UnusedViewsRuleTest.php
```diff
@@ -42,14 +42,6 @@ protected function getCollectors(): array
...
     }
 
-    protected function setUp(): void
-    {
-        parent::setUp();
-
-        // This is a workaround for a weird PHPStan container cache issue.
-        require __DIR__ . '/../../bootstrap.php';
-    }
-
```
canvural, line 52: Why is it removed?

tests/Type/GeneralTypeTest.php
```diff
@@ -6,7 +6,7 @@
 
 use PHPStan\Testing\TypeInferenceTestCase;
 
-use function version_compare;
+use function Orchestra\Testbench\laravel_version_compare;
```
canvural, line 9: What does this function do differently?

canvural (changes requested):
Hi,

There are too many changes that are not really related to the Laravel 12 compatibility itself. If I run locally with Laravel 12 I only get 3 errors and they are all related to `partition` method. I do not undertsand why would we need the other changes.

Also, looks like there are some conflicts. Let's fix those and keep the changes in the PR to minimal. It also has some unnecessary style changes. If PHP_CodeSniffer doesn't change we don't need to change it.

tests/Type/GeneralTypeTest.php
calebdw: This function properly handles the `12.x-dev` version, otherwise you would have to do:

image

tests/Rules/UnusedViewsRuleTest.php
calebdw: This is no longer required and it was causing PHPUnit to report it as a risky test:

image

tests/Integration/data/test-case-extension.php
calebdw: These changes were required because the PHPUnit TestCase (which this extends) has a final, strong-typed constructor:

image

tests/Integration/IntegrationTest.php
calebdw: The line number changed because the constructor was removed

composer.json
calebdw: The minimum 11.x version was updated to fix compatibility issues on higher 11.x versions.

In my opinion, it is much cleaner and maintainable to fix the issue and raise the minimum version than to create two **nearly identical** files and have an if condition in the GeneralTypeTest like you did here: https://github.com/larastan/larastan/commit/cef7cfdfe6cf0905823e932e24046175dc6d05bd

composer.json
calebdw: same as above

tests/Integration/IntegrationTest.php
calebdw: @canvural,

The other changes in this file are for developer quality of life when there's a test failure.

I got really tired of tests failing with the message: "Failed asserting that array has the key x"
- I don't know what the actual error is
- I don't know what test file it occurred in

These changes now print the exact file, error, and line number that occurred making it much easier to diagnose the problem.

composer.json
canvural: I'd like to keep the support for wide range of versions.

tests/Rules/UnusedViewsRuleTest.php
canvural: hmm. But we do not need to use PHPUnit 11. Laravel 12 seems to still support PHPUnit 10

tests/Type/GeneralTypeTest.php
canvural: @crynobone Looks like this function is deprecated. Should we require sidekick here and use the function?

tests/Rules/UnusedViewsRuleTest.php
calebdw: I allow both and there's no reason not to support it---it's going to have to be done eventually.

The fact also remains that this "weird PHPStan container cache issue" no longer seems to exist given that the tests are passing just fine

tests/Type/GeneralTypeTest.php
calebdw: Done

calebdw: Actually this is not doable given your constraint to stay on `11.15`:

image

composer.json
calebdw: I'll update since you insist, however, this is not very useful given that a `composer update` is going to update people to the latest 11.x---there's no reason for anyone to not be on the latest minor version.

calebdw:
@canvural, this has been rebased and tests are passing.

> I do not undertsand why would we need the other changes.

I'm not making changes willy-nilly---there is a good reason for every change on this branch. If you're still confused on some changes then please ask.

> Also, looks like there are some conflicts. Let's fix those and keep the changes in the PR to minimal. It also has some unnecessary style changes. If PHP_CodeSniffer doesn't change we don't need to change it.

I have formatters in Neovim that autorun when I save. I usually try to avoid these changes but sometimes they slip through without me realizing it, although they shouldn't hurt anything

canvural:
> I have formatters in Neovim that autorun when I save. I usually try to avoid these changes but sometimes they slip through without me realizing it, although they shouldn't hurt anything

well they hurt the PR review process. I need to think if there was an actual change there or if it was only formatting. Like I said let's let the PHP_CodeSniffer to handle the formatting.

Anyway, thank you.

calebdw:
> well they hurt the PR review process. I need to think if there was an actual change there or if it was only formatting. Like I said let's let the PHP_CodeSniffer to handle the formatting.

I'll try to be better about not including changes like that in the future :+1: 

Thank you sir!

szepeviktor:
This PR is much nicer in its new, slim form.

PR #2196, opened 2025-02-03
[3.x] fix: baseline extension

calebdw:
@szepeviktor, this fix is needed for the e2e tests to run

Thanks!


szepeviktor:
All right.

Issue #2011, opened 2024-08-02
Casts defined in Trait are ignored


makroxyz:
- Larastan Version: 2.9.8
- Laravel Version: 11.19

### Description
Casts defined in Trait are ignored

### Laravel code where the issue was found

```php
trait MyTrait {

    public function initializeServiceTrait()
    {
        $this->mergeCasts([
             'price' => 'double',
             'quantity' => 'integer',
        ])
    }
}

```

szepeviktor:
Hello @makroxyz 👋🏻 

Is your trait in use?
PHPStan does not analyze traits.

makroxyz:
Yes it is, used in a couple of models

szepeviktor:
I think Larastan does not support `mergeCasts`.
https://github.com/search?q=repo%3Alarastan%2Flarastan%20mergeCasts&type=code

calebdw:
Yes, this is not supported. While we do use the `getCasts` method from the model (but not for the new L11 casts), we new up the model bypassing the constructor---which means traits are not booted/initialized:

https://github.com/larastan/larastan/blob/c8ec6f04f800d9425be743998574f1034e2ffb3f/src/Properties/ModelCastHelper.php#L240-L249

canvural:
Hi,

I'm not sure if this can be ever solved without doing more voodoo black magic type of things. So I'm gonna close it for now.

PR #2187, opened 2025-01-30
Fix/ulid or UUID primary key

Wulfheart:
- [x] Added or updated tests
- [x] Documented user facing changes (There are none)

Fixes #2188.

**Changes**

Currently, there is an error in larastan which treats primary uuids or ulids without further documentation as an int. Laravel however returns them as string. This PR fixes this by aligning larastan's behaviour to Laravel's. 

**Breaking changes**

None

Issue #2076, opened 2024-10-18
Internal error: Call to undefined method PHPStan\Type\ObjectType::getValue() while analysing file


Afrowson:
- Larastan Version: 2.9.9
- Laravel Version: 11.28.1

### Description
This is happening while upgrading to Laravel 11:
I cast Enums from the DB in to PHP Enums, if i do not annotate, i get warnings for trying to  calling a method on string. If i do annotate like the following, as described in [here](https://github.com/larastan/larastan/blob/2.x/UPGRADE.md#upgrading-to-296-from-295). My guess is that the detection of casts in to PHP Classes instead of basic types like array/int/string is not working correctly.

adding an old fashion style class `@property` annotation still works, but it would be nice if it couldd be detected from the array it self. 

### Laravel code where the issue was found
```php
    /**
     * @return array{ background_position:  ImagePosition, background_size:  ImageSize, background_repeat:  ImageRepeat, font_stack:  DesignFontStack }
     */
    protected function casts(): array
    {
        return [
            'background_position' => ImagePosition::class,
            'background_size' => ImageSize::class,
            'background_repeat' => ImageRepeat::class,
            'font_stack' => DesignFontStack::class,
        ];
    }
```

```
## /var/www/vendor/larastan/larastan/src/Properties/ModelCastHelper.php(261)                                                                                        
#0 [internal function]: Larastan\Larastan\Properties\ModelCastHelper::Larastan\Larastan\Properties\{closure}()                                                      
#1 /var/www/vendor/larastan/larastan/src/Properties/ModelCastHelper.php(261): array_map()                                                                           
#2 /var/www/vendor/larastan/larastan/src/Properties/ModelCastHelper.php(225): Larastan\Larastan\Properties\ModelCastHelper->getModelCasts()                         
#3 /var/www/vendor/larastan/larastan/src/Properties/ModelPropertyHelper.php(119): Larastan\Larastan\Properties\ModelCastHelper->getCastForProperty()                
#4 /var/www/vendor/larastan/larastan/src/Properties/ModelPropertyExtension.php(30): Larastan\Larastan\Properties\ModelPropertyHelper->getDatabaseProperty()         
#5 phar:///var/www/vendor/phpstan/phpstan/phpstan.phar/src/Reflection/ClassReflection.php(656): Larastan\Larastan\Properties\ModelPropertyExtension->getProperty()  
[...]

PR #2197, opened 2025-02-05
[3.x] fix: property type for uuid and ulid primary keys

### LINKED ISSUE #2188: Primary ulids and uuids are treated as int even though they are string
- Larastan Version: 3.x

There is already a fix in #2187.

### Description

Currently, there is an error in larastan which treats primary uuids or ulids without further documentation as an int. Laravel however returns them as string.

See #2187 for a failing test case.

calebdw:
Closes #2188, supersedes #2187

 
**Changes**

This PR allows Larastan to properly detect the correct model key type when using unique string ids.

We also had a lot of places in the code where we were obtaining a model instance, moving this logic to a helper class removed the duplication and cleans up the code.

This has been released on my [fork](https://github.com/calebdw/larastan) if anyone would like to take advantage of this immediately.

Thanks!

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -448,6 +448,9 @@ services:
             - phpstan.phpDoc.typeNodeResolverExtension
 
     -
+        class: Larastan\Larastan\Support\ModelHelper
+
+    -
         class: Larastan\Larastan\Properties\MigrationHelper
         arguments:
             databaseMigrationPath: %databaseMigrationsPath%
--- a/src/Properties/MigrationHelper.php
+++ b/src/Properties/MigrationHelper.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Larastan\Larastan\Properties;
 
+use Larastan\Larastan\Support\ModelHelper;
 use PHPStan\File\FileHelper;
 use PHPStan\Parser\Parser;
 use PHPStan\Parser\ParserErrorsException;
@@ -29,6 +30,7 @@ class MigrationHelper
         private FileHelper $fileHelper,
         private bool $disableMigrationScan,
         private ReflectionProvider $reflectionProvider,
+        private ModelHelper $modelHelper,
     ) {
     }
 
@@ -47,7 +49,7 @@ class MigrationHelper
             $this->databaseMigrationPath = [database_path('migrations')];
         }
 
-        $schemaAggregator = new SchemaAggregator($this->reflectionProvider, $tables);
+        $schemaAggregator = new SchemaAggregator($this->modelHelper, $this->reflectionProvider, $tables);
         $filesArray       = $this->getMigrationFiles();
 
         if (empty($filesArray)) {
--- a/src/Properties/ModelCastHelper.php
+++ b/src/Properties/ModelCastHelper.php
@@ -15,12 +15,12 @@ use Illuminate\Database\Eloquent\Casts\AsCollection;
 use Illuminate\Database\Eloquent\Casts\AsEncryptedArrayObject;
 use Illuminate\Database\Eloquent\Casts\AsEncryptedCollection;
 use Illuminate\Database\Eloquent\Casts\AsStringable;
-use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Carbon as IlluminateCarbon;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Date;
 use Illuminate\Support\Stringable as IlluminateStringable;
+use Larastan\Larastan\Support\ModelHelper;
 use PHPStan\Analyser\OutOfClassScope;
 use PHPStan\Reflection\ClassReflection;
 use PHPStan\Reflection\MissingMethodFromReflectionException;
@@ -41,7 +41,6 @@ use PHPStan\Type\ObjectType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
-use ReflectionException;
 use stdClass;
 use Stringable;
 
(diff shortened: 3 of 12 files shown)


canvural:
Hi,

Looks like [this](https://github.com/laravel/framework/pull/54096) the PR that broke it.

I think the real fix here is to use `newInstance` ? I don't like that actually even though it seems it's fine. We can check for if the model uses `HasUlids` or `HasUuids` traits like:
```diff
diff --git a/src/Properties/ModelCastHelper.php b/src/Properties/ModelCastHelper.php
index 13cf81e7..85fdf9df 100644
--- a/src/Properties/ModelCastHelper.php
+++ b/src/Properties/ModelCastHelper.php
@@ -15,6 +15,8 @@ use Illuminate\Database\Eloquent\Casts\AsCollection;
 use Illuminate\Database\Eloquent\Casts\AsEncryptedArrayObject;
 use Illuminate\Database\Eloquent\Casts\AsEncryptedCollection;
 use Illuminate\Database\Eloquent\Casts\AsStringable;
+use Illuminate\Database\Eloquent\Concerns\HasUlids;
+use Illuminate\Database\Eloquent\Concerns\HasUuids;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Carbon as IlluminateCarbon;
@@ -246,6 +248,13 @@ class ModelCastHelper

         $modelCasts = $modelInstance->getCasts();

+        if (
+            ($modelClassReflection->hasTraitUse(HasUlids::class) || $modelClassReflection->hasTraitUse(HasUuids::class))
+            && array_key_exists($modelInstance->getKeyName(), $modelCasts)
+        ) {
+            $modelCasts[$modelInstance->getKeyName()] = 'string';
+        }
+
         $castsMethodReturnType = $modelClassReflection->getMethod(
             'casts',
             new OutOfClassScope(),
```

calebdw:
@canvural, using `newInstance` also fixes other issues---there's other traits (from the framework and other packages) that rely on the trait booting. I'd rather go with this fix as-is than manually hardcoding these two traits.

canvural:
What are those other issues? I really don't like creating new objects like this. I'd rather add hardcoded exceptions that can keep the analysis static.

calebdw:
Well for one, this allows people to use the new `casts()` method without having to duplicate the array shapes on the method---that's been the cause of quite a few issues since Laravel 11. Many people don't like having to repeat the entire array in the PHPDoc above the method and there's be a lot of confusion on how to get it working (things like proper quoting and double slashes).

This also closes #2011 without doing any "voodoo black magic type of things", there's also other packages that define casts in traits (like my Laraflake package for snowflake ids).

It's worth noting that I do catch exceptions and fallback to `newInstanceWithoutConstructor` so this should not break anything. If everything works well then you get better analysis, and if something goes wrong then analysis continues but it might not know certain things.

Someone tried to PR a `#[Cast]` attribute (https://github.com/laravel/framework/pull/51948) to help static analysis but it wasn't accepted

nunomaduro:
@canvural Going to let you review this one.

raz-iacob:
Nice PR!

calebdw:
Unfortunately Github doesn't automatically create links to discussions, but here's another instance where this would solve the problem:

https://github.com/larastan/larastan/discussions/2264

redelschaap:
I really think this PR should be merged asap. It's such a pain in the ... to put all casts also in the docblock!

tr4pezohedron:
It's been about half a year since this PR was opened. Manually patching this is getting tiresome. Can it be merged, or is something blocking it? @nunomaduro @canvural

LWlook:
@nunomaduro @canvural Please merge this PR, it is such a big help!

@calebdw Thank you for this PR. I just switched to your package till this is merged.

itsrafsanjani:
Any updates on this one?

mengidd:
Yeah, would be nice to get this merged!

canvural:
Did it with 4894e25, which does _less_ runtime checks (still do not like it, but currently it is the easiest way)

calebdw:
> Did it with https://github.com/larastan/larastan/commit/4894e25455bc0773f83666125c0eaf95494a01e7, which does less runtime checks (still do not like it, but currently it is the easiest way)

This just completely disregarded all of the other linked issues that this PR would have fixed that's now left unresolved:
- #2011 
- #2076 
- #2250 
- #2261 
- #2264 
- https://github.com/driftingly/rector-laravel/issues/336
- #2330
- #2387 
- https://github.com/larastan/larastan/pull/2408#discussion_r2568928246

The primary issue that this would have fixed is having to duplicate the casts in `@return` phpdocs on the `casts()` methods. The community does not really like having to do that and there has been a ton of confusion about how exactly to format it so that Larastan properly recognizes it (e.g., https://github.com/larastan/larastan/pull/2408#discussion_r2568928246).

This also would have allowed any package or end user that uses `->mergeCasts()` in a booted trait to work out of the box

I've been using this in an enterprise codebase for almost a year now with no ill effects.

mengidd:
I would also ask to reconsider the chosen solution for this, Calebs solution fixes some annoying issues related to casts that would be nice to have fixed.

canvural:
Hi,

I understand your concerns. But something we should not forget is that this is supposed to be _static_ analysis. Sure, Larastan bootstraps the framework, which defeats the purpose, but I'd say that is necessary evil and still something I do not like too much. 

You can go back and check my comments when the first `casts` method future came to Laravel/Larastan. I've always said I'm against adding more runtime checks, and even proposed alternatives (like parsing the file to get the returned array)

Also, I've seen code bases that do crazy things like fetching the casts from the database. Sure it sounds crazy, but doing a runtime stuff here would break the analysis for them. And we need to think about all the users, even though it's the minority. 

All the other linked issues should be fixed separately. The fix I did was only for #2188 

And as a side note, recently I added an option (`parseModelCastsMethod`, not yet released) that when enabled would parse the `casts` method and understand the returned array thus removing the need for the return docblock with array shape. So as you can see, everything has a solution. 

It's easy to see a couple lines of change that seems to fix an issue and wanted it to be merged. But we need to understand the overall picture.

PR #2198, opened 2025-02-06
[2.x] Supports Laravel 12

crynobone:
While there's already Larastan 3 (PHPStan 2) it would still be useful to support PHPStan 1, especially for Laravel packages where most will still support Laravel versions lower than 11 (Larastan 3 only support PHPStan 2 and Laravel 11).

Sure Laravel packages can use `"larastan/larastan": "^2.0|^3.0"` but the different PHPStan and Larastan formatting make testing different versions almost impossible (you either locked into Larastan 2 or Larastan 3)


calebdw:
Thanks for this, but Larastan v2 only supports up the Laravel 11.15---Laravel 12 support isn't going to be added to this version.

I'm in the process of trying to add L12.x support to the 3.x branch but ran into failures and haven't had time to fix

crynobone:
This is unfortunate since it would be useful to still support PHPStan 1, especially for Laravel packages where most will still support Laravel versions lower than 11 (Larastan 3 only support PHPStan 2 and Laravel 11). 

Sure Laravel packages can use `"larastan/larastan": "^2.0|^3.0"` but the different PHPStan and Larastan formatting make testing different version almost impossible (you either locked into Larastan 2 or Larastan 3)

calebdw:
That was a decision @canvural made, not sure if he'd be willing to support it. 


By the way, do you know if Laravel 11.x supports PHPUnit 11? I was getting risky test failures because a test didn't unregister error/exception handlers. It looked like a fix was merged to Laravel 12, but the Laravel 11 composer json allows PHPUnit 11...

crynobone:
> do you know if Laravel 11.x supports PHPUnit 11? 

It definitely supports PHPUnit 11, and will support PHPUnit 12.0 (available tomorrow)

calebdw:
> It definitely supports PHPUnit 11,

Any idea why there's a risky test failure on Laravel 11 & PHPUnit 11, but not with Laravel 11 & PHPUnit 10?

https://github.com/larastan/larastan/actions/runs/13172112603/job/36764217549?pr=2195

Edit: nevermind, this seemed to be the cause: https://github.com/larastan/larastan/pull/2195/commits/d494dd89c8cc299e8c3edb1941bbbe9082012640

.github/workflows/static-analysis.yml
```diff
@@ -25,15 +25,15 @@ jobs:
       fail-fast: false
       matrix:
         php: ["8.0", "8.1", "8.2", "8.3"]
-        laravel: ["^9.0", "^10.0", "^11.16"]
+        laravel: ["^9.0", "^10.0", "^11.0"]
```
calebdw: should have 12.x here as well right?

calebdw (approved):

.github/workflows/static-analysis.yml
crynobone: fixed

canvural:
Hi,

I really don't want to support 4 different versions at the same time. It's already chaotic enough. And I did not get the issue with doing `"larastan/larastan": "^2.0|^3.0"` in packages? 🤔 If you are on Laravel 11 you should be able to upgrade to Larastan 3 easily. So I don't see the problem.

On the other hand, Larastan 2.x will be maintained until PHPStan 1.x is no longer maintained.

canvural:
> Thanks for this, but Larastan v2 only supports up the Laravel 11.15

Actually [looks like](https://github.com/larastan/larastan/blob/d0df39658a8c5172b63afc6bd5cb84f457d4d80c/composer.json#L20) we support `^11.16` in 2.x. The table in the readme is wrong.

crynobone:
> If you are on Laravel 11 you should be able to upgrade to Larastan 3 easily. So I don't see the problem.

Are you certain that `phpstan analyse` generated from Larastan 2 will be the same as Larastan 3? If that the case the argument for `^2.0|^3.0` is valid.

canvural:
> > If you are on Laravel 11 you should be able to upgrade to Larastan 3 easily. So I don't see the problem.
> 
> Are you certain that `phpstan analyse` generated from Larastan 2 will be the same as Larastan 3? If that the case the argument for `^2.0|^3.0` is valid.

Of course not 😅 PHPStan itself had a major version increase and that alone will cause new errors that needs to be fixed.

I see your point now actually. But package maintainers can use different PHPStan config and baseline for this. At the end it'll not affect the end users, and I really don't want to juggle 4 different Laravel versions.

PR #2209, opened 2025-02-13
[3.x] Add connection to Redis facade

Riley19280:
- [x] Added or updated tests

**Changes**

Exposes redis connection methods that are available to the facade. They are exposed in Laravel via `__call` to the Connection class.
https://github.com/laravel/framework/blob/11.x/src/Illuminate/Redis/RedisManager.php#L276

Currently, using the `Redis::client()` method expects a parameter to be passed, when that is not actually the case, because the `\Redis` extension expects a parameter. Mixing in the connection class first fixes this.

### DIFF
--- a/stubs/common/Facades.stub
+++ b/stubs/common/Facades.stub
@@ -5,6 +5,7 @@ namespace Illuminate\Support\Facades;
 abstract class Facade {}
 
 /**
+ * @mixin \Illuminate\Redis\Connections\Connection
  * @mixin \Redis
  * @mixin \Illuminate\Redis\RedisManager
  * @mixin \Illuminate\Contracts\Redis\Factory
--- a/tests/Type/data/facades.php
+++ b/tests/Type/data/facades.php
@@ -43,6 +43,7 @@ function test(): void
     assertType('bool', Storage::cloud()->deleteDirectory('foo'));
     assertType('string|false', Storage::putFile('foo', 'foo/bar'));
     assertType('mixed', Redis::get('foo'));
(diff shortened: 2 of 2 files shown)


stubs/common/Facades.stub
```diff
@@ -5,6 +5,7 @@ namespace Illuminate\Support\Facades;
 abstract class Facade {}
 
 /**
+ * @mixin \Illuminate\Redis\Connections\Connection
  * @mixin \Redis
```
szepeviktor, line 9:
```suggestion
 * @mixin \Redis
 * @mixin \Illuminate\Redis\Connections\Connection
```

szepeviktor:
Thank you.

We already have those 51 failing assertions.
https://github.com/larastan/larastan/actions/runs/13319507210/job/37201241002

stubs/common/Facades.stub
Riley19280: The connection needed to be before `\Redis` so that the `client()` method signature on the connection class takes priority

Yedfree (approved):

stubs/common/Facades.stub
```diff
@@ -5,6 +5,7 @@ namespace Illuminate\Support\Facades;
 abstract class Facade {}
 
 /**
+ * @mixin \Illuminate\Redis\Connections\Connection
```
canvural, line 8: Can we use `\Illuminate\Redis\RedisManager` instead?

nunomaduro: It's already there on the line 10.

PR #2217, opened 2025-02-26
Add possibility to specify config directory

### LINKED ISSUE #2173: Feature request/Bug report: possibility to specify config/view directory

AJenbo:
Fixes #2173
Replaces #2183 (solution based on https://github.com/larastan/larastan/pull/2183#issuecomment-2671705754)

**Changes**
This adds a config parameter for specifying the paths where `env()` is allowed:

```neon
parameters:
    configDirectories:
        - src/config
        - tests
```

This is useful for packages with vendor configs, or applications that uses `env()` to drive some of there tests.

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -396,6 +396,15 @@ parameters:
     noEnvCallsOutsideOfConfig: false
 ```
 
+By default, this rule checks for env calls outside the application config directory. If your configuration files are stored elsewhere, you can use the configDirectories option to specify them.
+
+```neon
+parameters:
+    configDirectories:
+        - src/config
+        - tests
+```
+
 ## ModelAppendsRule
 
 Checks model's `$appends` property for computed properties. The properties added to `$appends` array should both exist in the model and be computed properties.
--- a/extension.neon
+++ b/extension.neon
@@ -19,6 +19,7 @@ parameters:
     databaseMigrationsPath: []
     disableMigrationScan: false
     disableSchemaScan: false
+    configDirectories: []
     viewDirectories: []
     checkModelProperties: false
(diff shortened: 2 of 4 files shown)


systemsolutionweb:
It would be great if it had better support for default path, like #2218
I mean like `/config`, which is usually the config directory in packages

AJenbo:
Can the folder be read from the application somehow? If not then I don't think it's possible to automate it like in #2218.

systemsolutionweb:
I already saw that #2183 did it and it was rejected

AJenbo:
Ok, well that isn't helpful since I made this PR based on the information in that one. If you can figure out how to properly getting the package config folder in an automated fashion then feel free to point it out and I can look at implementing a PR for it and we can see if it will be accepted by the maintainer.

Issue #2128, opened 2024-11-25
AppMakeHelper::resolveTypeFromCall() never returns ErrorType, doesn't throw exception, but pretends `mixed`


rudiedirkx:
- Larastan Version: v2.9.11
- Laravel Version: v11.33.2
- Phpstan version: 1.12.10

### Description

`AppMakeHelper::resolveTypeFromCall()` resolves a service from the container, and if that fails, it's supposed to return an `ErrorType`, but it never does, because the 'if that fails' exception is caught twice.

`resolveTypeFromCall()` calls `HasContainer::resolve()` which also has a try/catch, so that's where the exception is caught, and `null` is returned, so there is never an exception to catch in `resolveTypeFromCall()`.

But that doesn't even seem to be the 'real' problem. If `resolveTypeFromCall` finds a `null` service (or the never-thrown exception is caught), it returns a `new ErrorType()`:

```php
if ($resolved === null) {
    return new ErrorType();
}
```

but that's never used, and Phpstan will just assume the service is `mixed`.

### Laravel code where the issue was found

Anywhere a service is requested:

```php
app(MyService::class)->doSomething();
```

Result is not an `**ERROR**` from Phpstan, but a stan `mixed` error:

> `Cannot call method doSomething() on mixed.`

Larastan's intention is to make the error very clear, and I agree, but it doesn't work. Maybe Phpstan is too forgiving? It accepts `app()`'s `@return mixed` even if an extension returns an `ErrorType`? I don't know.

It's a very sneaky bug. My local stan worked perfectly, but on Github CI it failed with above `mixed` error. Not because `app()` returns mixed, but because Larastan can't resolve the service (because of a `null` config that didn't exist in Github CI's `.env`). Maybe Larastan shouldn't catch the exception at all, and just let the stan process crash? Make the error very obvious.

If Phpstan/Larastan crashes on an extension exception it's very obvious something failed-failed, not stan-failed, and it tells you what:

```
 -- --------------------------------------------------------------------------------------------------------------------
     Error
 -- --------------------------------------------------------------------------------------------------------------------
     Internal error: Cannot assign null to property App\Service\MyService::$someConfig of type string while
     analysing file app/Observers/MyObserver.php

     This message is coming from Laravel Framework itself.
     Larastan boots up your application in order to provide
     smarter static analysis of your codebase.

     In order to do that, the environment you run PHPStan in
     must match the environment you run your application in.

     Make sure you've set your environment variables
     or the .env file correctly.
[...]

PR #2223, opened 2025-03-05
Add fallback to existing classes/interfaces for improved type resolution

AJenbo:
### Overview:
This change improves type resolution in Larastan by falling back to the class string if it cannot be instantiated in the current environment, but only if the given class or interface exists.

### Problem:
When Larastan cannot instantiate a class or interface due to missing configuration or other environment-specific issues, it falls back to `mixed` without a clear indication. This leads to missing or ambiguous type information, requiring manual type hints, an error-prone ultimately and unnecessary workaround.

### Solution:
- Check if the class or interface exists before giving up.
- If it exists, use the class/interface as the type.
- If it does not exist, the error handling remains unchanged.

### Benefits:
- **Improved Type Resolution**:  Safely preserves type information when instantiation fails.
- **Compatibility**: No breaking changes; maintains existing behavior.
- **Fewer Workarounds**: Reduces the need for manual type hints when Larastan previously failed to resolve types.

This change makes Larastan's type resolution more reliable and predictable, helping developers avoid manual workarounds and improving overall analysis in dynamic environments.

It should also resolve the issue that @rudiedirkx encountered in #2128, which stemmed from environment differences between local development and GitHub CI. Based on testing, this fix eliminates around 140 issues in my application, which contains 3,700 classes, and makes an unknown number of workarounds redundant.

### DIFF
--- a/src/ReturnTypes/AppMakeHelper.php
+++ b/src/ReturnTypes/AppMakeHelper.php
@@ -43,7 +43,7 @@ final class AppMakeHelper
                     $resolved = $this->resolve($constantString->getValue());
 
                     if ($resolved === null) {
-                        if ($constantString->isClassString()) {
+                        if ($constantString->isClassString()->yes()) {
                             $types[] = $constantString->getClassStringObjectType();
                             continue;
                         }
--- a/src/ReturnTypes/AppMakeHelper.php
+++ b/src/ReturnTypes/AppMakeHelper.php
@@ -43,8 +43,8 @@ final class AppMakeHelper
                     $resolved = $this->resolve($constantString->getValue());
 
                     if ($resolved === null) {
-                        if (class_exists($constantString->getValue()) || interface_exists($constantString->getValue())) {
-                            $types[] = new ObjectType($constantString->getValue());
+                        if ($constantString->isClassString()) {
+                            $types[] = $constantString->getClassStringObjectType();
                             continue;
                         }
 
--- a/src/ReturnTypes/AppMakeHelper.php
+++ b/src/ReturnTypes/AppMakeHelper.php
@@ -16,7 +16,9 @@ use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use Throwable;
 
+use function class_exists;
 use function count;
+use function interface_exists;
 
 final class AppMakeHelper
 {
@@ -41,6 +43,11 @@ final class AppMakeHelper
                     $resolved = $this->resolve($constantString->getValue());
 
                     if ($resolved === null) {
+                        if (class_exists($constantString->getValue()) || interface_exists($constantString->getValue())) {
+                            $types[] = new ObjectType($constantString->getValue());
+                            continue;
+                        }
+
                         return new ErrorType();
                     }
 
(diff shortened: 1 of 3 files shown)


AJenbo:
@szepeviktor I fixed t he code style issue could you run the the actions again?

nunomaduro:
Can you add tests for this please?

AJenbo:
@nunomaduro done

AJenbo:
@nunomaduro anything else needed for this? I would love to  start removing the workarounds from our app :)

src/ReturnTypes/AppMakeHelper.php
```diff
@@ -41,6 +43,11 @@ public function resolveTypeFromCall(FuncCall|MethodCall|StaticCall $call, Scope
                     $resolved = $this->resolve($constantString->getValue());
 
                     if ($resolved === null) {
+                        if (class_exists($constantString->getValue()) || interface_exists($constantString->getValue())) {
```
canvural: We can't use `class_exists` We need to inject ` PHPStan\Reflection\ReflectionProvider` and ask with `hasClass`

canvural (changes requested):

src/ReturnTypes/AppMakeHelper.php
```diff
@@ -41,6 +43,11 @@ public function resolveTypeFromCall(FuncCall|MethodCall|StaticCall $call, Scope
...
+                            $types[] = new ObjectType($constantString->getValue());
+                            continue;
+                        }
```
calebdw: You can directly use the constant string api:

```suggestion
                        if ($constantString->isClassString()) {
                            $types[] = $constantString->getClassStringObjectType();
                            continue;
                        }
```

AJenbo: Hm PHPStan didn't like this change

calebdw: Oh you may need `->yes()` on the condition

src/ReturnTypes/AppMakeHelper.php
```diff
@@ -41,6 +43,11 @@ public function resolveTypeFromCall(FuncCall|MethodCall|StaticCall $call, Scope
                     $resolved = $this->resolve($constantString->getValue());
 
                     if ($resolved === null) {
+                        if ($constantString->isClassString()) {
```
calebdw: sorry, this returns a TrinaryLogic instead of a boolean

```suggestion
                        if ($constantString->isClassString()->yes()) {
```

AJenbo: Thanks hadn't had time to really look at ityet.

src/ReturnTypes/AppMakeHelper.php
AJenbo: This is now fixed.

src/ReturnTypes/AppMakeHelper.php
```diff
@@ -16,7 +16,9 @@
 use PHPStan\Type\TypeCombinator;
 use Throwable;
 
+use function class_exists;
 use function count;
+use function interface_exists;
```
AJenbo:
```suggestion
use function count;
```

AJenbo:
The failing tests appear unrelated to my changes.

szepeviktor:
> The failing tests appear unrelated to my changes.

Yes https://github.com/larastan/larastan/actions/runs/15502277881

PR #2231, opened 2025-03-19
Fix compatibility with Laravel 12.3.0

AJenbo:
Fixes: #1774
Fixes: #2233

Depends on https://github.com/larastan/larastan/pull/2208

PR #2234, opened 2025-03-24
Only set macro methods as static in facades

simon-tma:
- [ ] Added or updated tests

**Changes**

With phpstan-strict-rules installed, errors are generated when calling a static method on an instance or when calling an instance method statically.

Larastan is currently adding all macro methods that are defined using a closure as static methods to the base class. In most cases, this is incorrect as macro methods are instance methods. The only case where they should be static methods is when being added to a class behind a facade.

Unable to add a test as this requires strict rules to be installed.


simon-tma:
Test failures are waiting on #2231, not related to this change.

nunomaduro:
can you rebase and add a few tests?

PR #2240, opened 2025-03-29
[3.x] Use generics in Arrayable::toArray

### LINKED ISSUE #2239: Arrayable::toArray Generics
# Feature Request

Currently, Larastan's stub for Arrayable does not use generics, in the toArray method (whilst [Laravel's implementation does](https://github.com/laravel/framework/blob/450f8803ce022b3b73c7b97f45e2cea82e10e0bf/src/Illuminate/Contracts/Support/Arrayable.php#L14).

```php
use App\Models\User;
use Illuminate\Contracts\Support\Arrayable;

/**
 * @implements Arrayable<string, mixed>
 */
class TaskData implements Arrayable
{
    public function __construct(
        protected string $name,
        protected User $owner,
    ) {}

    public function toArray(): array
    {
        return [
            'name' => $this->name,
            'owner_id' => $this->owner->id,
        ];
    }
}

$user = User::first();
$data = new TaskData('Wash the Car', $user);
$array = $data->toArray();

\PHPStan\dumpType($array) // array<mixed> when I'd like this to be array<string, mixed>
```

For anyone else with this issue (until Larastan supports it), just add the desired docblock above the toArray method like so:

```php
/**
 */
public function toArray(): array
{
    return [
        'name' => $this->name,
        'owner_id' => $this->owner->id,
    ];
}
```

liamduckett:
Hey, heads up - first contribution to Larastan.

- [N/A] Added or updated tests
- [N/A] Documented user facing changes

Resolves #2239

**Changes**

Uses Arrayable's generics, for the return type of it's toArray method.

**Breaking changes**

I don't think this is a breaking change despite potentially raising new issues, as it just improves analysis. This seems to be inline with [PHPStan's BC Promise](https://phpstan.org/user-guide/backward-compatibility-promise#type-inference-capabilities)?

### DIFF
--- a/stubs/common/Contracts/Support.stub
+++ b/stubs/common/Contracts/Support.stub
@@ -14,7 +14,7 @@ interface Arrayable
     /**
      * Get the instance as an array.
      *
-     * @return array<mixed>
+     * @return array<TKey, TValue>
      */
     public function toArray();
 }


liamduckett:
Are the test failures unrelated - waiting for https://github.com/larastan/larastan/pull/2231?

nunomaduro:
Can you re-submit this pull request but with a bunch of tests? Don't forget about nested toArray calls.

PR #2249, opened 2025-04-03
[3.x] fix: collection template types being overwritten

calebdw:
Hello!

This was merged in #2193 and then reverted so reopening.

This closes #2044 by preventing the collection template types from being overwritten.

This has been released on my [fork](https://github.com/calebdw/larastan) if anyone would like to take advantage of this immediately.

Thanks!

### DIFF
--- a/src/ReturnTypes/EloquentBuilderExtension.php
+++ b/src/ReturnTypes/EloquentBuilderExtension.php
@@ -14,11 +14,11 @@ use PHPStan\Reflection\MethodReflection;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
-use PHPStan\Type\Generic\GenericObjectType;
-use PHPStan\Type\IntegerType;
-use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
 
+use function collect;
+use function count;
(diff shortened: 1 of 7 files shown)


nunomaduro:
@calebdw how confident are you about this PR? And that i won't break existing larastan apps?

calebdw:
@nunomaduro,

I'm very confident in this PR, as you can see from #2044 PHPStan handles this properly and this is a bug in Larastan.

This is also needed to propagate the `pivot` property from the BelongsToMany relation into the Collection generics (which is why the tests failed because I forgot to update this PR after #2231 was merged)

That said, sure some users might experience some warnings and need to go update the types, but this is in line with the [PHPStan BC promise](https://phpstan.org/user-guide/backward-compatibility-promise#type-inference-capabilities) about PHPStan getting smarter when bugs are fixed causing the output to change

AJenbo:
@nunomaduro is there any chance we could get this merged, there are features we can't use in Laravel (accessing the pivot) without this fix.

CoolDude53:
Wondering when this will get merged...

apasquini95:
Any update on this PR? It could be very useful if merged :)

pindab0ter:
I would also like to see this bug fix merged.

AJenbo:
@canvural is there anything we can do to help land this? Would it be helpful if I do a review?

pindab0ter:
I'm assuming the closing of this issue relates to [the notice in the top of the readme of your fork](https://github.com/calebdw/larastan/blob/3.x/README.md).

PR #2254, opened 2025-04-04
Arrayable generics

### LINKED ISSUE #2239: Arrayable::toArray Generics

liamduckett:
Hey, re-submit of #2240 ([as per](https://github.com/larastan/larastan/pull/2240#issuecomment-2776769327)).

    [✅] Added or updated tests
    [N/A] Documented user facing changes

Resolves https://github.com/larastan/larastan/issues/2239

Changes:
1. Uses Arrayable's generics, for the return type of the toArray method.
2. Added some tests


tests/application/app/ValueObjects/FooList.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace App\ValueObjects;
+
+use Illuminate\Contracts\Support\Arrayable;
+
+/**
+ * @implements Arrayable<int, Arrayable<string, mixed>>
```
calebdw: You're testing that Arrayable::toArray can output specific types instead of the `mixed` default so this should not used `mixed` here---make it more specific

calebdw: Also, there's no need for two different value objects---a single one should suffice

liamduckett: Hey - appreciate the speedy review. I should have resolved the first issue.

As for the second one, how should I approach [this](https://github.com/larastan/larastan/pull/2240#issuecomment-2776769327).

> Don't forget about nested toArray calls

calebdw: The nested `toArray` calls has to do with Collections recursively calling `->toArray()` on items that are Arrayable.

However, that is not applicable here because this is just a simple interface---any class that implemented a recursive `toArray` solution would need to implement the interface as `@implement Arrayable<TKey, mixed>` for instance

You should be able to remove the FooList value object and test

tests/application/app/ValueObjects/Foo.php
calebdw, line 1: This simple object can just live in the arrayable.php file under the test function, there's no need for it to be across the codebase

liamduckett: Appreciate your time on this!

tests/Type/data/arrayable.php
```diff
@@ -0,0 +1,23 @@
...
+    {
+
+    }
+}
+
+function test(): void
+{
+    $foo = new Foo;
+    assertType('array<string, int>', $foo->toArray());
+}
```
calebdw: This is good, but here's how we generally write the tests:

```suggestion
function test(Foo $foo): void
{
    assertType('array<string, int>', $foo->toArray());
}

/** @implements Arrayable<string, int>*/
class Foo implements Arrayable
{
    public function toArray()
    {
        return [];
    }
}
```


However, it's up to you if you want to make the changes.

calebdw (approved):
Looks good! Thanks!

tests/Type/data/arrayable.php
liamduckett: Thanks. Any reason we leave the return type declaration from toArray?

calebdw: You can add the php return type, but there's no reason to duplicate the `@return` because it's already defined on the parent:

```php
{
    public function toArray(): array
    {
        return [];
    }
```


You can also use `/** @inheritDoc */` or `#[Override]` (preferred, but PHP >=8.4) if that helps you remember, but it's not necessary:

```php
{
    /** @inheritDoc */
    public function toArray(): array
    {
        return [];
    }
```

liamduckett: I guess my personal attitude has always been that static analysis (or code analysis in this case) should build on top of the native type system - rather than replace it entirely.

Fully appreciate this is a rather simple example - just curious if this is the approach you (given your wider experience) would take for other situations too, and if so why?

stubs/common/Contracts/Support.stub
```diff
@@ -14,7 +14,7 @@ interface Arrayable
     /**
      * Get the instance as an array.
      *
-     * @return array<mixed>
+     * @return array<TKey, TValue>
```
calebdw: I just realized that the correct generics exist in the framework---there's no reason for us to duplicate them.

- the `toArray` method here can just be removed
- there's no need for the added test

liamduckett: I've made those changes - I also included a slight tweak to the Model stub, to prevent some failing tests:

```
1) Type\GeneralTypeTest::testFileAsserts with data set "<redacted>/larastan/tests/Type/data/helpers.php:141" ('type', '<redacted>...rs.php', 'array<mixed>|null', 'array|null', 141)
Expected type array<mixed>|null, got type array|null in <redacted>/larastan/tests/Type/data/helpers.php on line 141.
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'array<mixed>|null'
+'array|null'

phar:///<redacted>/larastan/vendor/phpstan/phpstan/phpstan.phar/src/Testing/TypeInferenceTestCase.php:82
<redacted>/larastan/tests/Type/GeneralTypeTest.php:94

2) Type\GeneralTypeTest::testFileAsserts with data set "<redacted>/larastan/tests/Type/data/helpers.php:142" ('type', '<redacted>...rs.php', 'array<mixed>', 'array', 142)
Expected type array<mixed>, got type array in <redacted>/larastan/tests/Type/data/helpers.php on line 142.
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'array<mixed>'
+'array'
```

I appreciate this isn't particularly important (as they represent the same type) - and should probably be narrowed anyway - but changing this felt out of scope for this PR? I've left it as a commit on the end, so it can be popped off if required.

stubs/common/Model.stub
```diff
@@ -71,6 +71,13 @@ abstract class Model implements \JsonSerializable, \ArrayAccess
      * @return \Illuminate\Database\Eloquent\Builder<static>
      */
     public static function with($relations);
+
+    /**
+     * Convert the model instance to an array.
+     *
+     * @return array<mixed>
+     */
+    public function toArray();
```
calebdw, line 81: This shouldn't be needed as Model implements Arrayable, consider updating the Model stub to:

```php

/**
 * @implements \ArrayAccess<string, mixed>
 * @implements Arrayable<string, mixed>
 */
abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToString, HasBroadcastChannel, Jsonable, JsonSerializable, QueueableEntity, Stringable, UrlRoutable
```

liamduckett: I've left a comment about this in the other thread, mind taking a look before I do anything here?

calebdw: I saw, but the model `toArray` technically is `array<string, mixed>` and not `array<mixed>`. I think it would be fine to update in this PR

liamduckett: I've done that. Truthfully I'm not sure how valuable the interface is here, as Laravel's Model specifies:

```php
  /**
   * Convert the model instance to an array.
   *
   * @return array
   */
  public function toArray()
  {
    //
  }
```

So it needed overwriting anyway - [see example](https://phpstan.org/r/5f6a8ed8-1998-4c0b-a9da-abbae8ae4826)

calebdw: Ah, that's lovely :upside_down_face:

liamduckett: I'll leave it in anyway, as it does no harm. Appreciate your continued time on this!

PR #2270, opened 2025-04-24
Post and query can also be array with a key given (fixes #2268)

### LINKED ISSUE #2268: InteractsWithInput stub has wrong return typehints
- Larastan Version: 2.11.0
- Laravel Version: 11.x/12.x

### Description
The new InteractsWithInput stubs has wrong return types. It now assumes that `$request->query(...)` cannot return an array. But as the Laravel 'InteractsWithInput' trait gives, it can.

https://github.com/laravel/framework/blob/12.x/src/Illuminate/Http/Concerns/InteractsWithInput.php
https://github.com/laravel/framework/blob/11.x/src/Illuminate/Http/Concerns/InteractsWithInput.php

https://github.com/larastan/larastan/blob/2.x/stubs/common/Http/InteractsWithInput.stub

```php
        $searchTerm = request()->query('searchTerm');
        if (is_array($searchTerm)) {
            $searchTerm = null;
```
This gives that '$searchTerm' cannot be an array. But when passing `&searchterm[]=foo` in a request, `$searchTerm` really becomes: `$searchTerm = ['foo'];`

rikvdh:
Resolves #2268

**Changes**

Added the array type as query and post can return an array with the key set.

**Breaking changes**

No

### DIFF
--- a/stubs/common/Http/InteractsWithInput.stub
+++ b/stubs/common/Http/InteractsWithInput.stub
@@ -10,7 +10,7 @@ trait InteractsWithInput
      * @template TDefault
      * @param  string|null  $key
      * @param  TDefault  $default
-     * @return ($key is null ? array : string|TDefault)
+     * @return ($key is null ? array : string|array|TDefault)
      */
     public function query($key = null, $default = null);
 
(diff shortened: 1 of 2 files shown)


szepeviktor:
Thank you!

Could you add tests?

rikvdh:
@szepeviktor I was looking for tests for InteractsWithInput to extend them. But they were recently added without tests (as far as I can see).

szepeviktor:
Here https://github.com/larastan/larastan/blob/3.x/tests/Type/data/request-object.php ?

rikvdh:
I'll try to make something. I see that the post and query methods are not tested there at all.

szepeviktor:
Very good ❤️

szepeviktor:
Thank you.

PR #2276, opened 2025-04-29
Fix NoEnvCallsOutsideOfConfigRule to handle Windows paths properly

adgower:
- [x] Added or updated tests

Resolves an issue where NoEnvCallsOutsideOfConfigRule fails to correctly identify files in the config directory on Windows systems.

**Changes**

This PR fixes an issue with the NoEnvCallsOutsideOfConfigRule where it fails to correctly identify files in the config directory on Windows systems due to different path separators.

The changes include:
- Modified the `isCalledOutsideOfConfig` method to normalize both the file path and config directory path before comparison using `str_replace('\\', '/', $path)`
- Added the necessary `str_replace` function import
- Added a test that simulates Windows paths to verify the fix works correctly
- Updated the documentation to mention cross-platform compatibility

On Windows, file paths use backslashes (`\`) as directory separators, while on Unix-like systems, forward slashes (`/`) are used. The previous implementation used `str_starts_with()` to compare paths directly, which failed when the paths used different separators.

By normalizing both paths to use forward slashes before comparison, the rule now works correctly on both Windows and Unix-like systems.

**Breaking changes**

None. This change maintains backward compatibility and only improves the functionality on Windows systems.

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -1,6 +1,6 @@
 # Rules
 
-All rules that are specific to Laravel applications 
+All rules that are specific to Laravel applications
 are listed here with their configurable options.
 
 ## NoModelMake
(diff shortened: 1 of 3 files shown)


canvural:
Fixed in https://github.com/larastan/larastan/pull/2298 instead

PR #2287, opened 2025-05-17
[3.x] Fix: Public scopes using attribute

### LINKED ISSUE #2286: Scope attribute generates method.notFound error, but old scope prefix works
- Larastan Version: 3.4.0
- Laravel Version: 12.12.0

### Description

When using the Scope attribute on a model, and calling it, I get the following error:
```
Call to an undefined method Illuminate\Database\Eloquent\Builder<App\Models\Connection>::active().  
🪪 method.notFound
```
If I change the scope to use the old prefix format (`scopeActive`), it doesn't generate any errors.

See the previous [discussion](https://github.com/larastan/larastan/discussions/2281#discussioncomment-13160271) for context.

I'm using a custom relationship that extends HasMany in Account:
```php
    /**
     * @return HasManyConnections<Connection, $this>
     */
    public function connections(): HasManyConnections
    {
        return new HasManyConnections($this->newRelatedInstance(Connection::class)->newQuery(), $this);
    }
```
With an `active` scope:
```php
    /** @param Builder<static> $query */
    #[Scope]
    public function active(Builder $query): void
    {
        $query->where('request_accepted', true);
    }
```
And I get the error when trying to eager load active connections:
```php
    Account::whereHas('connections', function ($query) {
        return $query->active()->where('uses_default', true);
    })
```

I tested this without using the custom relation and using a regular HasMany, and I still get the error, so I don't think that's affecting it.

liamduckett:
- [ 🟢 ] Added or updated tests
- [ N/A ] Documented user facing changes

Resolves #2286 

**Changes**

Currently, Larastan does not detect public methods that are declared with the `Scope` attribute as valid scopes. This seems to have [been done intentionally](https://github.com/larastan/larastan/pull/2244) however I cannot immediately see why, as public scopes are valid (see original issue). This brings detection logic for scopes via the `Scope` attribute inline with other methods.

**Breaking changes**

Public scopes declared using the `Scope` attribute will now be recognised as valid scopes by Larastan. I do not think this warrants any documentation.

### DIFF
--- a/src/Methods/BuilderHelper.php
+++ b/src/Methods/BuilderHelper.php
@@ -153,7 +153,7 @@ class BuilderHelper
                     }
                 }
 
-                if (! $methodReflection->isPublic() && $hasScopeAttribute) {
+                if ($hasScopeAttribute) {
                     $parametersAcceptor = $methodReflection->getVariants()[0];
 
                     $parameters = $parametersAcceptor->getParameters();
--- a/tests/Type/data/model-scope-attribute-l12.php
+++ b/tests/Type/data/model-scope-attribute-l12.php
@@ -17,7 +17,6 @@ function test(Builder $query): void
     assertType('Illuminate\Database\Eloquent\Builder<ModelScopeAttribute\Activity>', Activity::withVoidReturn());
     assertType('Illuminate\Database\Eloquent\Builder<ModelScopeAttribute\Activity>', $query->withVoidReturn());
     assertType('ModelScopeAttribute\Activity', $query->withVoidReturn()->firstOrFail());
-    assertType('*ERROR*', $query->notAScope());
 }
 
 class Activity extends Model
@@ -35,13 +34,6 @@ class Activity extends Model
     /** @param  Builder<static>  $query */
     #[Scope]
     protected function withVoidReturn(Builder $query): void
-    {
-        $query->where('whyuse', 'void');
-    }
-
-    /** @param  Builder<static>  $query */
-    #[Scope]
-    public function notAScope(Builder $query): void
     {
         $query->where('whyuse', 'void');
     }


calebdw:
I don't care for this. Scopes, like computed attributes, should be protected as they are not supposed to be called directly on the class.

szepeviktor:
@liamduckett What do you think? Scope methods could be public?
https://laravel.com/docs/12.x/eloquent#local-scopes

liamduckett:
@szepeviktor I think that scopes (in Laravel) **can** be public - and as such, Larastan should support that.

calebdw:
> and as such, Larastan should support that

I still disagree---PHPStan / Larastan is often much stricter than what is technically allowed in PHP / Laravel. This is no different than the [Laravel 9 Attributes](https://github.com/larastan/larastan/blob/3.x/docs/features.md#laravel-9-attributes) which Larastan also requires to be `protected`

liamduckett:
I think that makes sense - do you think it'd make sense to add the same "they must be protected" warning / message in the scopes section?

Additionally, do you think it'd be feasible / reasonable for Larastan to report something like "scope (and attribute for consistency?) should be protected"?

szepeviktor:
> do you think

I'm just a janitor here.

calebdw:
I would be okay with adding something about it in the docs, but the scope discovery doesn't happen in a place that's easy to report an error in Larastan

liamduckett:
> I still disagree---PHPStan / Larastan is often much stricter than what is technically allowed in PHP / Laravel. This is no different than the [Laravel 9 Attributes](https://github.com/larastan/larastan/blob/3.x/docs/features.md#laravel-9-attributes) which Larastan also requires to be protected

@calebdw how do you feel about the current inconsistency between whether public scopes are currently recognised - based on whether they are defined via attribute or method name? 

I don't feel like requiring scopes defined by method name to be protected is a feasible change, as people are likely relying on this behaviour - unless we implemented this:

> Additionally, do you think it'd be feasible / reasonable for Larastan to report something like "scope (and attribute for consistency?) should be protected"?

Personally I think I'm still on the side of changing to allowing public scopes (in attribute declared scopes) - despite only ever using `protected` ones myself - mostly due to historical decisions that have been made (like allowing public scopes in method name declared scopes). I assume a maintainer will make the call as to which approach (if any) to go with?

Synchro:
It just took me a while to track down this problem. The thing that I'd got wrong was declaring the scope method as `public`. The way that is reported is quite unhelpful, as it simply makes it look like Larastan doesn't support the attribute syntax. A notice or warning saying that scopes are expected to be `protected` would have been really helpful!

calebdw:
[rector-laravel](https://github.com/driftingly/rector-laravel) can now automatically make model attributes and scopes `protected` for you: https://github.com/driftingly/rector-laravel/pull/370

i-bajrai:
If the convention is that scopes should be protected (why would you want it to be public when Laravel's magic is handling it?) then public scopes should fail. 

I don't think this PR should go ahead.

Somehow I'm getting this error for my scopes using the old way, maybe I'm doing something wrong 🤔 

```
Call to an undefined method Illuminate\Database\Eloquent\Relations\HasMany::completed().  
         🪪  method.notFound
```

calebdw:
@i-bajrai, please create a new issue, and show the full code

canvural:
I agree with the overall comments here. So I'm gonna close this PR, but then we can add a new rule that warns about model scopes being `public` That way we can guide the developers to the correct solution.

PR #2297, opened 2025-06-17
[3.x] Fix: Address the return type for Date::createFromFormat and Date::createSafe

walkonthemarz:
Address the return type of Date::createFromFormat and Date::createSafe.

Date::createFromFormat and Date::createSafe returns `Illuminate\Support\Carbon|null`, not `Illuminate\Support\Carbon|false`.

See https://api.laravel.com/docs/12.x/Illuminate/Support/DateFactory.html#method_createFromFormat


walkonthemarz:
Closing this as this duplicates #2299, which has broader compatibility.

PR #2298, opened 2025-06-18
Replace slash with DIRECTORY_SEPARATOR on rule `NoEnvCallsOutsideOfConfigRule`

sertxudev:
**Changes**

This PR allows Windows users to define in the `phpstan.neon` file custom `configDirectories`.

```
parameters:
    configDirectories:
        - config
        - modules/*/Config
```

Before this change only Unix-like system could use this functionality as the absolute path was wrong for Windows users.

Before PR: `C:\Users\User\Herd\laravel\modules/ModuleA/Config/config.php`
After PR: `C:\Users\User\Herd\laravel\modules\ModuleA\Config\config.php`

The `glob` method returns the path following the directory separator in the provided pattern.

Before the PR this condition was always false for Windows users as the path did not match:

https://github.com/larastan/larastan/blob/36706736a0c51d3337478fab9c919d78d2e03404/src/Rules/NoEnvCallsOutsideOfConfigRule.php#L83

### DIFF
--- a/src/Rules/NoEnvCallsOutsideOfConfigRule.php
+++ b/src/Rules/NoEnvCallsOutsideOfConfigRule.php
@@ -33,12 +33,13 @@ class NoEnvCallsOutsideOfConfigRule implements Rule
     public function __construct(private array $configDirectories, private FileHelper $fileHelper)
     {
         if (count($configDirectories) !== 0) {
+            /** @var list<non-empty-string> $normalizedDirectories */
             $normalizedDirectories = [];
-    
+
             foreach ($configDirectories as $directory) {
                 $normalizedDirectories[] = $this->fileHelper->normalizePath($directory);
             }
-    
+
             $this->configDirectories = $normalizedDirectories;
 
             return;
--- a/src/Rules/NoEnvCallsOutsideOfConfigRule.php
+++ b/src/Rules/NoEnvCallsOutsideOfConfigRule.php
@@ -33,9 +33,13 @@ class NoEnvCallsOutsideOfConfigRule implements Rule
     public function __construct(private array $configDirectories, private FileHelper $fileHelper)
     {
         if (count($configDirectories) !== 0) {
-            foreach ($this->configDirectories as &$directory) {
-                $directory = $this->fileHelper->normalizePath($directory);
+            $normalizedDirectories = [];
+    
+            foreach ($configDirectories as $directory) {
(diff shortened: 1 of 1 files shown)


sertxudev:
There's already the PR https://github.com/larastan/larastan/pull/2276 targeting this issue but this makes fewer changes.

src/Rules/NoEnvCallsOutsideOfConfigRule.php
```diff
@@ -73,6 +73,7 @@ public function processNode(Node $node, Scope $scope): array
     protected function isCalledOutsideOfConfig(FuncCall $call, Scope $scope): bool
     {
         foreach ($this->configDirectories as $configDirectoryGlob) {
+            $configDirectoryGlob = str_replace('/', DIRECTORY_SEPARATOR, $configDirectoryGlob);
```
canvural: I think we can do this in constructor. And even use `$this->fileHelper->normalizePath(...)`. It covers some other edge cases too. What do you think?

sertxudev: I've updated the code, good idea to use the `normalizePath` method!

src/Rules/NoEnvCallsOutsideOfConfigRule.php
```diff
@@ -33,6 +33,10 @@ class NoEnvCallsOutsideOfConfigRule implements Rule
     public function __construct(private array $configDirectories, private FileHelper $fileHelper)
     {
         if (count($configDirectories) !== 0) {
+            foreach ($this->configDirectories as &$directory) {
```
canvural: It'll be very nitpicky, but we should avoid using references if we can. Can we make it not promoted property and then we can loop over it and assign it to the array property?

szepeviktor: This is not nitpicking at all.

canvural: @sertxudev I meant declaring `private array $configDirectories` in class level, and accepting just `array $configDirectories` as an argument in the constructor.

sertxudev: > @sertxudev I meant declaring `private array $configDirectories` in class level, and accepting just `array $configDirectories` as an argument in the constructor.

Gotcha!

canvural:
Thank you!

PR #2301, opened 2025-06-18
[3.x] `getCountForPagination` as a pass-thru for Laravel versions >= 12.15

cosmastech:
- [x] Added or updated tests
- [x] ~~Documented user facing changes~~

This passes through to the underlying Builder now. See: https://github.com/laravel/framework/pull/55752

**Changes**

Versions >= 12.15.0 will return an int from `Eloquent\Builder@getCountForPagination()`.
Other versions will continue to return the Eloquent Builder.

**Breaking changes**
No breaking changes to my knowledge.


cosmastech:
Ooof. I'll have to look into this more. This is my first time contributing, and I'm not sure how the patterns work here. This would only be a passthru IF the version if 12.15.0 or later.

Any input would be appreciated.

calebdw:
> This would only be a passthru IF the version if 12.15.0 or later.

I don't think it really matters, no one would use this under 12.15 anyways

cosmastech:
@calebdw are there tests to update for this? I took a quick scan to look for tests verifying existing builder methods in the passthru and didn't see much.

canvural:
> Ooof. I'll have to look into this more. This is my first time contributing, and I'm not sure how the patterns work here. This would only be a passthru IF the version if 12.15.0 or later.
> 
> Any input would be appreciated.

Yeah this needs to be conditional. `User::query()->getCountForPagination()` would return builder on <12.15.0 but return `int` on >= 12.15.0

You can use `laravel_version_compare('12.15.0', '>=')` to add it to the array conditionally. And for the tests, you can use any eloquent builder test file.

szepeviktor:
AFAIK Testbench is not available for applications, only for packages.
https://github.com/larastan/larastan/blob/36706736a0c51d3337478fab9c919d78d2e03404/composer.json#L13-L25

tests/Type/data/eloquent-builder.php
```diff
@@ -342,6 +343,13 @@ function test(
     assertType('Illuminate\Database\Eloquent\Builder<App\Team|App\User>', $userOrTeamBuilder->where('id', 5));
 
     assertType('Illuminate\Database\Eloquent\Builder<TModel of Illuminate\Database\Eloquent\Model (function EloquentBuilder\test(), argument)>', $templateBuilder->select());
+
+    if (laravel_version_compare('12.15.0', '<')) {
+        assertType('Illuminate\Database\Query\Builder', User::query()->getCountForPagination());
+    }
+    else {
+        assertType('int', User::query()->getCountForPagination());
+    }
```
canvural: As a convention, we do not do the condition check inside the test cases like this.As it wouldn't work correctly.  We do this inside the data provider. In this case in `GeneralTypeTest` There you can also see how it is done for example.

cosmastech: Thanks for the insight!

How do I correctly resolve the application version? Is it safe to use this inside of the `BuilderHelper@__construct()` method?

```php
if (version_compare(LARAVEL_VERSION, '12.15.0', '<')) {
    return;
}
```

---

Re: `GeneralTypeTest`: should I create a new eloquent-builder-l12-15.php for this type assertion?

canvural: > How do I correctly resolve the application version? Is it safe to use this inside of the BuilderHelper@__construct() method?

Aah, I was referring to source code with the `laravel_version_compare` suggestion. Sorry for confusion. Inside the `src` directory we can use `laravel_version_compare` function. I think we already do it in one place, you can search for it.

> GeneralTypeTest: should I create a new eloquent-builder-l12-15.php for this type assertion?

Yes, please. And conditionally `yield` depending on version.

canvural: @cosmastech You need to ignore the error on that line with `@phpstan-ignore-line` Because we run analysis with different versions of Laravel.

canvural:
Perfect. Thank you!

cosmastech:
> Perfect. Thank you!

Thanks for all your help on this. 🙇 

There are a few other issues I'd like to clean up and now that I _somewhat_ understand the testing flow, it should be easier.

PR #2302, opened 2025-06-19
feat: add support for UseEloquentBuilder attribute

canvural:
**Changes**

Adds support for `UseEloquentBuilder` attribute

### DIFF
--- a/tests/Type/GeneralTypeTest.php
+++ b/tests/Type/GeneralTypeTest.php
@@ -86,6 +86,10 @@ class GeneralTypeTest extends TypeInferenceTestCase
             yield from self::gatherAssertTypes(__DIR__ . '/data/model-scope-attribute-l12.php');
         }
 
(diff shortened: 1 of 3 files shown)


tests/Type/GeneralTypeTest.php
```diff
@@ -86,6 +86,10 @@ public static function dataFileAsserts(): iterable
             yield from self::gatherAssertTypes(__DIR__ . '/data/model-scope-attribute-l12.php');
         }
 
+        if (laravel_version_compare('12.15.0', '>=')) {
```
cosmastech, line 89: This should 12.19.0, no?

canvural: Yes, fixed in a commit later: https://github.com/larastan/larastan/commit/e8ccd73008487ba91da9877b373f8c447743f1ce

PR #2311, opened 2025-07-09
[3.x] feat: add new rule to detect unnecessary Enumerable::toArray() calls

calebdw:
- [x] Added or updated tests

**Changes**

Hello!

This creates a new rule that's disabled by default.

If none of the items in an Enumerable are Arrayables, then there's no need to call [`toArray`](https://github.com/laravel/framework/blob/fac2d778f5f3c899cbcf8890011f7cd0727568bf/src/Illuminate/Collections/Traits/EnumeratesValues.php#L946-L954)---`all()` can be used instead.

Thanks!

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -84,6 +84,41 @@ parameters:
     noUnnecessaryCollectionCallExcept: ['contains']
 ```
 
+## NoUnnecessaryEnumerableToArrayCalls
+
+This rule checks for unnecessary calls `Enumerable::toArray()` that
+could have used `all()` instead. The `toArray()` method recursively
+converts all Arrayable items in the Enumerable to an array and if
+none of the items are Arrayable, it is unnecessary map call.
+
+### Examples
+
+```php
+collect([1, 2, 3])->toArray();
+```
+
+Will result in the following error:
+
+```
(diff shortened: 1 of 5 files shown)


szepeviktor:
Would you like to take a look at [failing tests](https://github.com/larastan/larastan/blob/b9159cf7f75a668f6341895cbe8d4dcbebd3a775/tests/Type/data/custom-eloquent-builder.php#L23) in another PR?

canvural:
Hi,

Makes sense I think. I'll test it with some projects later, before releasing. Thanks.

calebdw:
Thanks! I've tested in my work project and it works pretty great!

The warning only shows up when we're 100% sure that the values are not Arrayable.


It seems like this case isn't quite supported:

```php
$types = OrderType::query()
    ->orderBy('id')
    ->pluck('name', 'id')
    ->toArray();
```

But this is because the `->pluck()` call just returns `mixed` for the value type---I think we need to update the dynamic method call extension to properly detect the collection template types when plucking. We currently do this when checking if the `pluck` can be called on a query instead of the collection so it should be possible

AJenbo:
Really grate feature, I see people use `toArray() `a lot when they should have used `all()`.

PR #2313, opened 2025-07-12
fix: match covariant TValue of collections with Laravel

Bloemendaal:
- [ ] Added or updated tests

**Changes**
In the Laravel code of the Collection the TValue is used as template-covariant, see https://github.com/illuminate/collections/blob/master/Collection.php. For me, using it as covariant solves the following issue:

```php
<?php

enum MyEnum: string
{
    case Foo = 'foo';
    case Bar = 'bar';
}

class MyClass
{
    /**
     * @return Collection<int, MyEnum>
     */
    public function getSomeEnumValues(): Collection
    {
        return new Collection([MyEnum::Foo]);
    }
}
```

Currently, phpstan will give the following error:

```
Method MyClass::getSomeEnumValues() should return Illuminate\Support\Collection<int, MyEnum> but returns  
Illuminate\Support\Collection<int, MyEnum::Foo>.                                                                         
🪪 return.type  
```

However, when the `@template-covariant` is used instead of `@template` this issue (like in the Laravel source) doesn't exist.

In this function above I could've just changed the return type, but in my actual code the `getSomeEnumValues` is actually part of an interface that the class implements, meaning I cannot really change the return type to match every enum value.

**Breaking changes**

I don't think anything breaks

### DIFF
--- a/stubs/common/Collection.stub
+++ b/stubs/common/Collection.stub
@@ -8,7 +8,7 @@ use Illuminate\Support\Traits\EnumeratesValues;
 
 /**
  * @template TKey of array-key
- * @template TValue
+ * @template-covariant TValue
  *
  * @implements \ArrayAccess<TKey, TValue>
  * @implements \Illuminate\Support\Enumerable<TKey, TValue>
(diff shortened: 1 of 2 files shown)


Bloemendaal:
I don't know how to fix these invariant issues... When I use try out the laravel version in phpstan sandbox it gives me the same issues as shown in these tests... 😐

PR #2315, opened 2025-07-12
[3.x] Add support for public const in Schema table names

riasvdv:
- [x] Added or updated tests

**Changes**

This adds support for public constants in the Schema's table name

### DIFF
--- a/src/Properties/SchemaAggregator.php
+++ b/src/Properties/SchemaAggregator.php
@@ -111,14 +111,43 @@ final class SchemaAggregator
 
     private function alterTable(PhpParser\Node\Expr\StaticCall|PhpParser\Node\Expr\MethodCall $call, bool $creating): void
     {
-        if (
-            ! isset($call->args[0])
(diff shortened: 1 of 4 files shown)


szepeviktor (approved):

src/Properties/SchemaAggregator.php
```diff
@@ -111,14 +112,43 @@ private function addUpMethodStatements(array $stmts): void
...
+                return;
+            }
+
+            $class = $this->reflectionProvider->getClass($value->class->name);
+
+            $constantValueType = $class->getConstant($value->name->toString())->getValueType();
+
+            if (method_exists($constantValueType, 'getValue')) {
+                $tableName = $constantValueType->getValue();
+            }
```
canvural:
```suggestion
            if ($constantValueType->getConstantStrings() !== []) {
                $tableName = $constantValueType->getConstantStrings()[0]->getValue();
            }
```

We can use type system like this to do the same thing.

canvural:
Thank you!

PR #2317, opened 2025-07-14
[3.x] feat: add optional noModelForwardingToBuilder and noModelStaticForwardingToBuilder for concise model usage

Smurrlawa:
- [x] Added or updated tests  

**Changes**

This PR introduces 2 new rules `noModelForwardingToBuilder ` and `noModelStaticForwardingToBuilder ` that detects wrong usage of model methods and static methods.

**Breaking changes**

None. Existing use cases are unaffected. This rules only adds an additional safety check to improve code correctness. The rules are optional.

**Motivation**

Calling `first()` on a model instance is often a logic error that may cause unexpected database queries and subtle bugs. This first rule helps developers catch such mistakes early during static analysis. This issue was encountered when a junior developer refactored an `->where()->first()` clause to `->find()` but accidentally left the `->first()` call, resulting in always fetching the first record in the database unexpectedly.

The second rule encourages the usage of query() before each static method.

---

**Example model methods**

```php
$user = User::find(1);
$user->first(); // Redundant and potentially problematic
```

**Example static model methods**

```php
User::first();
```


calebdw:
I think a more useful rule would be to completely disable non-static builder mixins on model instances---for example you have to start a query with `::query()` or another `::method()` (even though I don't particularly care for this).

This would effectively eliminate all of the following, not just `first()`
```php
$model->first();
$model->firstOrFail();
$model->firstOrCreate();
$model->where();
$model->value('...');
$model->with();
// etc
```

Smurrlawa:
> I think a more useful rule would be to completely disable non-static builder mixins on model instances---for example you have to start a query with `::query()` or another `::method()` (even though I don't particularly care for this).
> 
> This would effectively eliminate all of the following, not just `first()`
> 
> ```
> $model->first();
> $model->firstOrFail();
> $model->firstOrCreate();
> $model->where();
> $model->value('...');
> $model->with();
> // etc
> ```

Hi @calebdw,
you are correct!

I wanted to make an first example for this but indeed we should expand this further to also include the other calls as well.

Yes it is very niche but our project pushed to prod because of this wrong usage of eloquent and I thought this may be a good addition for some rare edge cases 😄 _(also having more tests now to cover these things too in our project ^^)_

calebdw:
Personally, I'd like to see two optional rules, for increasing levels of strictness:
1. `NoModelForwardingToBuilder`: this would prevent all methods like `$model->first()`, but still allow methods like `Model::first()`
2. `NoModelStaticForwardingToBuilder`: this would be more strict and prevent all methods like `Model::first()`, instead requiring `Model::query()->first()`


The first rule would target `MethodCall` nodes and the second would target `StaticCall` nodes. Let me know if you need help with this, it should be a simple as getting the declaring class (either via `getPrototype` or `getDeclaringClass`) and returning the error if that is not `->is(Model::class)`

src/Rules/NoFirstOnSingularModelRule.php
```diff
@@ -0,0 +1,95 @@
...
+     *
+     * @return array<RuleError>
+     *
+     * @throws ShouldNotHappenException
+     */
+    public function processNode(Node $node, Scope $scope): array
+    {
+        if (! $node instanceof MethodCall) {
+            return [];
+        }
```
calebdw: There is no need for this if check---this is guaranteed via the `getNodeType` method

src/Rules/NoFirstOnSingularModelRule.php
```diff
@@ -0,0 +1,95 @@
...
+
+/** @implements Rule<MethodCall> */
+final class NoFirstOnSingularModelRule implements Rule
+{
+    /**
+     * Specifies the node type this rule is interested in.
+     * We are interested in method calls.
+     *
+     * @return class-string<Node>
+     */
```
calebdw: There is no need to duplicate the lengthy phpdocs---you can actually remove them completely, but if you insist on having them then just use:


```suggestion
    /** @inheritDoc */
```

src/Rules/NoFirstOnSingularModelRule.php
```diff
@@ -0,0 +1,95 @@
...
+            return [];
+        }
+
+        if (! $node->name instanceof Identifier || $node->name->toLowerString() !== 'first') {
+            return [];
+        }
+
+        $callerType = $scope->getType($node->var);
+
+        if ($callerType instanceof UnionType) {
```
calebdw: This is not the best way to traverse a type, here you should be able to call `getObjectClassReflections()` on the `$callerType` and iterate over them

Smurrlawa:
> Personally, I'd like to see two optional rules, for increasing levels of strictness:
> 
> 1. `NoModelForwardingToBuilder`: this would prevent all methods like `$model->first()`, but still allow methods like `Model::first()`
> 2. `NoModelStaticForwardingToBuilder`: this would be more strict and prevent all methods like `Model::first()`, instead requiring `Model::query()->first()`
> 
> The first rule would target `MethodCall` nodes and the second would target `StaticCall` nodes. Let me know if you need help with this, it should be a simple as getting the declaring class (either via `getPrototype` or `getDeclaringClass`) and returning the error if that is not `->is(Model::class)`

Great idea! Will do.
But I guess the `NoModelStaticForwardingToBuilder` should be disabled by default? Still many of my projects still use the none `query()`style ^^
Do we want to enable the other one by default then?

calebdw:
> Do we want to enable the other one by default then?

No, generally when adding new rules they should be disabled by default or you will get complaints by people for "breaking changes".

If the rule is not super opinionated / strict then they can be made the default in the next major release

src/Rules/NoFirstOnSingularModelRule.php
Smurrlawa: > This is not the best way to traverse a type, here you should be able to call `getObjectClassReflections()` on the `$callerType` and iterate over them

Sry I tried several ways but couln't get it to work with the reflections, could you provide an example please?
Rest of the MR changes to have the 2 rules is already in work tough.

Also do we have a list of all model functions we need to check agains?

calebdw: > Also do we have a list of all model functions we need to check agains?

There is no need to do this, for any method call we can find the declaring class and determine if that was forwarded to the builder.

These two rules should work:

```php
<?php

declare(strict_types=1);

namespace Tests\Architecture\Rules;

use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
use Illuminate\Database\Query\Builder as QueryBuilder;
use PhpParser\Node;
use PhpParser\Node\Expr\StaticCall;
use PhpParser\Node\Identifier;
use PhpParser\Node\Name;
use PHPStan\Analyser\Scope;
use PHPStan\Rules\Rule;
use PHPStan\Rules\RuleErrorBuilder;

/** @implements Rule<StaticCall> */
final class NoModelStaticForwardingToBuilder implements Rule
{
    /** @inheritDoc */
    public function getNodeType(): string
    {
        return StaticCall::class;
    }

    /** @inheritDoc */
    public function processNode(Node $node, Scope $scope): array
    {
        $errors = [];

        $calledMethods = $node->name instanceof Identifier
            ? [$node->name->toString()]
            : array_map(
                fn ($name) => $name->getValue(),
                $scope->getType($node->name)->getConstantStrings(),
            );

        $calledOnType = $node->class instanceof Name
            ? $scope->resolveTypeByName($node->class)
            : $scope->getType($node->class);

        foreach ($calledOnType->getObjectClassReflections() as $classReflection) {
            if (! $classReflection->is(Model::class)) {
                continue;
            }

            foreach ($calledMethods as $method) {
                if (! $classReflection->hasMethod($method)) {
                    continue;
                }

                $methodReflection = $classReflection->getMethod($method, $scope);
                $declaringClass   = $methodReflection->getDeclaringClass();

                if (
                    ! $declaringClass->is(QueryBuilder::class)
                    && ! $declaringClass->is(EloquentBuilder::class)
                ) {
                    continue;
                }

                $errors[] = RuleErrorBuilder::message("Static method [{$method}] is forwarded to a Builder instance, which is not allowed.")
                    ->tip("Use [::query()->{$method}()] instead.")
                    ->identifier('larastan.noModelStaticForwardingToBuilder')
                    ->line($node->name->getStartLine())
                    ->build();
            }
        }

        return $errors;
    }
}
```

```php
<?php

declare(strict_types=1);

namespace Tests\Architecture\Rules;

use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
use Illuminate\Database\Query\Builder as QueryBuilder;
use PhpParser\Node;
use PhpParser\Node\Identifier;
use PHPStan\Analyser\Scope;
use PHPStan\Rules\Rule;
use PHPStan\Rules\RuleErrorBuilder;

/** @implements Rule<MethodCall> */
final class NoModelForwardingToBuilder implements Rule
{
    /** @inheritDoc */
    public function getNodeType(): string
    {
        return MethodCall::class;
    }

    /** @inheritDoc */
    public function processNode(Node $node, Scope $scope): array
    {
        $errors = [];

        $calledMethods = $node->name instanceof Identifier
            ? [$node->name->toString()]
            : array_map(
                fn ($name) => $name->getValue(),
                $scope->getType($node->name)->getConstantStrings(),
            );

        $calledOnType = $scope->getType($node->var);

        foreach ($calledOnType->getObjectClassReflections() as $classReflection) {
            if (! $classReflection->is(Model::class)) {
                continue;
            }

            foreach ($calledMethods as $method) {
                if (! $classReflection->hasMethod($method)) {
                    continue;
                }

                $methodReflection = $classReflection->getMethod($method, $scope);
                $declaringClass   = $methodReflection->getDeclaringClass();

                if (
                    ! $declaringClass->is(QueryBuilder::class)
                    && ! $declaringClass->is(EloquentBuilder::class)
                ) {
                    continue;
                }

                $errors[] = RuleErrorBuilder::message("Method [{$method}] is forwarded to a Builder instance, which is not allowed.")
                    ->tip("Use [::{$method}()], [::query()->{$method}()] or [->newQuery()->{$method}()] instead.")
                    ->identifier('larastan.noModelForwardingToBuilder')
                    ->line($node->name->getStartLine())
                    ->build();
            }
        }

        return $errors;
    }
}
```

Smurrlawa: Thank you for the implementation!
Couldn't have done it at that level. Never worked with reflections yet... ^^

Yes but we need a list of methods we want to test inside the Unit test.
That's what I meant

calebdw: Ah, should be able to use any of the methods in my previous comment that forward to EloquentBuilder, for the QueryBuilder you'd have to find methods that forward there (maybe `orderBy` or `max/min`?)


If you want I could put all of this in a PR that supersedes this, though you're welcome to update this PR with these changes (will have to update the title and description).

Smurrlawa: > Ah, should be able to use any of the methods in my previous comment that forward to EloquentBuilder, for the QueryBuilder you'd have to find methods that forward there (maybe `orderBy` or `max/min`?)
> 
> If you want I could put all of this in a PR that supersedes this, though you're welcome to update this PR with these changes (will have to update the title and description).

I am already doing the PR right now, thanks for the help!
Yes I'm searching for all query builder and eloquent builder methods.

e.g.
```php
$user->first();
$user->get();
$user->findOrFail();
$user->find();
$user->paginate();
$user->simplePaginate();
$user->cursor();
$user->where('id', 1);
$user->orWhere('id', 1);
$user->orderBy('id');
$user->groupBy('id');
$user->having('id', '>', 1);
...
```

calebdw: > Yes I'm searching for all query builder and eloquent builder methods.

You don't need to do **all** methods, just do maybe 2-3 with one QueryBuilder method

tests/Rules/NoModelForwardingToBuilderTest.php
```diff
@@ -0,0 +1,36 @@
...
+            ["Method [where] is forwarded to a Builder instance, which is not allowed.\n    💡 Use [::where()], [::query()->where()] or [->newQuery()->where()] instead.", 9],
+            ["Method [take] is forwarded to a Builder instance, which is not allowed.\n    💡 Use [::take()], [::query()->take()] or [->newQuery()->take()] instead.", 10],
+        ]);
+    }
+
+    /** @return string[] */
+    public static function getAdditionalConfigFiles(): array
+    {
+        return [__DIR__ . '/../phpstan-tests.neon'];
+    }
```
calebdw, line 36: I'm pretty sure these tests will pass without this additional config, and if that's the case then this should be remove as it's not needed

tests/Rules/data/NoModelForwardingToBuilderInstance.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+/** @var \App\User $user */
+$user = \App\User::find(1);
+$user->first();
+$user->get();
+$user->find();
+$user->paginate();
+$user->where('id', 1);
+$user->take(1);
```
calebdw, line 10: On both of these tests, you need a method that gets forwarded to the QueryBuilder (not the EloquentBuilder), you should be able to use `$user->max('foo')` and `User::max('foo')` for this purpose

calebdw:
These rules should also be documented under `docs/rules.md`

tests/Rules/NoModelForwardingToBuilderTest.php
Smurrlawa: the additional config is actually needed for a successful run.

calebdw: Good to know, on my recent rule test `NoUnnecessaryEnumerableToArrayCallsRule` it was not needed so I just wanted to make sure it wasn't copied over unnecessarily

tests/Rules/NoModelForwardingToBuilderTest.php
```diff
@@ -0,0 +1,37 @@
...
+
+/** @extends RuleTestCase<NoModelForwardingToBuilder> */
+class NoModelForwardingToBuilderTest extends RuleTestCase
+{
+    protected function getRule(): Rule
+    {
+        return new NoModelForwardingToBuilder();
+    }
+
+    public function testRuleDetectsModelForwardingToBuilderInstance(): void
```
calebdw: just a nitpick, but these methods can just be called `testRule` (pairs well with `getRule`)

calebdw (approved):
Looks great, thanks for putting this together!! Definitely looking forward to using this!

Smurrlawa:
Alright I finished the review changes @calebdw.
Feel free to edit the rules.md docs as needed.

Cheers

docs/rules.md
```diff
@@ -559,3 +559,72 @@ parameters:
...
+
+### Examples
+
+Given the following Eloquent model:
+
+```php
+class Post extends Model {}
+```
+
+The following code would result in refetching all posts from the database which is not intended usage:
```
calebdw: For both of these rules, it would be helpful to show what the error message is (see above for an example), you can include the message from the test (with the tip)

docs/rules.md
```diff
@@ -559,3 +559,72 @@ parameters:
...
+```
+
+### Configuration
+
+This rule is disabled by default. To enable it, add the following to your `phpstan.neon` file:
+
+```neon
+parameters:
+    larastan:
+        noModelForwardingToBuilder: true
```
calebdw: I agree that all Larastan options should be under a `larastan:` parameter in the config, but this currently does not exist:

```suggestion
    noModelForwardingToBuilder: true
```

calebdw:
> Feel free to edit the rules.md docs as needed.

I'm not a maintainer here or I would :upside_down_face:

docs/rules.md
```diff
@@ -559,3 +559,74 @@ parameters:
...
+
+The following code would result in refetching all posts from the database which is not intended usage:
+
+```php
+$post = Post::find(1);
+$post->first();
+```
+
+```
+Method [first] is forwarded to a Builder instance, which is not allowed.
```
calebdw, line 507: Please add the tip part as well to help show what needs to be done to fix it
```suggestion
Method [first] is forwarded to a Builder instance, which is not allowed.
    💡 Use [::first()], [::query()->first()] or [->newQuery()->first()] instead.
```

docs/rules.md
```diff
@@ -559,3 +559,74 @@ parameters:
...
+```
+
+To fix this, you should explicitly call `query()` or `newQuery()`:
+
+```php
+Post::query()->first();
+
+// or
+Post::newQuery()->first();
+```
```
calebdw: you should not use `Post::newQuery()` as newQuery is not a static method

Smurrlawa:
@calebdw should I close my PR and you continue on your?

calebdw:
> @calebdw should I close my PR and you continue on your?

No, you should leave this one open :)

calebdw:
I just enabled the first rule in my work repo and fixed some hidden bugs!!

docs/rules.md
```diff
@@ -72,51 +72,16 @@ You can also configure the collection methods which this rule
...
 ```
 will only throw errors on the `count` and `first` methods.
 The inverse is also configurable, to not throw an exception
 on the `contains` method, one could set the following value:
-```neon
-parameters:
-    noUnnecessaryCollectionCallExcept: ['contains']
 ```
-
-## NoUnnecessaryEnumerableToArrayCalls
```
calebdw: Sorry, I didn't mean to confuse you, but the only changes to this file should be the new sections you added in this PR

calebdw:
@Smurrlawa, please see the following fix that I needed on my fork: https://github.com/calebdw/larastan/commit/bd469576769d80c26330979d94d838e66dffb1d3

These changes need to be incorporated into this PR

PR #2324, opened 2025-07-23
[3.x] feat: add NoConstructorDependencyInjectionRule for console commands

ggiak:
## Description

  This PR adds a new rule that enforces Laravel's best practice of using method injection in the `handle()` method rather than constructor injection for console commands.

  ### Background

  Laravel's documentation and community best practices recommend avoiding constructor dependency injection in console commands. Instead, dependencies should be injected into the `handle()`
  method. This pattern:
  - Improves testability
  - Follows Laravel conventions
  - Avoids potential issues with command initialization
  - Makes commands more consistent with the framework

  ### What this rule does

  - Detects console commands extending `Illuminate\Console\Command`
  - Checks if the constructor has any parameters
  - Reports violations with helpful error messages and tips
  - Is disabled by default for backward compatibility

  ### Example

  ```php
  // This will trigger an error:
  class MyCommand extends Command
  {
      public function __construct(private UserService $userService)
      {
          parent::__construct();
      }
  }

  // Error: Console command "MyCommand" should not have constructor arguments. Use dependency injection in the handle() method instead.
```

 ### Configuration

To enable the rule, add to your phpstan.neon:
  parameters:
      noConsoleCommandConstructorInjection: true

- Added or updated tests
- Documented user facing changes

###  Changes

- Added new rule NoConstructorDependencyInjectionRule in src/Rules/ConsoleCommand/
- Added comprehensive test coverage with various scenarios (commands with/without constructors, abstract commands, etc.)
- Added documentation in docs/rules.md explaining the rule, its purpose, and configuration
- Updated extension.neon to register the rule with opt-in configuration parameter

###  Breaking changes

None. The rule is disabled by default to maintain backward compatibility. Users must explicitly enable it in their configuration.

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -559,3 +559,59 @@ parameters:
     checkConfigTypes: true
 ```
 
+## NoConstructorDependencyInjectionRule
+
+This rule ensures that console commands follow Laravel's best practice of using method injection in the `handle()` method rather than constructor injection.
+
+### Examples
+
+Constructor dependency injection:
+```php
+class MyCommand extends Command
+{
+    public function __construct(
+        private UserService $userService
+    ) {
+        parent::__construct();
+    }
+    
+    public function handle(): int
+    {
+        // ...
+    }
+}
+```
+
+This will result in the following error:
+```
+Console command "MyCommand" should not have constructor arguments. Use dependency injection in the handle() method instead.
+```
+
+To fix the error, use method injection in the handle method:
+```php
+class MyCommand extends Command
+{
+    public function handle(UserService $userService): int
+    {
+        // ...
+    }
+}
+```
+
+### Why This Rule Exists
+
+Laravel's documentation and community best practices recommend avoiding constructor dependency injection in console commands. This pattern:
+- Improves testability
+- Follows Laravel conventions
+- Avoids potential issues with command initialization
+- Makes commands more consistent with the framework
+
+### Configuration
+
+This rule is disabled by default. To enable, add the following to your `phpstan.neon` file:
+
+```neon
+parameters:
+    noConsoleCommandConstructorInjection: true
+```
+
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -561,7 +561,7 @@ parameters:
 
 ## NoConstructorDependencyInjectionRule
 
-This rule ensures that console commands follow Laravel's best practice of using method injection in the `handle()` method rather than constructor injection.
+This rule ensures that console commands follow Laravel's best practice of using method injection in the `handle()` or `__invoke()` method rather than constructor injection.
 
 ### Examples
 
@@ -584,14 +584,25 @@ class MyCommand extends Command
 
 This will result in the following error:
 ```
-Console command "MyCommand" should not have constructor arguments. Use dependency injection in the handle() method instead.
+Console command "MyCommand" should not have constructor arguments. Use dependency injection in the handle() or __invoke() method instead.
 ```
 
-To fix the error, use method injection in the handle method:
+To fix the error, use method injection in the handle or __invoke method:
 ```php
 class MyCommand extends Command
 {
     public function handle(UserService $userService): int
+    {
+        // ...
+    }
+}
+```
+
+Or with an invokable command:
+```php
+class MyInvokableCommand extends Command
+{
+    public function __invoke(UserService $userService): int
     {
         // ...
     }
--- a/src/Rules/ConsoleCommand/NoConstructorDependencyInjectionRule.php
+++ b/src/Rules/ConsoleCommand/NoConstructorDependencyInjectionRule.php
@@ -57,7 +57,7 @@ final class NoConstructorDependencyInjectionRule implements Rule
                     ),
                 )->line($methodNode->getLine())
                     ->tip('Move all dependencies to the handle() or __invoke() methods.')
-                    ->identifier('larastan.console.constructorInjection')
+                    ->identifier('larastan.consoleConstructorInjection')
                     ->build(),
             ];
         }
--- a/src/Rules/ConsoleCommand/NoConstructorDependencyInjectionRule.php
+++ b/src/Rules/ConsoleCommand/NoConstructorDependencyInjectionRule.php
@@ -38,7 +38,7 @@ final class NoConstructorDependencyInjectionRule implements Rule
 
         $classReflection = $node->getClassReflection();
 
-        if (! $classReflection->isSubclassOf(Command::class)) {
+        if (! $classReflection->is(Command::class)) {
             return [];
         }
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: src/Rules/ConsoleCommand/NoConstructorDependencyInjectionRule.php [hunk 1]
```diff
@@ -0,0 +1,79 @@
...
+
+use function count;
+use function sprintf;
+
+/** @implements Rule<InClassNode> */
+final class NoConstructorDependencyInjectionRule implements Rule
+{
+    public function getNodeType(): string
+    {
+        return InClassNode::class;
```
canvural: Why don't we hook into `InClassMethod` node? And then we can also do something like
```php
$method = $node->getMethodReflection();

if (!$method->isConstructor()) {
	return [];
}
```

in `processNode` method.

src/Rules/ConsoleCommand/NoConstructorDependencyInjectionRule.php
```diff
@@ -0,0 +1,79 @@
...
+            return [];
+        }
+
+        $params = $constructor->params;
+
+        if (count($params) > 0) {
+            return [
+                RuleErrorBuilder::message(
+                    sprintf(
+                        'Console command "%s" should not have constructor arguments. Use dependency injection in the handle() method instead.',
```
calebdw: Commands can also be invokable (use `__invoke` method instead of `handle`)

szepeviktor: We could be strict and allow only 1 way (`./artisan make:command`) to implement a command.

canvural: No. We can be dynamic here and suggest `__invoke` or `handle` depending which one the class has.

src/Rules/ConsoleCommand/NoConstructorDependencyInjectionRule.php
ggiak: Great point. 

added an additional commit that uses

InClassMethodNode instead.

https://github.com/larastan/larastan/pull/2324/commits/87f9219ed17cd8c5ce673d06d18da0e123e88e34

src/Rules/ConsoleCommand/NoConstructorDependencyInjectionRule.php
ggiak: @calebdw not a great fan of __invoke but you are correct.

@szepeviktor not our job to be opinionated 👍  

@canvural 
As per making the message dynamic, I would suggest to simply change the message instead of identifying the used method. 
Reasoning: People change preferences over time and it's good to raise awareness that someone can use both.

In this spirit, I din't change the logic but I changed the corresponding message to:

1) Console command "%s" should not have constructor arguments. Use dependency injection in the handle() or __invoke() method instead.

2) Tip: "Move all dependencies to the handle() or __invoke() method parameters for better testability and Laravel best practices."

However, I do not have a strong opinion on the matter so feel free to suggest otherwise  and I'd be happy to apply the changes.

docs/rules.md
```diff
@@ -559,3 +559,59 @@ parameters:
     checkConfigTypes: true
 ```
 
+## NoConstructorDependencyInjectionRule
+
+This rule ensures that console commands follow Laravel's best practice of using method injection in the `handle()` method rather than constructor injection.
```
calebdw: docs should be updated to reflect the __invoke method as well

src/Rules/ConsoleCommand/NoConstructorDependencyInjectionRule.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        $method = $node->getMethodReflection();
+
+        if (! $method->isConstructor()) {
+            return [];
+        }
+
+        $classReflection = $node->getClassReflection();
+
+        if (! $classReflection->isSubclassOf(Command::class)) {
```
calebdw: `isSubclassOf` with a class string is deprecated, this should be:


```suggestion
        if (! $classReflection->is(Command::class)) {
```

src/Rules/ConsoleCommand/NoConstructorDependencyInjectionRule.php
```diff
@@ -0,0 +1,67 @@
...
+        if (count($methodNode->params) > 0) {
+            return [
+                RuleErrorBuilder::message(
+                    sprintf(
+                        'Console command "%s" should not have constructor arguments. Use dependency injection in the handle() or __invoke() method instead.',
+                        $classReflection->getName(),
+                    ),
+                )->line($methodNode->getLine())
+                    ->tip('Move all dependencies to the handle() or __invoke() method parameters for better testability and Laravel best practices.')
+                    ->identifier('larastan.console.constructorInjection')
```
calebdw: generally, I've only ever seen two components to the identifier, I would recommend:


```suggestion
                    ->identifier('larastan.consoleConstructorInjection')
```

src/Rules/ConsoleCommand/NoConstructorDependencyInjectionRule.php
```diff
@@ -0,0 +1,67 @@
...
+
+        if (count($methodNode->params) > 0) {
+            return [
+                RuleErrorBuilder::message(
+                    sprintf(
+                        'Console command "%s" should not have constructor arguments. Use dependency injection in the handle() or __invoke() method instead.',
+                        $classReflection->getName(),
+                    ),
+                )->line($methodNode->getLine())
+                    ->tip('Move all dependencies to the handle() or __invoke() method parameters for better testability and Laravel best practices.')
```
calebdw: you're saying the same thing in the message and the tip---I would recommend shortening the actual message. Additionally, I would shorten the tip:


```suggestion
                    sprintf(
                        'Console command "%s" should not have constructor arguments.',
                        $classReflection->getName(),
                    ),
                )->line($methodNode->getLine())
                    ->tip('Move all dependencies to the handle() or __invoke() methods.')
```

calebdw (approved):
Looks pretty good, just a few last nits

src/Rules/ConsoleCommand/NoConstructorDependencyInjectionRule.php
ggiak: there are a bunch:

- src/Rules/DeferrableServiceProviderMissingProvidesRule.php:55
- src/Rules/UselessConstructs/NoUselessValueFunctionCallsRule.php:51

and more, but the most important one is the:
- src/Rules/ConsoleCommand/UndefinedArgumentOrOptionRule.php:78

which resides at the same namespace level as the newly proposed rule.
for the sake of continuety, I would suggest to leave this one out.

wdty ?

calebdw: I would still say those are the exception and not the rule, but it's up to you

ggiak:
> Looks pretty good, just a few last nits

Thank you for the prompt reply / check

Latest changes: https://github.com/larastan/larastan/pull/2324/commits/7cd8b4ed01eccc984964f98069685d4e8764115b

Let me know about: [larastan.console.constructorInjection](https://github.com/search?q=repo%3Alarastan%2Flarastan+identifier%28%27larastan.&type=code)

src/Rules/ConsoleCommand/NoConstructorDependencyInjectionRule.php
ggiak: Then I am performing the change accordingly and I will create a corresponding issue with "help wanted"

ggiak: @calebdw https://github.com/larastan/larastan/pull/2324/commits/5874cac31b32fb54f0aed5dc8ee0f3dbf9f86757

ggiak:
@canvural mind if you check this one? it's an important rule. 

When dependency injection goes to the constructor then the dependencies gets initiailized during the application load.

PR #2334, opened 2025-08-12
[3.x] fix: allow private methods to be used as accessors

calebdw:
Hello!

Rector automatically converts protected methods to private in final classes (as it should). However, this breaks attribute detection.

Thanks!


calebdw:
Actually, it looks like model attributes (and scopes) **have to be protected and cannot be private**---otherwise the parent Model is not able to call the methods

Issue #2255, opened 2025-04-05
Internal error: View


rcerljenko:
- Larastan Version: 3.3.1
- Laravel Version: 12.7.2

### Description

Hi,

Since the latest version we started to receive this internal error:

![Image](https://github.com/user-attachments/assets/97160162-476c-464b-ab3f-92482c0323ee)

I think it has something to do with the fact that we have dedicated error views for some of the common http status codes like so:

![Image](https://github.com/user-attachments/assets/2991d2f4-7d0e-4410-9bf5-987548c7c8b4)

szepeviktor:
Hello 👋🏻

Try
```neon
parameters:
    checkUnusedViews: false
```

rcerljenko:
Hi,

Yes this helps but then we don't have this check at all...

calebdw:
This might be related: https://github.com/larastan/larastan/pull/2218

CC: @AJenbo

AJenbo:
Can someone provide a way to replicate this?

rcerljenko:
Hi @AJenbo ,

You ca replicate it by creating the `errors` dir inside views dir and then inside some blade files named like status codes (the picture that I attached above).

Laravel docs for this feature:

https://laravel.com/docs/12.x/errors#custom-http-error-pages

AJenbo:
I have that in both the project I work on and didn't have an issue

calebdw:
@rcerljenko, we also use custom error view in our project without any issues.

Can you please create a new laravel repo with a minimal reproducible example?

AJenbo:
Arh it's because `checkUnusedViews: true` has to be set in the options.

I don't think it has anything to do with having the error templates though as i get it even when running it on a laravel package that doesn't have any templates.

Here is a stack trace:
```
Uncaught InvalidArgumentException: View [403] not found. in ./vendor/laravel/framework/src/Illuminate/View/FileViewFinder.php:138
#0 ./vendor/laravel/framework/src/Illuminate/View/FileViewFinder.php(78): Illuminate\View\FileViewFinder->findInPaths()
#1 ./vendor/larastan/larastan/src/Rules/UnusedViewsRule.php(78): Illuminate\View\FileViewFinder->find()
#2 phar://./vendor/phpstan/phpstan/phpstan.phar/src/Analyser/AnalyserResultFinalizer.php(51): Larastan\Larastan\Rules\UnusedViewsRule->processNode()
[...]

PR #2337, opened 2025-08-23
[3.x] Only check existing views

ramonrietdijk:
- [x] Added or updated tests

This pull request resolves issue #2255.

**Introduction**

Pull request #2218 has added additional view paths. The `UnusedViewsRule` receives all the views from the `getAllViewNames` method, which now contain views from the vendor folder that may not be accessible directly. Because these views can't be found in the application itself, an `InvalidArgumentException` is thrown.

In a fresh Laravel project, the path `vendor/laravel/framework/src/Illuminate/Foundation/Exceptions/views` and others are added to the directory list. The first file in this folder is `401.blade.php`. As this file does not exist in the application, the exception mentioned above will be thrown which causes issue #2255. Note that the error mentions view `401` and not `errors.401`.

Another side effect of adding multiple paths is that duplicate view names can appear. If the project itself would also contain an unused `401.blade.php` in the root, you'll get the unused file error twice for that file.

**Changes**

I've changed the code to check the existence of every view. This will prevent the rule from checking views that are only available in the vendor directory. Also, an `array_unique` has been wrapped around the unused views to prevent duplicate errors.

The test has also been modified to add a vendor view path.

**Reproduce**

The issue can be reproduced by following these steps:

1. Run `composer create-project laravel/laravel larastan-view-rule`
2. Run `cd larastan-view-rule`
3. Run `composer require --dev "larastan/larastan:^3.0"`
4. Add a `phpstan.neon` file with the following content:

```neon
includes:

parameters:
    paths:
        - app/
        - routes/
    checkUnusedViews: true
    level: 9
```

5. Run `./vendor/bin/phpstan` and an internal error will appear.

**Breaking changes**

### DIFF
--- a/src/Rules/UnusedViewsRule.php
+++ b/src/Rules/UnusedViewsRule.php
@@ -21,6 +21,7 @@ use PHPStan\Rules\RuleErrorBuilder;
 
 use function array_diff;
 use function array_filter;
+use function array_unique;
 use function collect;
 use function iterator_to_array;
 use function view;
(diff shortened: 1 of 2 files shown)


canvural:
Thank you.

PR #2338, opened 2025-08-24
fix: self::query() for final models

calebdw:
Hello!

For some reason, when using `$scope->resolveTypeByName`, PHPStan resolves `self::` as a `ThisType`. However, this was causing final models that returned builders from `self::query()` to fail with:

```
Method App\Mail\Models\Message::prunable() should return Illuminate\Database\Eloquent\Builder<App\Mail\Models\Message> but returns Illuminate\Database\Eloquent\Builder<$this(App\Mail\Models\Message)>.
```

This PR fixes that by downgrading to a static object type.

Thanks!

### DIFF
--- a/src/ReturnTypes/ModelDynamicStaticMethodReturnTypeExtension.php
+++ b/src/ReturnTypes/ModelDynamicStaticMethodReturnTypeExtension.php
@@ -21,6 +21,7 @@ use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Type\DynamicStaticMethodReturnTypeExtension;
 use PHPStan\Type\Generic\GenericObjectType;
 use PHPStan\Type\ObjectType;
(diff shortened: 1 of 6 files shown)


canvural:
> For some reason, when using $scope->resolveTypeByName, PHPStan resolves self:: as a ThisType.

Doesn't that mean it is a PHPstan bug? 🤔

calebdw:
There is no `SelfType`, it looks like Ondre is using ThisType for both purposes (likely to check that private method calls are okay)

I have this an issue opened for over a year: https://github.com/phpstan/phpstan/issues/11557, but I'm not sure what/when the resolution is going to be or even if it's going to happen at all. It's an easy fix on this end for the time being (not the first place we use this check) so that this doesn't error on final models

canvural:
Thank you!

Issue #1822, opened 2024-01-22
Find missing translation strings in blade files


nerg4l:
Hello,

I recently developed a package ([coding-socks/lost-in-translation](https://github.com/coding-socks/lost-in-translation)) which is able to scan Blade files to identify translation strings and report if the translation for that string is missing from a locale.

[u/boreasaurus](https://www.reddit.com/user/boreasaurus/) on Reddit said the following:

> It would be amazing if somehow this functionality existed in larastan, as that is already reporting errors for e.g. view('some.unknown.view')

My package provides a `lost-in-translation:find {locale}` artisan command which reads all blade files, compiles them to PHP, converts them to tokens, and then finds the relevant nodes in the AST (Abstract syntax tree). As far as I know, this is similar to how static analysis tools work.

I'm not sure if I will have enough time in the upcoming days but I would happily work on adding this feature if you think it's a good addition. However, I might require some guidance on where to start because I'm not familiar enough with Larastan/PHPStan.

calebdw:
Awesome!

However, this seems better suited for the [Bladestan](https://github.com/TomasVotruba/bladestan) project which is already analyzing blade files. To the best of my knowledge (and I could be wrong), but Larastan does not report missing views.

nerg4l:
Yes, my bad. The comment on Reddit misled me. I will close this PR and open one there. Thanks for the info.

philbates35:
> To the best of my knowledge (and I could be wrong), but Larastan does not report missing views.

It does actually:

https://github.com/larastan/larastan/blob/2.x/src/Types/ViewStringType.php

However it doesn't currently seem to be working properly, see https://github.com/larastan/larastan/issues/1607

canvural:
Hi,

Yes this is possible with Larastan. I even had a working PR about this: https://github.com/larastan/larastan/pull/1456

Actually I would say this suits to Larastan more than Bladestan 😄 Because Bladestan compiles the blade files to PHP and then does static analysis. To do this task you do not need to compile the blade files. You can check the linked PR on how it works.

If you want, you can try to revive my linked PR and we can add this to Larastan. But it needs to be an optional rule disabled by default.

nerg4l:
I will keep this issue open because it looks like a possible enhancement. Also, I will try to find some time to check the linked PRs and code. If I'm enthusiast enough, I can add it to both library.

nerg4l:
Closed by #2340

PR #1456, opened 2022-11-27
feat: find unused translations

canvural:
wip

PR #2340, opened 2025-08-24
[3.x] Add NoMissingTranslationsRule

ramonrietdijk:
- [x] Added or updated tests

I've added the `NoMissingTranslationsRule` to automatically check if all used translations are available. This pull request has been heavily inspired by #1456.

This pull request resolves issue #1822.

Curious about your thoughts!

**Changes**

By setting the option `checkMissingTranslations` to `true`, the rule will automatically check if all used translations exist.

It will check the following:

* Helper methods `__`, `trans` and `trans_choice`.
* Methods `get` and `choice` on the `Translator` class itself.
* Methods `get` and `choice` on the `Lang` facade.
* All of the above within views, including Blade directives.

**Breaking changes**

No breaking changes have been introduced. The rule is disabled by default.

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -399,7 +399,9 @@ parameters:
 
 ## NoMissingTranslationsRule
 
-This rule will find any untranslated strings in your application. This rule is primarily meant for applications that make use of the dot syntax like `messages.greet`. If you're using translation strings as keys, this rule is unnecessary.
+This rule will find any untranslated strings in your application. This rule is primarily meant for applications that make use of the dot syntax like `messages.greet`. If you're using translation strings as keys, this rule may be unnecessary.
+
+> **NOTE**: If you store your translations in a database, this rule will not be able to detect them. You should leave this rule disabled in such case.
 
 ### Examples
 
@@ -421,6 +423,16 @@ To enable, add the following to your `phpstan.neon` file:
 ```neon
 parameters:
     checkMissingTranslations: true
+```
+
+By default, the path `resources/lang` is scanned. If you have translations elsewhere, make sure to register all the paths.
+
+```neon
+parameters:
+    checkMissingTranslations: true
+    translationDirectories:
+        - resources/lang
+        - resources/translations
 ```
 
 ## NoEnvCallsOutsideOfConfig
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -397,6 +397,32 @@ parameters:
 - `@includeWhen` Blade directive.
 - `@includeFirst` Blade directive.
 
+## NoMissingTranslationsRule
+
+This rule will find any untranslated strings in your application. This rule is primarily meant for applications that make use of the dot syntax like `messages.greet`. If you're using translation strings as keys, this rule is unnecessary.
+
+### Examples
+
+For the following code:
+```php
+__('messages.greet')
+```
+
+Larastan may report the following error:
+```
+Translation "messages.greet" has not been found.
+```
+
+### Configuration
+
+This rule is disabled by default.
+To enable, add the following to your `phpstan.neon` file:
+
+```neon
+parameters:
+    checkMissingTranslations: true
+```
+
 ## NoEnvCallsOutsideOfConfig
 
 Checks for `env` calls outside the `config` directory, which return `null` when the config is cached.
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -399,9 +399,9 @@ parameters:
 
 ## NoMissingTranslationsRule
 
-This rule will find any untranslated strings in your application. This rule is primarily meant for applications that make use of the dot syntax like `messages.greet`. If you're using translation strings as keys, this rule may be unnecessary.
+This rule will find any untranslated strings in your application. It is primarily meant for applications that make use of the dot syntax like `messages.greet`. If you're using translation strings as keys, this rule may be unnecessary. Enabling this rule may decrease performance as it will scan the available views and translations.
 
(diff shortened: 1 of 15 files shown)


szepeviktor (approved):

canvural:
This might make database calls if users have their translations in database, right? That's a deal breaker I'd say 🙁

ramonrietdijk:
Hi @canvural,

Thank you for your response!

I've taken a look at Laravel's [documentation](https://laravel.com/docs/12.x/localization) regarding translations and I can't seem to find anything related to databases. Only arrays and JSON files are supported out of the box. There are packages available like [spatie/laravel-translation-loader](https://github.com/spatie/laravel-translation-loader) in which they can be stored in the database. In that case, I agree that this rule wouldn't help in a static context. However, in the perspective of Laravel, I'd say that the rule can definitely be beneficial. Users that make use of translation packages that interact with the database can leave the rule disabled, as it is by default.

Happy to hear your thoughts!

canvural:
> Users that make use of translation packages that interact with the database can leave the rule disabled, as it is by default.

This is correct, but still users will blame Larastan 😄 

So here is my proposal, we can keep the logic like how it was in the PR you mentioned. We can scan the `lang` directory (or any directory the user specified via config) and find the translations that way and compare it with the found translations. Then we can make a disclaimer that the rule will not work if you load your translations via database which is reasonable.

What do you think?

ramonrietdijk:
Hi @canvural,

Thanks for sharing your perspective, I can definitely understand why we should be careful with this and agree with you.

I'll update the pull request once I've got the opportunity to do so!

ramonrietdijk:
Hi @canvural,

I've modified the code so available translations are now retrieved manually. The used paths can also be configured. At last, a disclaimer has also been added to the documentation.

canvural:
Thank you. I'll review it in the coming couple days.

docs/rules.md
```diff
@@ -397,6 +397,44 @@ parameters:
 - `@includeWhen` Blade directive.
 - `@includeFirst` Blade directive.
 
+## NoMissingTranslationsRule
+
+This rule will find any untranslated strings in your application. This rule is primarily meant for applications that make use of the dot syntax like `messages.greet`. If you're using translation strings as keys, this rule may be unnecessary.
+
+> **NOTE**: If you store your translations in a database, this rule will not be able to detect them. You should leave this rule disabled in such case.
```
canvural:
```suggestion
> **NOTE**: If you store your translations in a database, this rule will not be able to detect them. You should leave this rule disabled in such cases.
```

Also maybe we can include a disclaimer about the performance. Because I think this rule might be heavier side on the processing. We are parsing view files, loading all the translations into the memory etc.

src/Rules/NoMissingTranslationsRule.php
```diff
@@ -0,0 +1,118 @@
...
+use function array_merge;
+use function in_array;
+use function is_dir;
+use function json_decode;
+use function lang_path;
+
+/** @implements Rule<CollectedDataNode> */
+final class NoMissingTranslationsRule implements Rule
+{
+    use HasContainer;
```
canvural: I don't see this trait being used 🤔

src/Collectors/UsedTranslationViewCollector.php
```diff
@@ -0,0 +1,74 @@
...
+
+        foreach ($this->viewFileHelper->getRootViewFilePaths() as $viewFile) {
+            try {
+                $parserNodes = $this->parser->parseFile($viewFile);
+
+                $translations[$viewFile] = $this->processNodes($parserNodes);
+            } catch (ParserErrorsException) {
+                continue;
+            }
+        }
```
canvural: I know we do the similar thing in `UsedViewInAnotherViewCollector`

Could you maybe look into if we can cache the parsed views somehow? Maybe a new class with a property inside? Or a static variable, though I'd rather not.

canvural (changes requested):
Looks good. Thank you. Just left some remarks about the performance.

src/Rules/NoMissingTranslationsRule.php
ramonrietdijk: Correct, this seems to have been a leftover from the previous implementation

ramonrietdijk:
Thanks for reviewing my pull request, I've learned quite a lot about PHPStan and Larastan!

I've resolved all of your feedback. Now, all parsed views will be kept in the `ViewParser` class. As it is registered as a service, both `UsedTranslationViewCollector` and `UsedViewInAnotherViewCollector` will make use of the same instance. Files that have already been processed can be used directly, improving performance.

canvural:
Looks good! Let's try this.

src/Collectors/UsedTranslationTranslatorCollector.php
```diff
@@ -0,0 +1,59 @@
...
+
+        $class = $node->var;
+
+        if (! (new ObjectType(Translator::class))->isSuperTypeOf($scope->getType($class))->yes()) {
+            return null;
+        }
+
+        $template = $node->getArgs()[0]->value;
+
+        if (! $template instanceof Node\Scalar\String_) {
```
jvandreae, line 53: If you use `getConstantStrings` it will have better type inference support, e.g. https://github.com/jbboehr/phpstan-lost-in-translation/blob/5cb24e9b205d8fd7f8482a61160758876710e8a6/src/LostInTranslationHelper.php#L225

It does mean there may be more than one value in the case of e.g. "foo"|"bar"

PR #2343, opened 2025-08-26
[3.x] Use explicit nullable types

ramonrietdijk:
After making changes to the stub files, running Larastan for the first time on PHP 8.4 causes deprecation notices.

```
Deprecated in PHP 8.4: Parameter #1 $get (callable) is implicitly nullable via default value null.
```

This pull request has been made to resolve them.

**Changes**

I've changed the implicitly nullable parameters to be explicitly nullable. Note that the types themselves have already been updated in Laravel.

**Breaking changes**

No breaking changes have been introduced. Marking parameters explicitly as nullable won't break existing code.

### DIFF
--- a/stubs/common/Attribute.stub
+++ b/stubs/common/Attribute.stub
@@ -31,7 +31,7 @@ class Attribute
      * @param  (callable(TMakeSet, array<string, mixed>): mixed)|null  $set
      * @return Attribute<TMakeGet, TMakeSet>
      */
-    public static function make(callable $get = null, callable $set = null);
+    public static function make(?callable $get = null, ?callable $set = null);
 
     /**
      * Create a new attribute accessor.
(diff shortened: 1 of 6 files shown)


calebdw:
Thanks for this!

> Note that the types themselves have already been updated in Laravel.

And this is why it would be great to push more of these stubs into the framework, less we have to maintain and keep in sync

szepeviktor (approved):

canvural:
Thank you.

---

> And this is why it would be great to push more of these stubs into the framework, less we have to maintain and keep in sync

This is partially true. Stubs are still the best way to add new types, also they have the added bonus to support multiple Laravel version. If it goes into framework only that version and above will have the types.

I have plans to automate the syncing of the stubs of some sort. Will see if it can work.

PR #2346, opened 2025-08-30
feat: add support for Collection, Builder, and Arr pluck

calebdw:
Hello!

This adds `pluck` support to the Collection, Builder, and Arr classes:

```php
// before
User::pluck('name', 'id');        // Collection<(int|string), mixed>
User::all()->pluck('name', 'id'); // Collection<(int|string), mixed>
Arr::pluck($users, 'name', 'id'); // array<int, User> INCORRECT!

// after
User::pluck('name', 'id');        // Collection<int, string>
User::all()->pluck('name', 'id'); // Collection<int, string>
Arr::pluck($users, 'name', 'id'); // array<int, string>
```

Thanks!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     },
     "require-dev": {
         "doctrine/coding-standard": "^13",
-        "nikic/php-parser": "^5.4",
+        "nikic/php-parser": "^5.6",
         "laravel/framework": "^11.44.2 || ^12.7.2",
         "mockery/mockery": "^1.6.12",
         "orchestra/canvas": "^v9.2.2 || ^10.0.1",
--- a/src/ReturnTypes/ArrPluckExtension.php
+++ b/src/ReturnTypes/ArrPluckExtension.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Larastan\Larastan\ReturnTypes;
+
+use Illuminate\Support\Arr;
+use Larastan\Larastan\Support\PluckHelper;
+use PhpParser\Node\Expr\StaticCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Type\DynamicStaticMethodReturnTypeExtension;
+use PHPStan\Type\Type;
+
+final class ArrPluckExtension implements DynamicStaticMethodReturnTypeExtension
+{
+    public function __construct(private PluckHelper $pluckHelper)
+    {
+    }
+
+    public function getClass(): string
+    {
+        return Arr::class;
+    }
+
+    public function isStaticMethodSupported(MethodReflection $methodReflection): bool
+    {
+        return $methodReflection->getName() === 'pluck';
+    }
+
+    public function getTypeFromStaticMethodCall(
+        MethodReflection $methodReflection,
+        StaticCall $methodCall,
+        Scope $scope,
+    ): Type|null {
+        $arrayArg = $methodCall->getArg('array', 0);
+        $valueArg = $methodCall->getArg('value', 1);
+        $keyArg   = $methodCall->getArg('key', 2);
+
+        if ($arrayArg === null || $valueArg === null) {
+            return null;
+        }
+
+        $from = $scope->getType($arrayArg->value)->getIterableValueType();
+
+        return $this->pluckHelper->getArrayType($from, $valueArg, $keyArg, $scope);
+    }
+}
--- a/src/ReturnTypes/EloquentBuilderPluckExtension.php
+++ b/src/ReturnTypes/EloquentBuilderPluckExtension.php
@@ -1,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Larastan\Larastan\ReturnTypes;
+
+use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
+use Larastan\Larastan\Support\PluckHelper;
+use PhpParser\Node\Expr\MethodCall;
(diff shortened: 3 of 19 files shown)


canvural:
Thank you. I'll review it in the coming couple days.

src/ReturnTypes/EloquentBuilderPluckExtension.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        return $methodReflection->getName() === 'pluck';
+    }
+
+    public function getTypeFromMethodCall(
+        MethodReflection $methodReflection,
+        MethodCall $methodCall,
+        Scope $scope,
+    ): Type|null {
+        $valueArg = $methodCall->getArg('value', 0);
```
canvural: Isn't it called `column` ?

src/ReturnTypes/ArrPluckExtension.php
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public function getTypeFromStaticMethodCall(
+        MethodReflection $methodReflection,
+        StaticCall $methodCall,
+        Scope $scope,
+    ): Type|null {
+        $arrayArg = $methodCall->getArg('array', 0);
+        $valueArg = $methodCall->getArg('value', 1);
+        $keyArg   = $methodCall->getArg('key', 2);
```
canvural: PHPStan already normalizes the arguments for us. So we can keep using `$methodCall->getArgs()[0]` etc. Is there any other advantage of using this `getArg` method with name and position?

tests/Type/data/arr-pluck-l12-20.php
```diff
@@ -0,0 +1,52 @@
...
+ * @param  array<int, array{id: int, name: string}>  $array
+ * @param  iterable<array{user: array{id: int, name: string}}>  $iterable
+ * @param  list<array{user: User}>  $list
+ */
+function test(
+    array $array,
+    iterable $iterable,
+    array $list,
+): void {
+    assertType('array<int, string>', Arr::pluck($array, 'name'));
```
canvural: I'm adding my comment here but should be applicable to others too.

Shouldn't it be `list<string>` here?

tests/Type/data/eloquent-builder.php
canvural, line 1: Can't understand why this file has changes 🤔

tests/Type/GeneralTypeTest.php
```diff
@@ -42,7 +43,10 @@ public static function dataFileAsserts(): iterable
             yield from self::gatherAssertTypes(__DIR__ . '/data/date-extension-l11.php');
         }
 
+        yield from self::gatherAssertTypes(__DIR__ . '/data/eloquent-builder.php');
```
canvural: This file is already included, we don't need it here.

composer.json
```diff
@@ -21,11 +21,11 @@
         "illuminate/http": "^11.44.2 || ^12.4.1",
         "illuminate/pipeline": "^11.44.2 || ^12.4.1",
         "illuminate/support": "^11.44.2 || ^12.4.1",
-        "phpstan/phpstan": "^2.1.11"
+        "phpstan/phpstan": "^2.1.21"
```
canvural: Why do we need the bump?

canvural (changes requested):
Thanks for the PR! I left some comments for the changes, but I also have some comments in general:

The pluck method is very similar to PHP's built-in `array_column` which PHPStan has support for. So what I was thinking is that we could get advantage of that. We could do something like
```php
$scope->getType(
    new FuncCall(
        new Name('array_column'),
        $args
    )
)
```

and it already handles lots of cases. For example, the `list<>` result type. If we need more control maybe we could even directly use the `ArrayColumnHelper` (would first need to add `@api`)

The only cases where we need to custom support would be:
- if the column name contains `.`, thus nested access
- Closure

but I think those are doable too.

Apart from that, not every case needs the `PluckHelper` class. For example, the `EloquentBuilderPluckExtension` could be simplified and it could use the property fetch directly in the extension, something like:
```php
return new GenericObjectType(
    Collection::class,
    [
        $keyType,
        $from->getProperty($scope->getType($valueArg->value)->getConstantStrings()[0]->getValue(), $scope)->getReadableType()
    ]
);
```

Overall I think by taking advantage of what's already there (PHPStan code in this case) we'd greatly reduce the code complexity here and benefit from future updates.

What do you think?

src/ReturnTypes/EloquentBuilderPluckExtension.php
calebdw: Ah yeah, good catch!

src/ReturnTypes/ArrPluckExtension.php
calebdw: > PHPStan already normalizes the arguments for us

Good to know, I was not aware of this.

>  Is there any other advantage of using this getArg method with name and position?

Yes, it's a new method on the CallLike, it's more clear exactly what arg we are targeting, and it returns `null` for first-class callables and unpacks so we don't have to check for that.

composer.json
calebdw: This was bumped on 3.x so no longer relevant

tests/Type/data/eloquent-builder.php
calebdw: Ah, that stray line in the GeneralTypeTest was causing all these assertions to fail and I couldn't figure out why :sweat_smile:, reverted :+1:

FeBe95:
Closed? 👀

calebdw:
Hey there! Not closed, just moved---see below :)

https://x.com/calebdw/status/2091781493127270889

PR #2351, opened 2025-09-03
fix: allow null values in trans replacements

oprypkhantc:
- [ ] Added or updated tests

**Changes**

Make `null` an allowed value for the "replacements" array in `trans()` and `__()` helpers. 

This is supported by Laravel intentionally, judging by this PR, which explicitly added handling for `null` values when migrating to a newer PHP version: https://github.com/laravel/framework/pull/42216

**Breaking changes**

None, completely backwards compatible.


canvural:
Thank you!

PR #2356, opened 2025-09-08
fix: Correct generics forCollection::chunk() method in stubs

khaled-sadek:
**Description**
This PR fixes a false positive type error reported by Larastan when using the `chunk()->each()` pattern on an Eloquent collection.

**1. The Problem:**
When analyzing code that chunks a collection of Eloquent models and then iterates over each chunk, Larastan incorrectly reports a parameter type mismatch.

```
$all = ExampleModel::get(); // \Illuminate\Database\Eloquent\Collection<int, ExampleModel>
$all->chunk(10)->each(fn(\Illuminate\Support\Collection $chunk) => $this->processBatch($chunk));
```
**Error Output:**
```
Parameter #1 $callback of method Illuminate\Support\Collection<(int|string), Illuminate\Database\Eloquent\Model>::each() expects
         callable(Illuminate\Database\Eloquent\Model, int|string): mixed, Closure(Illuminate\Support\Collection): void given.
```
**2. Root Cause:**
The issue lies in the chunk method is missed in the `Collection.stub` file.

**3. The Solution:**
This PR add the chunk method in `collection.stub` to more accurately represent its return type using nested generics.


stubs/common/Collection.stub
```diff
@@ -76,6 +76,15 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
      * @return static<($keyBy is string ? array-key : ($keyBy is array ? array-key : TNewKey)), TValue>
      */
     public function keyBy($keyBy);
+
+    /**
+     * Chunk the collection into chunks of the given size.
+     *
+     * @param  int  $size
+     * @param  bool  $preserveKeys
+     * @return $preserveKeys is true ? static<int, static<TKey, TValue>> : static<int, static<int, TValue>>
```
szepeviktor: (Maybe) (you) (need)
https://phpstan.org/blog/phpstan-1-6-0-with-conditional-return-types
```suggestion
     * @return ($preserveKeys is true ? static<int, static<TKey, TValue>> : static<int, static<int, TValue>>)
```

khaled-sadek: thank you.
it works now

khaled-sadek: Tests passed in Laravel 12. but fails in Laravel 11.

PR #2365, opened 2025-09-11
Add autowired to Filesystem in extension.neon

### LINKED ISSUE #2361: Error on 3.7.0
<!-- First, thank you for reporting a bug. That takes time and we appreciate that! -->

- Larastan Version: 3.7.0
- Laravel Version: 12.28.1

### Description
Upgrading to 3.7.0 gives this error

```
> vendor/bin/phpstan analyze --error-format=blade --memory-limit=-1
Note: Using configuration file /home/paul/www/nl-rdo-zvs-web-private-develop/phpstan.neon.

In Resolver.php line 477:
                                                                                                                                                                                                                         
  Service of type Larastan\Larastan\Rules\NoMissingTranslationsRule: Multiple services of type Illuminate\Filesystem\Filesystem found: 0824, 0844 (required by $filesystem in NoMissingTranslationsRule::__construct())  
                                                                                                                                                                                                                         

In Autowiring.php line 53:
                                                                                
  Multiple services of type Illuminate\Filesystem\Filesystem found: 0824, 0844  
                                                                                

analyse [-c|--configuration CONFIGURATION] [-l|--level LEVEL] [--no-progress] [--debug] [-a|--autoload-file AUTOLOAD-FILE] [--error-format ERROR-FORMAT] [-b|--generate-baseline [GENERATE-BASELINE]] [--allow-empty-baseline] [--memory-limit MEMORY-LIMIT] [--xdebug] [--tmp-file TMP-FILE] [--instead-of INSTEAD-OF] [--fix] [--watch] [--pro] [--fail-without-result-cache] [--] [<paths>...]                                                                                                                                                                                 

Script vendor/bin/phpstan analyze --error-format=blade --memory-limit=-1 handling the phpstan event returned with error code 1
```

ramonrietdijk:
Resolves #2361

Also see issue https://github.com/bladestan/bladestan/issues/166

**Introduction**

In version [v3.7.0](https://github.com/larastan/larastan/releases/tag/v3.7.0) the service `Illuminate\Filesystem\Filesystem` has been added to the configuration file. Additional extensions like [bladestan](https://github.com/bladestan/bladestan) have already registered this class, causing issues with dependency injection.

This process is documented on [doc.nette.org](https://doc.nette.org/en/dependency-injection/autowiring):

> To be able to use autowiring, there must be exactly one service of each type in the container. If there were more, autowiring wouldn't know which one to pass and would throw an exception

To solve this, a preference can be set using the `autowired` attribute. Note that multiple `autowired` attributes will cause the same issue.

Alternatively, users may configure their own `phpstan.neon` by adding the following content:

```neon
...

services:
    -
        class: Illuminate\Filesystem\Filesystem
        autowired: self
```

This, of course, may not be desirable.

**Changes**

I've added `autowired: self` to the `Illuminate\Filesystem\Filesystem` service configuration.

I'm not quite sure if this should be added to Larastan as it works out of the box. Happy to hear your thoughts!

**Breaking changes**

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -668,6 +668,7 @@ services:
 
     -
         class: Illuminate\Filesystem\Filesystem
+        autowired: self
 
 rules:
     - Larastan\Larastan\Rules\UselessConstructs\NoUselessWithFunctionCallsRule


AJenbo:
This seems like the right way to solve this as it avoids tightly coupling between specific versions for no real reason.

ramonrietdijk:
Agreed! However, I do prefer #2366 as adding `autowired: self` on a generic class should maybe not be part of Larastan. Creating a separate class will ensure it's not affected by the configuration of other extensions.

canvural:
I think I prefer this one because adding an empty class just to get around of the issue seems wrong.

Ideally in another PR we can get rid of the `Illuminate\Filesystem\Filesystem` dependency all together. We already use Symfony's `Finder` in other cases, we can refactor to use that here too.

Thank you!

PR #2366, opened 2025-09-11
[3.x] Add Larastan Filesystem

### LINKED ISSUE #2361: Error on 3.7.0

ramonrietdijk:
Resolves #2361

Also see issue https://github.com/bladestan/bladestan/issues/166

This is an alternative approach to https://github.com/larastan/larastan/pull/2365

**Introduction**

In version [v3.7.0](https://github.com/larastan/larastan/releases/tag/v3.7.0) the service `Illuminate\Filesystem\Filesystem` has been added to the configuration file. Additional extensions like [bladestan](https://github.com/bladestan/bladestan) have already registered this class, causing issues with dependency injection.

This process is documented on [doc.nette.org](https://doc.nette.org/en/dependency-injection/autowiring):

> To be able to use autowiring, there must be exactly one service of each type in the container. If there were more, autowiring wouldn't know which one to pass and would throw an exception

By adding a separate Larastan class, the extension won't be affected by other extensions.

**Changes**

I've added a `Filesystem` class that extends `Illuminate\Filesystem\Filesystem` and will only be autowired to itself and not the parent class(es).

**Breaking changes**

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -667,7 +667,8 @@ services:
         class: Larastan\Larastan\Rules\ConfigCollectionRule
 
     -
-        class: Illuminate\Filesystem\Filesystem
+        class: Larastan\Larastan\Support\Filesystem
+        autowired: self
 
 rules:
     - Larastan\Larastan\Rules\UselessConstructs\NoUselessWithFunctionCallsRule
--- a/src/Support/Filesystem.php
+++ b/src/Support/Filesystem.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Larastan\Larastan\Support;
+
+use Illuminate\Filesystem\Filesystem as BaseFilesystem;
+
+final class Filesystem extends BaseFilesystem
+{
+}
--- a/src/Support/Filesystem.php
+++ b/src/Support/Filesystem.php
@@ -6,6 +6,7 @@ namespace Larastan\Larastan\Support;
 
 use Illuminate\Filesystem\Filesystem as BaseFilesystem;
 
+/** @internal */
 final class Filesystem extends BaseFilesystem
 {
 }
--- a/src/Rules/NoMissingTranslationsRule.php
+++ b/src/Rules/NoMissingTranslationsRule.php
@@ -4,13 +4,13 @@ declare(strict_types=1);
 
 namespace Larastan\Larastan\Rules;
 
-use Illuminate\Filesystem\Filesystem;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Larastan\Larastan\Collectors\UsedTranslationFacadeCollector;
 use Larastan\Larastan\Collectors\UsedTranslationFunctionCollector;
 use Larastan\Larastan\Collectors\UsedTranslationTranslatorCollector;
 use Larastan\Larastan\Collectors\UsedTranslationViewCollector;
+use Larastan\Larastan\Support\Filesystem;
 use PhpParser\Node;
 use PHPStan\Analyser\Scope;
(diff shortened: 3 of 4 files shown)


extension.neon
```diff
@@ -667,7 +667,8 @@ services:
         class: Larastan\Larastan\Rules\ConfigCollectionRule
 
     -
-        class: Illuminate\Filesystem\Filesystem
+        class: Larastan\Larastan\Support\Filesystem
+        autowired: self
```
szepeviktor, line 671: then you don't need autowire
```suggestion
        class: Larastan\Larastan\Support\Filesystem
```

src/Support/Filesystem.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Larastan\Larastan\Support;
+
+use Illuminate\Filesystem\Filesystem as BaseFilesystem;
+
+final class Filesystem extends BaseFilesystem
```
calebdw, line 10:
```suggestion
/** @internal */
final class Filesystem extends BaseFilesystem
```

extension.neon
calebdw: I'm pretty sure you can just add the `#[Autowired]` phpstan attribute to the class instead of adding it to the extension.neon

ramonrietdijk:
Hi @szepeviktor & @calebdw,

Thanks for your reviews!

Although `Larastan\Larastan\Support\Filesystem` is a new class, it is detected as a child of `Illuminate\Filesystem\Filesystem`. Leaving out the `autowired: self` will in that case still cause the same issue. By adding it, it will only be used for the class itself.

I also couldn't find anything about the mentioned `#[Autowired]` attribute which is why it's still configured in the `extension.neon` file.

calebdw:
image

ramonrietdijk:
Hmm, I can't seem to make it work using the `AutowiredService` attribute. Adding it doesn't seem to solve the issue. Solely relying on the attribute causes the `Larastan\Larastan\Support\Filesystem` class not to be found.

canvural:
Closing in favor of #2365

PR #2369, opened 2025-09-11
feat: new collection-of custom PHPDoc type

canvural:
- [x] Added or updated tests

**Changes**

**Breaking changes**


calebdw:
This looks useful! It might be worth having a similar `builder-of` type

canvural:
> This looks useful! It might be worth having a similar `builder-of` type

Yeah, that was gonna be the follow up. I could accept a PR that adds it, once I'm happy with this one. Finding some edge cases right now.

SlyDave:
I'm likely being dumb, but how does this differ to `@return Collection<int, Post>` or `@return PostCollection<int, Post>` ?

calebdw:
That works when you know that the model is a Post, but what about when you're not sure? This allows PHPStan to know about dynamic custom builders/collections

PR #2370, opened 2025-09-12
fix: request->server() returns float|int|list<string>|string

calebdw:
Hello!

Turns out that `$request->server()` can return ints, floats, or even a list of strings (argv) when running in cli

Thanks!


szepeviktor:
Are you sure?
https://github.com/laravel/framework/blob/9f375d63c4e965f11c26bff2821156c58b4b0506/src/Illuminate/Http/Concerns/InteractsWithInput.php#L17-L24

calebdw:
100%

canvural:
Thank you!

PR #2371, opened 2025-09-12
[3.x] Make regex for UsedTranslationViewCollector lazy

### LINKED ISSUE #2367: `NoMissingTranslationsRule` fails to match some translations in blade files
- Larastan Version: 3.7.1
- Laravel Version: 11.46.0

### Description

> This is a follow-up issue to #2362 (@ramonrietdijk).

After updating Larastan to version `3.7.1` today, I am happy to report that most of the false-positive translation errors are now gone. Unfortunately, about 400 of the 4,000+ errors remain. All errors (except a single one, which I will create a separate issue for) are coming from blade files. Here's an example:

```md
 ------ --------------------------------------------------------------------------------------------------------- 
  Line   resources\views\auth\passwords\reset.blade.php                                                           
 ------ --------------------------------------------------------------------------------------------------------- 
  0      Translation "auth/passwords/reset.e_mail_address')" value="{{ $email ?? old('email" has not been found.  
         🪪  larastan.missingTranslations                                                                         
         ✏️  C:\xampp\htdocs\bidX\resources\views\auth\passwords\reset.blade.php:0                                
 ------ --------------------------------------------------------------------------------------------------------- 
```

Again, I did some step debugging using XDebug. This time, however, there is a matching item in `$availableTranslations`:

Image

At least it seems that it _should_ match, but if you look closely at the error message, it seems that the extracted translation key is incorrect:
```
auth/passwords/reset.e_mail_address')" value="{{ $email ?? old('email
```

This is the case for all reported blade file errors. The translation key matcher never finds the "correct" ending quote. I guess the regular expression is lacking some fine-tuning:

https://github.com/larastan/larastan/blob/2e653fd19585a825e283b42f38378b21ae481cc7/src/Collectors/UsedTranslationViewCollector.php#L22

#### More examples:
```
- Translation "tool/activation.closing', ['appName' => Config::get('app.name" has not been found.
- Translation "tool/advertising-report.calendar_week')" + ("0" + startDate.getWeekNumber()).substr(-2) + "-" + sheetsData.index.period[0].replace(/[-]/g, '') + '-' + sheetsData.index.period[1].replace(/[-]/g, '" has not been found.
- Translation "tool/discover.discover_head_tooltip_Impr')">@lang('tool/discover.discover_table_header_15" has not been found.
```

```blade
<input id="email" type="email" class="form-control" name="email" placeholder="@lang('auth/passwords/reset.e_mail_address')" value="{{ $email ?? old('email') }}" required autofocus autocomplete="username">
```

```blade
<p>@lang('tool/activation.closing', ['appName' => Config::get('app.name')])</p>
```

```blade
<script>
filename = start[0].substring(0, 4) + "-" + "@lang('tool/advertising-report.calendar_week')" + ("0" + startDate.getWeekNumber()).substr(-2) + "-" + sheetsData.index.period[0].replace(/[-]/g, '') + '-' + sheetsData.index.period[1].replace(/[-]/g, '') + brandName + countryCode + " Advertising Report.pdf";
</script>
```

```blade
<th class="impressions concealable" style="width: 6%;" data-db="impressions"><span data-toggle="tooltip" title="@lang('tool/discover.discover_head_tooltip_Impr')">@lang('tool/discover.discover_table_header_15')<span class="unit">#</span></span></th>
```

ramonrietdijk:
Resolves #2367

**Changes**

I've changed `*` to `*?` in order to make the regex *lazy*. By doing so, the pattern will match as few characters as possible, stopping at the correct quote.

> **Note**: The linked regex101.com example does not seem to show the issue.

**Example**

```html
<input placeholder="@lang('a.b')" value="{{ old('email') }}">
Using *                    ---------------------------
Using *?                   ---
```

**Breaking changes**

### DIFF
--- a/src/Collectors/UsedTranslationViewCollector.php
+++ b/src/Collectors/UsedTranslationViewCollector.php
@@ -18,8 +18,8 @@ use const PREG_SET_ORDER;
 
 final class UsedTranslationViewCollector
 {
-    /** @see https://regex101.com/r/ksUgGz/1 */
-    private const TRANSLATION_REGEX = '/[^\w](trans|trans_choice|Lang::get|Lang::choice|Lang::trans|Lang::transChoice|@lang|@choice|__|\$t)\((?P<quote>[\'"])(?P<string>(?:\\k{quote}|(?!\k{quote}).)*)\k{quote}[\),]/m';
+    /** @see https://regex101.com/r/akPfPj/1 */
+    private const TRANSLATION_REGEX = '/[^\w](trans|trans_choice|Lang::get|Lang::choice|Lang::trans|Lang::transChoice|@lang|@choice|__|\$t)\((?P<quote>[\'"])(?P<string>(?:\\k{quote}|(?!\k{quote}).)*?)\k{quote}[\),]/m';
 
     public function __construct(private ViewParser $viewParser, private ViewFileHelper $viewFileHelper)
     {


FeBe95:
Awesome!

Side note: Translation keys that include variables are not supported by this regex, I guess?

E.g.:
```blade
@lang('foo.' . $bar . 'baz')
```

ramonrietdijk:
Correct, from the point of static analysis the values of `$bar` would only be known on run time.

FeBe95:
And with your regex, would it try to match `'foo'` alone in the example above?

ramonrietdijk:
Nope, it would ignore the entire tag as it is deemed an invalid translation.

src/Collectors/UsedTranslationViewCollector.php
```diff
@@ -18,8 +18,8 @@
 
 final class UsedTranslationViewCollector
 {
-    /** @see https://regex101.com/r/ksUgGz/1 */
-    private const TRANSLATION_REGEX = '/[^\w](trans|trans_choice|Lang::get|Lang::choice|Lang::trans|Lang::transChoice|@lang|@choice|__|\$t)\((?P<quote>[\'"])(?P<string>(?:\\k{quote}|(?!\k{quote}).)*)\k{quote}[\),]/m';
+    /** @see https://regex101.com/r/akPfPj/1 */
```
canvural, line 21: Thanks for updating the link too!

canvural:
Thank you!

PR #2372, opened 2025-09-12
[3.x] Ignore vendor translations

### LINKED ISSUE #2360: NoMissingTranslationsRule scans vendor views
- Larastan Version: 3.7.0
- Laravel Version: 12.28.1

### Description

Using the latest `NoMissingTranslationsRule` rule it reports missing translations from `resources/views/vendor` folder.

Example:
Image

ramonrietdijk:
- [x] Added or updated tests

Resolves  #2360

**Introduction**

All translations containing a vendor prefix like `vendor::key` will fail. This is due to the loading of the available translations reflecting file paths. The *namespace* or *hint* is not taken in account. Adding support for them seems quite the challenge.

The folder name does not necessarily correspond to the *namespace*. We wouldn't be able to know how it's called from a static analysis point of view.

```php
$this->loadTranslationsFrom('...', 'This Is My Namespace!');

__('This Is My Namespace!::messages.greeting');
```

Next up, as stated on the [documentation](https://laravel.com/docs/12.x/localization#overriding-package-language-files) of Laravel, the files may only contain overrides.

> Within this file, you should only define the translation strings you wish to override.

This means that we must merge the original translations from the package vendor as well.

All things considered, it would take quite some effort to implement vendor prefixes. Either automatically or introduce additional configuration to specify the used namespaces. I think scoping the rule for local translations would be the best option.

Wondering what you think!

**Changes**

Translations that include `::` will be ignored.

**Breaking changes**

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -401,6 +401,8 @@ parameters:
 
 This rule will find any untranslated strings in your application. It is primarily meant for applications that make use of the dot syntax like `messages.greet`. If you're using translation strings as keys, this rule may be unnecessary. Enabling this rule may decrease performance as it will scan the available views and translations.
 
+Translations from vendors like `vendor::key` will not be checked.
+
 > **NOTE**: If you store your translations in a database, this rule will not be able to detect them. You should leave this rule disabled in such cases.
 
 ### Examples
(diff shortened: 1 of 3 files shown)


canvural:
Yeah, makes sense to ignore these. 

Thank you.

PR #2373, opened 2025-09-13
[3.x] Add support for translation groups

### LINKED ISSUE #2368: `NoMissingTranslationsRule` should support translation arrays
# Feature Request

> This is a follow-up issue to #2362 and #2367.
> I guess this is also for @ramonrietdijk, but I might be able to assist, if needed.

I am getting the following error:
```md
 ------ -------------------------------------------------------------------- 
  Line   app\Models\AmazonCorrespondingFocus.php                             
 ------ -------------------------------------------------------------------- 
  70     Translation "tool/structure" has not been found.                    
         🪪  larastan.missingTranslations                                    
         ✏️  C:\xampp\htdocs\bidX\app\Models\AmazonCorrespondingFocus.php:70  
 ------ -------------------------------------------------------------------- 
```

If I look into the model, I see the following code:
```php
foreach (__('tool/structure') as $key => $trans) {
    // ...
}
```

The `__()` (or `trans()`) method can also return an array of translations instead of a single translation value:
- https://github.com/laravel/framework/blob/7b5e185e500093841c3f42ac943067131226331a/src/Illuminate/Foundation/helpers.php#L991
    ```php
    function trans($key = null, $replace = [], $locale = null): Translator|array|string 
    ```

**Additional context**

I am not 100% sure, but the fix for this should be relatively easy: When building the `$availableTranslations` array, you might also want to create a `$availableTranslationArray` array, which contains all the "parent" keys that will resolve to an array when requested. Or even easier: just add them to the `$availableTranslations` array directly. You will lose out on the distinction then obviously, which might be used to resolve the return type of the functions (`string` vs `array`).

Here's an example:

`resources/lang/en/tool/structure.php`

Content:
```php
return [
    'category' => [
        'device_type' => [
            'mobile' => 'Mobile',
            'desktop' => 'Desktop',
            'tablet' => 'Tablet',
            'unknown' => 'Unknown',
        ],
        // ...
    ],
    // ...
];
```

Possible translation strings:
```php
$availableTranslations = [
    'tool/structure.category.device_type.mobile',
    'tool/structure.category.device_type.desktop',
    'tool/structure.category.device_type.tablet',
    'tool/structure.category.device_type.unknown',
    // ...
];
```

Possible translation arrays:
```php
$availableTranslationArrays = [
    'tool/structure',
    'tool/structure.category.',
    'tool/structure.category.device_type',
    // ...
]
```

ramonrietdijk:
- [x] Added or updated tests

Resolves #2368

**Introduction**

In the `NoMissingTranslationsRule` class we make use of the `Arr::dot` method to transform the translations into a single array. As the array is flattened, the keys of translations that contain an array are skipped. Only the lowest level is added, as expected.

See the example below.

```php
// messages.php

return [
    'colors' => [
        'red' => 'Red',
        'green' => 'Green',
        'blue' => 'Blue',
    ],
];
```

Using the `Arr::dot` method with the filename will transform the array into the following keys:

```php
[
    'messages.colors.red',
    'messages.colors.green',
    'messages.colors.blue',
];
```

The translation `__('messages.colors')` will fail as it is not in the list. I've added a new method that will append the keys containing arrays. The file itself will also be added.

```php
[
    'messages',
    'messages.colors',
    'messages.colors.red',
    'messages.colors.green',
    'messages.colors.blue',
];
```

**Changes**

I've added support for translation groups.

**Breaking changes**

### DIFF
--- a/src/Rules/NoMissingTranslationsRule.php
+++ b/src/Rules/NoMissingTranslationsRule.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Larastan\Larastan\Rules;
 
 use Illuminate\Filesystem\Filesystem;
-use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Larastan\Larastan\Collectors\UsedTranslationFacadeCollector;
 use Larastan\Larastan\Collectors\UsedTranslationFunctionCollector;
@@ -24,6 +23,7 @@ use function array_keys;
 use function array_map;
 use function array_merge;
 use function in_array;
+use function is_array;
 use function is_dir;
 use function json_decode;
 use function lang_path;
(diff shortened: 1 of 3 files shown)


canvural:
Thank you!

PR #2398, opened 2025-11-07
fix: mark macro methods as static only if the closure is static

calebdw:
Hello!

This resolves https://github.com/larastan/larastan/discussions/2397 by only marking macros as static if the closure is static. This allows the user to control whether they want the closure to act like a static or an instance method.

Additionally, this add an [ignoreError](https://phpstan.org/developing-extensions/ignore-error-extensions) extension to automatically ignore static calls on facades (that don't exist on the facade) as these are always instance methods

CC: @dansan92

Thanks!

### DIFF
--- a/e2e/filamentphp-filament.baseline.neon
+++ b/e2e/filamentphp-filament.baseline.neon
@@ -1,151 +1,25 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Unreachable statement \\- code above always terminates\\.$#"
+			message: '#^Static call to instance method Illuminate\\Support\\Str\:\:sanitizeHtml\(\)\.$#'
+			identifier: method.staticCall
 			count: 1
-			path: ../../e2e/packages/actions/src/Commands/MakeExporterCommand.php
+			path: ../../e2e/packages/forms/src/Components/RichEditor/RichContentRenderer.php
 
 		-
-			message: "#^Unreachable statement \\- code above always terminates\\.$#"
+			message: '#^Static call to instance method Illuminate\\Support\\Str\:\:sanitizeHtml\(\)\.$#'
+			identifier: method.staticCall
 			count: 1
-			path: ../../e2e/packages/actions/src/Commands/MakeImporterCommand.php
+			path: ../../e2e/packages/infolists/src/Components/TextEntry.php
 
 		-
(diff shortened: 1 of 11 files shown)


Andyuu:
Hi Caleb. I'm using your branch so I got this change already.

I've started getting errors for this in your latest version. How should I handle this error?

  27     Static call to instance method Illuminate\Validation\Rule::dateTime().  
         🪪  method.staticCall     
         
         
Usage

Rule::dateTime()->after(now())

I have this set up as a macro in my Macro Service provider like so:

    /**
     * Bootstrap services.
     *
     * @return void
     */
    public function boot() {
        Rule::macro('dateTime', function (): DateTime {
            return new DateTime();
        });
    }
    
This date time class I use just like Rule::date()

```<?php

declare(strict_types=1);

namespace App\Rules;

use DateTimeInterface;
use Illuminate\Validation\Rules\Date;

class DateTime extends Date {
    /**
     * Format the date for the validation rule.
     */
    protected function formatDate(DateTimeInterface|string $date): string {
        return $date instanceof DateTimeInterface
            ? $date->format($this->format ?? 'Y-m-d H:i:s')
            : $date;
    }
}
```

In my _ide_helper.php file this works fine

```
namespace Illuminate\Validation {
    /**
     */
    class Rule {
        /**
         * @see \App\Providers\MacroServiceProvider::boot()
         * @return \App\Rules\DateTime
         * @static
         */
        public static function dateTime()
        {
            return \Illuminate\Validation\Rule::dateTime();
        }

            }
    }
```

but Larastan is not a fan

Thanks

calebdw:
Hi @Andyuu, you can fix this error by making the closure static since you don't access `$this` inside the closure

```php
/** @phpstan-ignore argument.staticClosure (always accessed statically) */
Rule::macro('dateTime', static function (): DateTime {
    return new DateTime();
});
```

Andyuu:
@calebdw I have tried applying the fix, but unfortunately I am still getting the same error. I also tried it on another similar macro and no results there. 

I'm not sure if there is any other information I need to give you to figure this out.

Thanks again

calebdw:
Make sure you clear the cache after applying fix (`phpstan clear-result-cache`)

Andyuu:
Great, it's working now. I've never had to run that before so I wasn't aware of it. Thanks a bunch.

dansan92:
This looks great, good job!

PR #2400, opened 2025-11-09
[3.x] fix: nullable detection for columns with DEFAULT values

faisuc:
- [x] Added or updated tests

**Changes**

Fixed nullable detection for columns with DEFAULT values. Columns like `float DEFAULT 0` without a NOT NULL constraint are now correctly identified as nullable, matching MySQL's default behavior.


src/SQL/IamcalSqlParser.php
```diff
@@ -63,6 +63,20 @@ public function parseTables(string $sql): array
         return $tables;
     }
 
+    /**
+     * @param array<string, mixed> $field
+     * 
+     * @return bool
+     */
```
szepeviktor:
```suggestion
     */
```

PR #2410, opened 2025-12-01
[3.x] Narrow HasEvents::dispatchesEvents

liamduckett:
- [✅] Added or updated tests
- [N/A] Documented user facing changes

This PR narrows the type of `HasEvents` property `dispatchesEvents` from `array<array-key, mixed>` to `array<string, class-string>`. It also narrows the getter for this property.

### DIFF
--- a/stubs/common/Database/Eloquent/Concerns/HasEvents.stub
+++ b/stubs/common/Database/Eloquent/Concerns/HasEvents.stub
@@ -1,0 +1,22 @@
+<?php
+
+namespace Illuminate\Database\Eloquent\Concerns;
+
+trait HasEvents
+{
+    /**
+     * The event map for the model.
+     *
+     * Allows for object-based events for native Eloquent events.
+     *
+     * @var array<string, class-string>
+     */
+    protected $dispatchesEvents = [];
+
+    /**
+     * Get the event map for the model.
+     *
+     * @return array<string, class-string>
+     */
+    public function dispatchesEvents();
+}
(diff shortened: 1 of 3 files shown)


calebdw:
This should really just be fixed in the framework if it's not already

liamduckett:
Agreed. I did [try and contribute it](https://github.com/laravel/framework/pull/57978) - admittedly not quite as a narrow as it could be. Given it was closed with no comment, I think a resubmit may be futile?

stubs/common/Database/Eloquent/Concerns/HasEvents.stub
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Illuminate\Database\Eloquent\Concerns;
+
+trait HasEvents
```
canvural, line 6:
```suggestion
/** @phpstan-require-extends Model */
trait HasEvents
```

This will make sure the trait is only in Models. I think it's a good addition.

canvural:
It's almost always better to add types to Larastan. That way more Laravel versions get the same benefit 🙂

liamduckett:
Sorry, I accepted the changes as is. Looks like we were missing the import for `Model`.

liamduckett:
I believe the failure to be a legitimate issue that should be raised: https://github.com/monicahq/monica/blob/main/app/Models/File.php

canvural:
Yeah, that looks fine.

Thank you!

PR #2420, opened 2025-12-18
feat: Add stubs for Cache facade

melvinversluijs:
- [x] Added or updated tests

**Issue**

When using the following code:

```php
Route::get('/', static function (): void {
    $result = \Illuminate\Support\Facades\Cache::remember(
        key: 'test',
        ttl: 60,
        callback: static fn (): string => '123',
    );

    dump(str_contains($result, '1'));
});
```

It will result in the following phpstan error:

```txt
 ------ ------------------------------------------------------------------------------
  Line   web.php
 ------ ------------------------------------------------------------------------------
  14     Parameter #1 $haystack of function str_contains expects string, mixed given.
         🪪  argument.type
 ------ ------------------------------------------------------------------------------
```

This is fixed by adding the Cache facade to the stub files of Larastan.

**Changes**

Will give you the proper return types when using cache functions using the facade.

**Breaking changes**

Not that I know of, but if I've missing anything please let me know!

### DIFF
--- a/stubs/common/Cache/CacheManager.stub
+++ b/stubs/common/Cache/CacheManager.stub
@@ -1,0 +1,9 @@
+<?php
+
+namespace Illuminate\Cache;
+
+/**
+ * @mixin \Illuminate\Cache\Repository
+ * @mixin \Illuminate\Contracts\Cache\LockProvider
+ */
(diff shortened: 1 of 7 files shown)


canvural:
Thank you!

PR #2422, opened 2025-12-23
[3.x] fix: nullability detection for TEXT and some other column types

hivokas:
- [x] Added or updated tests

**Changes**

Let's imagine there is a nullable `forms.autoresponse` column with `TEXT` type. Currently PHPStan assigns `string` type, while it should assign `string|null` instead, because the column is nullable. 

This leads to false reports like this in our project:

```
 ------ ------------------------------------------------------------------ 
  39     Property App\Models\Form::$autoresponse (string) does not accept  
         null.                                                             
         🪪  assign.propertyType                                           
 ------ ------------------------------------------------------------------ 
```

This happens because currently PHPStan defaults to `false` for "nullable" if `iamcal\SQLParser` doesn't include `'null'` key for `$field`. This happens for TEXT and some other types which you can see in "Files changed" tab.

This PR fixes this issue.

**Breaking changes**

### DIFF
--- a/src/SQL/IamcalSqlParser.php
+++ b/src/SQL/IamcalSqlParser.php
@@ -8,6 +8,7 @@ use iamcal\SQLParser as VendorIamcalSqlParser;
 use iamcal\SQLParserSyntaxException;
 
 use function array_key_exists;
+use function in_array;
 use function is_array;
 use function is_string;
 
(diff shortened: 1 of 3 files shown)


canvural (changes requested):
Hi,

Can you add also some test cases to `SquashedMigrationHelperTest`? I'd like to see test cases with a normal text column, text column with explicitly saying `NOT NULL`

hivokas:
@canvural thanks for checking, I've pushed some tests. Does it look good to you now?

hivokas:
@canvural I believe the failing test is not related to my changes.

canvural:
Thank you!

hivokas:
@canvural could you please release it when you're back from holidays?

PR #2428, opened 2026-01-09
feat: add ResponseInterface mixin to Response stub

### LINKED ISSUE #2419: Forgot the \Psr\Http\Message\ResponseInterface mixin in PendingRequest stub
- Larastan Version: 3.8.1
- Laravel Version: 12.43.1

### Description

With the new PendingRequest stub, the `\Psr\Http\Message\ResponseInterface` mixin defined in the phpdoc of the `Illuminate\Http\Client\Response` source code is no longer included. This is because the PendingRequest stub redefines the Response class as follows:

```php
/**
 * @implements \ArrayAccess<array-key, mixed>
 */
class Response implements \ArrayAccess, \Stringable {}
```

And in laravel source code : 
```php
/**
 * @mixin \Psr\Http\Message\ResponseInterface
 */
class Response implements ArrayAccess, Stringable
{
    ...
}
```

```php
$response = Http::post('https://api.example.com/data', ['key' => 'value']);

if ($response->getStatusCode() != 200) {
    Log::error("Log example: {$response->getBody()}");
}
```

phpstan errors : `Call to an undefined method Illuminate\Http\Client\Response::getStatusCode()` and `Call to an undefined method Illuminate\Http\Client\Response::getBody()`

### Expected behavior

No errors because original source code have a mixin and contains this code :

```php
/**
 * Dynamically proxy other methods to the underlying response.
 *
 * @param  string  $method
 * @param  array  $parameters
 * @return mixed
 */
public function __call($method, $parameters)
{
    return static::hasMacro($method)
        ? $this->macroCall($method, $parameters)
        : $this->response->{$method}(...$parameters);
}
```

Thank you for your work!

ramonrietdijk:
- [x] Added or updated tests

Resolves #2419

**Changes**

I've added the `ResponseInterface` as a mixin for the `Response` class just like [Laravel](https://github.com/laravel/framework/blob/12.x/src/Illuminate/Http/Client/Response.php#L15) has.

**Breaking changes**

### DIFF
--- a/stubs/12.41.0/Http/PendingRequest.stub
+++ b/stubs/12.41.0/Http/PendingRequest.stub
@@ -91,6 +91,7 @@ namespace Illuminate\Http\Client
 
     /**
      * @implements \ArrayAccess<array-key, mixed>
+     * @mixin \Psr\Http\Message\ResponseInterface
      */
     class Response implements \ArrayAccess, \Stringable {}
 
--- a/stubs/common/Http/PendingRequest.stub
+++ b/stubs/common/Http/PendingRequest.stub
@@ -91,6 +91,7 @@ namespace Illuminate\Http\Client
 
     /**
      * @implements \ArrayAccess<array-key, mixed>
+     * @mixin \Psr\Http\Message\ResponseInterface
      */
     class Response implements \ArrayAccess, \Stringable {}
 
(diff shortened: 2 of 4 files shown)


canvural:
Thank you!

PR #2432, opened 2026-01-20
[3.x] Update PendingRequest stubs

### LINKED ISSUE #2431: Incorrect type hint on PendingRequest methods
- Larastan Version: 3.8.1

This error doesn't seem to make sens, post and put should be able to take almost any data as long as it can be json serialized and get should be able to take more then one value when not a string key.

`Illuminate\Http\Client\PendingRequest<false>::get() expects array<0|string, mixed>|string|null`

This looks like a mistake where  `array{array{name: scalar, contents: scalar}}` should instead be   `list<array{name: scalar, contents: scalar}>` or `array<int, array{name: scalar, contents: scalar}>`

ramonrietdijk:
Resolves #2431

**Changes**

I've changed the multipart type to a list in the `PendingRequest` stub files. As pointed out by @AJenbo, the notation `array{array{...` causes issues with multiple arrays.

Example **without** errors:

```php
Http::asMultipart()->post('localhost', [
    [
        'name' => 'a',
        'contents' => 'a',
    ],
]);
```

Example **with** errors:

```php
Http::asMultipart()->post('localhost', [
    [
        'name' => 'a',
        'contents' => 'a',
    ],
    [
        'name' => 'b',
        'contents' => 'b',
    ],
]);

// Parameter #2 $data of method Illuminate\Http\Client\PendingRequest<false>::post() expects array<0|string, mixed>|Illuminate\Contracts\Support\Arrayable<string, mixed>|JsonSerializable, array{array{name: 'a', contents: 'a'}, array{name: 'b', contents: 'b'}} given.
```

**Breaking changes**

### DIFF
--- a/stubs/12.41.0/Http/PendingRequest.stub
+++ b/stubs/12.41.0/Http/PendingRequest.stub
@@ -28,7 +28,7 @@ namespace Illuminate\Http\Client
          * Issue a GET request to the given URL.
          *
          * @param  string  $url
-         * @param  array<string, mixed>|array{array{name: scalar, contents: scalar}}|string|null  $query
+         * @param  array<string, mixed>|list<array{name: scalar, contents: scalar}>|string|null  $query
          * @return (TAsync is true ? \Illuminate\Http\Client\Promises\LazyPromise : \Illuminate\Http\Client\Response)
          */
         public function get(string $url, $query = null);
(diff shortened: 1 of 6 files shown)


AJenbo:
Thanks, sorry for the very minimal issue report. It was super late when I wrote it.

ramonrietdijk:
No problem!

canvural (changes requested):
Thank you! Can you add a test with the example you gave? `tests/Type/data/facades-l12-41.php` looks suitable

ramonrietdijk:
Hi @canvural,

I've added an integration test instead as type assertions did not seem to cause any issues with the unmodified stubs.

canvural:
Thank you!

AJenbo:
I can confirm that I was able to update to the latest Larastan after this fix. Thanks!

PR #2434, opened 2026-01-25
Improve type inference for MySQL `ENUM` types in squashed migrations

ngmy:
- [x] Added or updated tests

**Changes**

In squashed migrations, MySQL `ENUM` types were previously inferred as `mixed`.
This PR improves type inference so that `ENUM` types are recognized as their exact values, such as `'active'|'inactive'|'don\'t know'`.

> [!NOTE]
> ~Currently, due to an issue in [iamcal/SQLParser](https://github.com/iamcal/SQLParser), MySQL `ENUM` values containing quotes are not correctly recognized, causing the tests to fail.~ 
> ~Once this PR is merged, the issue should be resolved and the tests should pass:~
>
> ~https://github.com/iamcal/SQLParser/pull/32~
>
> The PR has been merged 🎉 

**Breaking changes**

Because MySQL `ENUM` types in squashed migrations are no longer inferred as `mixed`, errors that were previously ignored may now be detected.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     "require": {
         "php": "^8.2",
         "ext-json": "*",
-        "iamcal/sql-parser": "^0.6.0",
+        "iamcal/sql-parser": "^0.7.0",
         "illuminate/console": "^11.44.2 || ^12.4.1",
         "illuminate/container": "^11.44.2 || ^12.4.1",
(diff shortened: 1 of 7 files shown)


canvural (approved):
Amazing! Thank you for fixing the issue in parser package also.

PR #2438, opened 2026-02-02
Handle wildcard string in GenericModelPropertyType

### LINKED ISSUE #2332: `model-property` type issue
- Larastan Version: 3.6.0
- Laravel Version: 12.22.1

### Description
Union types with `model-property` type work incorrectly.
For example, in the built-in larastan stub:
```php
// stubs/common/EloquentBuilder.stub

/**
 * Find a model by its primary key.
 *
 * @param  mixed  $id
 * @param  array<int, (model-property<TModel>|'*')>|model-property<TModel>|'*'  $columns
 * @phpstan-return ($id is (\Illuminate\Contracts\Support\Arrayable<array-key, mixed>|array<mixed>) ? \Illuminate\Database\Eloquent\Collection<int, TModel> : TModel|null)
 */
public function find($id, $columns = ['*']);
```

1. Fresh laravel app.
2. Install larastan

`phpstan.neon.dist`:

```yml
includes:
    - vendor/nesbot/carbon/extension.neon

parameters:

    paths:
        - app/

    level: max

    checkModelProperties: true
```

3 examples with issues in `app/Http/Controllers/Controller.php`:

```php
<?php

namespace App\Http\Controllers;

use App\Models\User;

abstract class Controller
{
    public function test(): void
    {
        $user1 = User::query()->find(1, ['*']);
        $user2 = User::query()->find(1, '*');
        $this->test1(['*']);
    }

    /**
     * @param array<int, (model-property<User>|'*')>|model-property<User>|'*' $columns
     */
    public function test1($columns): void
    {

    }
}
```

phpstan issues:

```
  11     Parameter #2 $columns of method Illuminate\Database\Eloquent\Builder<App\Models\User>::find() expects array<int, model property of App\Models\User>|model property of App\M  
         odels\User, array<int, string> given.                                                                                                                                        
         🪪  argument.type                                                                                                                                                            
         💡  Type #1 from the union: The given string should be a property of App\Models\User, * given.                                                                               
  12     Parameter #2 $columns of method Illuminate\Database\Eloquent\Builder<App\Models\User>::find() expects array<int, model property of App\Models\User>|model property of App\M  
         odels\User, string given.                                                                                                                                                    
         🪪  argument.type                                                                                                                                                            
         💡  Type #2 from the union: The given string should be a property of App\Models\User, * given.                                                                               
  13     Parameter #1 $columns of method App\Http\Controllers\Controller::test1() expects array<int, model property of App\Models\User>|model property of App\Models\User, array<int  
         , string> given.                                                                                                                                                             
         🪪  argument.type                                                                                                                                                            
         💡  Type #1 from the union: The given string should be a property of App\Models\User, * given.                                                                               
```

The issue is with the type of the `$columns` variable in the larastan stub and in another reproducible example - `test1()` method that I added with the same union type. `*` and `['*']` are acceptable values and should pass analysis, but phpstan reports them as invalid values.
If we simplify the type by removing `model-property`, replace it with `string`, for example, no issues are reported for `*` and `['*']` values:

```php
/**
 * @param array<int, (string|'*')>|string|'*' $columns
 */
public function test1($columns): void
{

}
```

szepeviktor:
Add check for wildcard string '*' in GenericModelPropertyType.

- [x] Added or updated tests

Fixes #2332

### DIFF
--- a/src/Types/ModelProperty/GenericModelPropertyType.php
+++ b/src/Types/ModelProperty/GenericModelPropertyType.php
@@ -58,6 +58,11 @@ class GenericModelPropertyType extends StringType
 
         if (count($type->getConstantStrings()) === 1) {
             $givenString = $type->getConstantStrings()[0]->getValue();
+
+            if ($givenString === '*') {
+                return AcceptsResult::createNo();
+            }
+
             $genericType = $this->getGenericType();
 
             if ($genericType instanceof MixedType) {
@@ -115,6 +120,10 @@ class GenericModelPropertyType extends StringType
         $constantStrings = $type->getConstantStrings();
 
         if (count($constantStrings) === 1) {
+            if ($constantStrings[0]->getValue() === '*') {
+                return IsSuperTypeOfResult::createNo();
+            }
+
             if (! $this->getGenericType()->hasInstanceProperty($constantStrings[0]->getValue())->yes()) {
                 return IsSuperTypeOfResult::createNo();
             }
--- a/tests/Type/GenericModelPropertyTypeTest.php
+++ b/tests/Type/GenericModelPropertyTypeTest.php
@@ -101,6 +101,15 @@ class GenericModelPropertyTypeTest extends PHPStanTestCase
             GenericModelPropertyType::class,
             'model property of App\User',
         ];
+
+        yield [
+            static fn ($container) => [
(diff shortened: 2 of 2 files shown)


canvural:
Not needed. Thank you.

PR #2445, opened 2026-02-19
Improve config type inference to respect @return docblocks

haratake22:
- [x] Added or updated tests

This is my first contribution to OSS. I apologize if there are any mistakes in the process or etiquette. I would appreciate your guidance.

**Changes**

### Problem
In Laravel config files (`config/*.php`), when dynamic values like `env()` are used, the array key types are often inferred as `mixed`.
Even if a type definition is provided using `@return array{...}` as a workaround, the existing `ConfigParser` ignored this docblock because it only parsed the content of the `return` statement (the array itself), not the statement itself.

### Solution
I modified `ConfigParser` to parse the `return` statement itself (`Return_` node) of the config file.
This allows retrieving and analyzing the `@return` tag (PHPDoc) attached to the `return` statement via `FileTypeMapper`.
As a result, if the user explicitly defines the type in the config file (e.g., `@return array{nested: string}`), Larastan will prioritize that type definition for nested keys and infer the correct type (e.g., `string`).

**Breaking changes**

None


tests/Type/ConfigReproductionTest.php
canvural, line 1: We don't need a new file. You can add the test case to `tests/Type/ConfigDynamicReturnTypeExtensionTest.php`

src/Support/ConfigParser.php
```diff
@@ -84,8 +91,43 @@ public function getTypes(array $constantStrings, Scope $scope): array
                 $this->parsedConfigFiles[$configFileName] = $cachedConfigFile;
             }
 
+            // Check if we have a type from the docblock
+            $docComment = $cachedConfigFile->getDocComment();
+
+            if ($docComment !== null) {
```
canvural: I have an idea. We can respect to [`treatPhpDocTypesAsCertain`](https://phpstan.org/config-reference#treatphpdoctypesascertain) here.

You can add it to the constructor as:
```php
#[AutowiredParameter]
private bool $treatPhpDocTypesAsCertain,
```

then change this if to `if ($docComment !== null && $this->treatPhpDocTypesAsCertain)`

(don't worry about testing this)

canvural (changes requested):
Hi,

This is a very good first contribution 👏🏽 

I only have an improvement idea. Let me know what you think about.

tests/Type/ConfigReproductionTest.php
haratake22: Thanks for the suggestion!  😃 
I've removed this file and added the test case to `tests/Type/ConfigDynamicReturnTypeExtensionTest.php`.

src/Support/ConfigParser.php
haratake22: Thanks for the idea!
 I've updated the constructor to include the `treatPhpDocTypesAsCertain` property using `#[AutowiredParameter]`.
I also added the argument definition to [extension.neon](https://github.com/larastan/larastan/pull/2445/changes#diff-c0eed1d1286633a549d491ece85cd721b01ea34037971a8f41cfa8cade5645b8R669) as I encountered a DI resolution error without it.

canvural:
You can run `./vendor/bin/phpcbf` to fix the code style issues. And also you committed a file permission change for one file. You need to undo that.

tests/application/config/reproduction.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+/** @return array{nested: array{key: string}} */
+return [
+    'nested' => [
+        'key' => env('SOME_ENV'),
+    ],
+];
```
calebdw, line 8: Just like with the `casts()` method, no one (and I repeat, literally no one) is going to want to have to duplicate the same config array structure in an `@return` phpdoc

This should be solvable by just explicitly casting `env()` calls to the appropriate type

```suggestion
return [
    'nested' => [
        'key' => (string) env('SOME_ENV'),
    ],
];
```

canvural: This is not required though. It is just, if it exists it'll prefer that one.

haratake22: @canvural 
I've restored the file permissions for tests/laravel-test.sh and ran `./vendor/bin/phpcbf`, which fixed the code style issues in `ConfigParser.php`.

haratake22:
@canvural 
I've updated the baseline files using the CI artifacts as the e2e tests failed due to stricter type inference. Is this acceptable?
Also, I noticed that one test was skipped. Is this expected behavior?

canvural:
Thank you!

PR #2454, opened 2026-03-06
fix: type inference for MySQL `ENUM` types in squashed migrations

FeBe95:
- [x] Added or updated tests

This is a follow-up PR to #2434.

**Changes**

Use `VerbosityLevel::precise()` instead of `VerbosityLevel::value()` when describing `ENUM` values from squashed migrations. This prevents long enum values from being truncated by `ConstantStringType::DESCRIBE_LIMIT`.

Truncated enum strings could lead to incorrect analysis results such as:

```md
 ------ -------------------------------------------------------------------------------------------------------
  Line   lib\SchedulingV1.php
 ------ -------------------------------------------------------------------------------------------------------
  738    Strict comparison using === between 'disable_stop…'|'keep_schedule'|'stop_schedule' and
         'disable_stop…' will always evaluate to false.
         🪪  identical.alwaysFalse
         ✏️  C:\xampp\htdocs\bidX\lib\SchedulingV1.php:738
```

**Tests**

Added a test verifying that long enum values are preserved.

```php
\PHPStan\dumpType($accounts->enum_status);
```

```diff
- `'a very long string…'|'active'|'don\'t know'|'inactive'`
+ `'a very long string that exceeds the ConstantStringType DESCRIBE_LIMIT'|'active'|'don\'t know'|'inactive'`
```

**Breaking changes**

None.

### DIFF
--- a/src/Properties/Schema/MySqlDataTypeToPhpTypeConverter.php
+++ b/src/Properties/Schema/MySqlDataTypeToPhpTypeConverter.php
@@ -28,7 +28,7 @@ final class MySqlDataTypeToPhpTypeConverter
                 return TypeCombinator::union(...array_map(
                     static fn (string $value): ConstantStringType => new ConstantStringType($value),
                     $values,
-                ))->describe(VerbosityLevel::value());
+                ))->describe(VerbosityLevel::precise());
             })(),
             default => 'mixed',
         };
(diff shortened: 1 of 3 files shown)


canvural:
Aah, nice catch! Thank you!

PR #2458, opened 2026-03-19
Also run on v13 sample application

jnoordsij:
Now that there is a stable release, this should work fine

- [x] Added or updated tests

**Changes**

**Breaking changes**


szepeviktor:
Thank you.

PR #2466, opened 2026-04-02
feat: improve map and mapWithKeys return types

canvural:
- [x] Added or updated tests

This PR improves the return type inference of `map` and `mapWithKeys` methods of Eloquent collection.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -692,6 +692,11 @@ services:
             - phpstan.broker.dynamicMethodReturnTypeExtension
 
     -
+        class: Larastan\Larastan\ReturnTypes\EloquentCollectionMapDynamicReturnTypeExtension
+        tags:
+            - phpstan.broker.dynamicMethodReturnTypeExtension
+
+    -
         class: Larastan\Larastan\Rules\NoAuthFacadeInRequestScopeRule
 
     -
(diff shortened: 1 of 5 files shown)


calebdw:
This PR introduced several issues that used to work just fine


1. This worked before and now it fails with 

`Method <class>::stations() should return Illuminate\\Support\\Collection<int, array{id: int, press_id: int, name: string}> but returns Illuminate\\Support\\Collection<int, array{id: mixed, press_id: mixed, name: mixed}>.`

```php
/** @return Collection<int, array{id: int, press_id: int, name: string}> */
public function stations(): Collection
{
    return ProductionStation::query()
        ->get()
        ->map(fn ($station) => [
            'id'       => $station->id,
            'press_id' => $station->press->id,
            'name'     => $station->name,
        ]);
}
```

2. This used to work as well and now fails (which I suspect is due to the collection `TValue` not being covariant):

`Illuminate\\Support\\Collection<(int|string), 0.0> does not accept float.`

```php
$accounts = Account::query()->get();
$category_sums = $accounts->mapWithKeys(fn ($a) => [$a->category_id => 0.]);

foreach ($accounts as $account) {
    // ...
    $category_sums[$account->category_id] += (float) (<some_expression>);
}
```

PR #2467, opened 2026-04-03
@phpstan-pure stub Illuminate\Auth\Access\Response helpers

rudiedirkx:
- [x] Added or updated tests

**Changes**

Makes Laravel's `Illuminate\Auth\Access\Response` static helpers `@phpstan-pure`, to force the use of their return values.

**Breaking changes**

No breaking changes.

----

I wouldn't know how to test. Does this require tests? Can I copy from somewhere?

### DIFF
--- a/stubs/common/Gate.stub
+++ b/stubs/common/Gate.stub
@@ -1,7 +1,7 @@
 <?php
 
 namespace Illuminate\Contracts\Auth\Access
-    {
+{
 
     interface Gate
     {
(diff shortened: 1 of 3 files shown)


szepeviktor:
> I wouldn't know how to test. Does this require tests? Can I copy from somewhere?

Yes, from here https://github.com/larastan/larastan/tree/3.x/tests/Integration

canvural:
Hi, before adding tests can you show before and after comparison of what this fix would achieve?

rudiedirkx:
Comparison? I think the test shows the use. Laravel should be full of `pure` and `impure` annotations, but they're never used, so Larastan can do that partially. It's not a Larastan bugfix, but additional userland bug detection.

PR #2478, opened 2026-04-16
fix: undefined array key internal error when view path has trailing slash

FeBe95:
- [x] Added or updated tests

**Changes**

Fixes the following internal error when running `UnusedViewsRule`:
```php
Undefined array key 1 while running CollectedDataNode rule Larastan\Larastan\Rules\UnusedViewsRule
```

The error occurs when a service provider registers views using a path with a trailing slash:
```php
$this->loadViewsFrom(__DIR__.'/resources/views/', 'some-namespace');
//                                            ↑
```

Real-world example: [jeremykenedy/laravel-logger/src/LaravelLoggerServiceProvider.php#L80](https://github.com/jeremykenedy/laravel-logger/blob/21402b998ce3eb40b266cbb0abf093f69d86b5b7/src/LaravelLoggerServiceProvider.php#L80)

Laravel does not normalize trailing slashes when storing namespace hints, so `$finder->getHints()` can return paths with a trailing separator. This causes the `explode()` delimiter in `getAllViewNames()` to never match, leaving `$viewName` without an index `1`:
```php
array:1 [
  0 => "C:\***\vendor\jeremykenedy\laravel-logger\src/resources/views\forms\clear-activity-log.blade.php"
] // vendor\larastan\larastan\src\Support\ViewFileHelper.php:84
```

The fix normalizes `$viewDirectory` with `rtrim` before constructing the explode delimiter, guarding against both `/` and `\` separators.

```php
array:2 [
  0 => ""
  1 => "forms\clear-activity-log.blade.php"
] // vendor\larastan\larastan\src\Support\ViewFileHelper.php:81
```

**Breaking changes**

_none._

### DIFF
--- a/src/Support/ViewFileHelper.php
+++ b/src/Support/ViewFileHelper.php
@@ -19,6 +19,7 @@ use function count;
 use function explode;
 use function is_dir;
 use function iterator_to_array;
+use function rtrim;
 use function str_contains;
 use function str_replace;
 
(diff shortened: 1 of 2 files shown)


canvural:
Thank you!

szepeviktor:
https://github.com/jeremykenedy/laravel-logger/pull/188

PR #2479, opened 2026-04-16
[3.x] Improve view name detection for `UnusedViewsRule`

FeBe95:
- [x] Added or updated tests

**Changes**

Improves unused view detection for mail content definitions, fluent mail view methods, and literal view names passed to `Mail::send` or `MailerContract::send`, including named arguments. Also normalizes slash-separated names in Blade directives and recognizes concatenated view names that resolve to a single constant string.

Keeps the existing error message for 3.x compatibility.

**Breaking changes**

Should be none.


canvural:
I adjusted few things. Thank you!

PR #2482, opened 2026-04-30
fix: avoid `NoEnvCallsOutsideOfConfigRule` false positives in editor mode

### LINKED ISSUE #2481: Make `NoEnvCallsOutsideOfConfigRule` editor mode aware
# Feature Request

`NoEnvCallsOutsideOfConfigRule` reports false positives in every config file when PHPStan runs in [editor mode](https://phpstan.org/user-guide/editor-mode) (`--tmp-file` / `--instead-of`). Please make the rule aware of editor-mode path swapping so it compares against the logical (real) file path rather than the analyzed temp path.

**Current behavior**

`Larastan\Rules\NoEnvCallsOutsideOfConfigRule::isCalledOutsideOfConfig()` (`src/Rules/NoEnvCallsOutsideOfConfigRule.php:80`) does:

```php
foreach ((glob($configDirectoryGlob) ?: []) as $configDirectory) {
    $absolutePath = $this->fileHelper->absolutizePath($configDirectory);
    if (! is_dir($absolutePath)) { continue; }
    if (str_starts_with($scope->getFile(), $absolutePath)) {
        return false;
    }
}
```

In editor mode, PHPStan analyzes the buffer as the `--tmp-file`, so `$scope->getFile()` (and `$scope->getFileDescription()`) return the temp path — either the literal `--tmp-file` value (e.g. `/tmp/buffer.php`) or a path under PHPStan's own mirror folder (e.g. `/private/var/folders/.../PHPStantemp_folderNNNNN/config/prism.php`). Neither starts with the real project's absolute config path, so the rule fires for every `env()` call in every config file.

**Reproduction**

With a Laravel project that has `configDirectories` set in `phpstan.neon`:

```neon
configDirectories:
    - config
    - modules/**/Infrastructure/Config
```

Run PHPStan against any config file using editor-mode flags from the project root:

```bash
cp config/prism.php /tmp/buffer.php
vendor/bin/phpstan analyse \
  --tmp-file=/tmp/buffer.php \
  --instead-of="$PWD/config/prism.php" \
  config/prism.php
```

Result: every `env()` call in `config/prism.php` is reported as `larastan.noEnvCallsOutsideOfConfig`. Without the editor-mode flags, zero errors. (Verified on `larastan/larastan` v3.9.6 with `phpstan/phpstan` 2.1.54.)

**Impact**

PhpStorm's PHPStan integration enables editor mode by default starting with PhpStorm 2025.1 (PHPStan 2.1.17+ / 1.12.27+). With editor mode on, every Laravel config file lights up red in the IDE while CLI runs and PhpStorm's "Full Project run (batch mode)" remain clean. Users either disable editor mode (losing live diagnostics) or disable `noEnvCallsOutsideOfConfig` entirely (losing a useful rule) until this is fixed.

**Suggested fix**

PHPStan exposes the original path via the `--instead-of` argument; the rule needs access to that mapping when deciding whether the analyzed file lives under a configured config directory. A few options:

1. Resolve via PHPStan's `OptionalInsteadOfFileFinder` / equivalent service so `$scope->getFile()` is mapped back to its `--instead-of` path before the prefix check.
2. Skip the rule entirely when editor-mode swapping is active (detectable from PHPStan's input options).
3. Compare against the basename / relative path under the analyzed file rather than the absolute prefix, so the temp mirror layout (`<tmpdir>/config/prism.php`) still matches `config/`.

**Additional context**

- larastan/larastan v3.9.6
- phpstan/phpstan 2.1.54
- PHP 8.5, Laravel 13
- Reported via PhpStorm's bundled inspection (Editor mode + Full Project run both enabled), reproduced standalone with the CLI invocation above.

billypoke:
# NoEnvCallsOutsideOfConfigRule — editor-mode false positives

Closes #2481

## Fix

PHPStan exposes normalized editor-mode paths through its container parameters. The rule now receives `singleReflectionFile` and `singleReflectionInsteadOfFile`, then checks the logical path when PHPStan analyzes a temporary file.

The protected `isCalledOutsideOfConfig(FuncCall, Scope)` hook remains unchanged.

## Tests

- Maps a temporary editor file to a config file without reporting errors.
- Reports calls when the logical file is outside configured directories.
- Preserves normal behavior when the analyzed file is not the editor temporary file.

PHPStan rejects incomplete `--tmp-file` / `--instead-of` pairs before it invokes rules.

### DIFF
--- a/src/Rules/NoEnvCallsOutsideOfConfigRule.php
+++ b/src/Rules/NoEnvCallsOutsideOfConfigRule.php
@@ -17,8 +17,12 @@ use PHPStan\Rules\RuleErrorBuilder;
 use function config_path;
 use function count;
 use function glob;
+use function is_array;
 use function is_dir;
+use function is_string;
 use function str_starts_with;
+use function strlen;
+use function substr;
 
 /**
  * Catches `env()` calls outside of the config directory.
@@ -32,6 +36,10 @@ class NoEnvCallsOutsideOfConfigRule implements Rule
     /** @var list<string> */
     private array $configDirectories = [];
 
+    private string|null $editorTmpFile;
+
+    private string|null $editorInsteadOfFile;
+
     /** @param  list<non-empty-string> $configDirectories */
     public function __construct(array $configDirectories, private FileHelper $fileHelper)
     {
@@ -39,11 +47,12 @@ class NoEnvCallsOutsideOfConfigRule implements Rule
             foreach ($configDirectories as $directory) {
                 $this->configDirectories[] = $this->fileHelper->normalizePath($directory);
             }
-
-            return;
+        } else {
+            $this->configDirectories = [config_path()]; // @phpstan-ignore-line
         }
 
-        $this->configDirectories = [config_path()]; // @phpstan-ignore-line
+        $this->editorTmpFile       = $this->resolveCliPath('tmp-file');
+        $this->editorInsteadOfFile = $this->resolveCliPath('instead-of');
     }
 
     public function getNodeType(): string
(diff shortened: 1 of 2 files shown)


billypoke:
@canvural There's now a better way of doing this, without relying on `$_SERVER['argv']` parsing. Please let me know if you want any further changes.

PR #2483, opened 2026-05-07
Add type narrowing for `where` method on collections

jnoordsij:
- [x] Added or updated tests

**Changes**

The `CollectionFilterRejectDynamicReturnTypeExtension` class provides inferred type-narrowing for methods on collection. This PR adds `where` with callbacks to the already supported `filter` and `reject` methods. This allows developers to prefer `where` is stylistically preferable, e.g. when chaining with other `where` type calls or to be on par with query methods or to be on par with other languages like C# LINQ, and still profit from narrowed types.

Note this could later be improved to also narrow other input types that `where` accepts, but for the purpose of this MR these situations are not considered.

**Breaking changes**

n/a


canvural:
Thank you!

PR #2488, opened 2026-05-25
fix: invalid parameter name

calebdw:
Hello!

I added some rules to [Rector](https://x.com/calebdw/status/2057548272307245071) and this is causing errors because dashes (`-`) are illegal in php parameter names. I also shortened it because there's  no reason for it to be so long

Thanks!


canvural:
Thank you!

PR #2496, opened 2026-06-06
fix: build on latest PHPStan

staabm:
types got more precise in recent phpstan-src releases

related upstream PR is https://github.com/phpstan/phpstan-src/pull/5794


staabm:
(not sure how the CI errors are related)

calebdw:
you also need to update the min phpstan version so all of the `prefer-lowest` tests don't fail

canvural:
Thank you!

Issue #236, opened 2019-02-07
FromRequest class say "Illuminate\Validation\ValidationException  : The given data was invalid."


hokutoasari:
- Larastan Version: 0.3.15
- `--level` used: 5 (default, and I tried `--level=1` also.)

### Description:

I run `$ php artisan code:analyse --debug -v` and I see following lines.

```
   Illuminate\Validation\ValidationException  : The given data was invalid.

  at /var/www/app/vendor/laravel/framework/src/Illuminate/Foundation/Http/FormRequest.php:126
    122|      * @throws \Illuminate\Validation\ValidationException
    123|      */
    124|     protected function failedValidation(Validator $validator)
    125|     {
  > 126|         throw (new ValidationException($validator))
    127|                     ->errorBag($this->errorBag)
    128|                     ->redirectTo($this->getRedirectUrl());
    129|     }
    130|

  Exception trace:

  1   Illuminate\Foundation\Http\FormRequest::failedValidation(Object(Illuminate\Validation\Validator))
      /var/www/app/vendor/laravel/framework/src/Illuminate/Validation/ValidatesWhenResolvedTrait.php:26

  2   Illuminate\Foundation\Http\FormRequest::validateResolved()
      /var/www/app/vendor/laravel/framework/src/Illuminate/Foundation/Providers/FormRequestServiceProvider.php:30
```

### Laravel code where the issue was found:

May be this errors fire in following code. But I can't have confidence in that.

```php
use App\Http\Requests\Contracts\Admin\Authors\AuthorStoreRequest;

final class AuthorStoreUseCase implements AuthorStoreUseCaseInterface
{
    /**
     * @param AuthorStoreRequest $request
     * @return AuthorEntity
     */
    public function __invoke(AuthorStoreRequest $request): AuthorEntity
    {
        // snip
    }
}
```

The `App\Http\Requests\Contracts\Admin\Authors\AuthorStoreRequest` is just interface like following lines.

```php
interface AuthorStoreRequest extends BaseRequest
{
    /**
     * @return array
     */
    public function rules(): array;
}
```

The `BaseRequest` as just a interface. That was bind to following abstract class.

```php
use Illuminate\Foundation\Http\FormRequest;

/**
 * Class BaseRequest
 * @package App\Http\Requests
 */
abstract class BaseRequest extends FormRequest
{
    //
}
```

It just extends `FormRequest` class.

But I can't have confidence in that...

[...]

PR #2499, opened 2026-06-17
feat: narrow getAppends and getMutatedAttributes return types

dfinchenko:
- [x] Added or updated tests

<!-- Link to related issues this PR resolves, e.g. "Resolves #236"-->

**Changes**
Narrows the return types of `Model::getAppends()` and `Model::getMutatedAttributes()` from `array` to `list<string>` via the `HasAttributes` stub.            
                                                                                                                                                                                                                                         Both return lists of attribute-name strings in Laravel's source (`getAppends()` returns `$this->appends`; `getMutatedAttributes()` returns the mapped mutator cache), but are declared as plain `@return array` upstream, so PHPStan infers `array`.          

<!-- Detail the changes in behaviour this PR introduces. -->

**Breaking changes**
None.

<!-- Are existing use cases affected and require changes when upgrading? 
If so, describe the necessary changes in UPGRADE.md. -->

### DIFF
--- a/stubs/common/Database/Eloquent/Concerns/HasAttributes.stub
+++ b/stubs/common/Database/Eloquent/Concerns/HasAttributes.stub
@@ -11,4 +11,18 @@ trait HasAttributes
      * @return array<string, string>
      */
     public function getCasts();
+
+    /**
+     * Get the accessors that are being appended to model arrays.
+     *
+     * @return list<string>
+     */
+    public function getAppends();
+
+    /**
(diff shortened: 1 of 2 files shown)


dfinchenko:
Larastan currently infers a bare `array` for `getAppends()` and `getMutatedAttributes()`, while both are lists of attribute names - the stubs narrow them to `list<string>`, with type assertions covering each.

@canvural is there anything you'd like changed before this can go in?

canvural:
Thank you!

PR #2500, opened 2026-06-23
fix: apply Eloquent PHP attributes when resolving model metadata

### LINKED ISSUE #2463: #[Table] attribute not respected in L13
- Larastan Version: 3.9.3
- Laravel Version: 13.2

### Description
Whereas Larastan will interpret the following in a model and correctly hint `$id` as `string`:

```php
    public $incrementing = false;

    protected $keyType = 'string';
```

If it's replaced by the new L13 attribute syntax:

```php
#[Table(keyType: 'string', incrementing: false)]
```

It will result in errors like the following surfacing in PHPStan:

```
Property App\Models\Foo::$id (int) does not accept string.

Parameter $fooId of class App\Services\FooConsumer constructor expects string, int given.
```

NikitaKirin:
- [x] Added or updated tests

Resolves #2463

**Changes**

Larastan creates model instances using `ReflectionClass::newInstanceWithoutConstructor()` to avoid
side effects. This skips the Eloquent constructor, which normally calls
`initializeModelAttributes()` — the method responsible for applying PHP 8 attributes such as
`#[Table]`, `#[ObservedBy]`, `#[WithoutTimestamps]`, and others introduced in Laravel 13. As a
result, attributes were silently ignored, causing false positives such as:

```
Access to an undefined property App\Models\Account::$user_id.
```

or incorrect primary key types:

```
Property App\Models\Foo::$id (int) does not accept string.
Parameter $id of constructor expects string, int given.
```

A minimal reproduction:

```php
#[Table(name: 'users')]
class Member extends Model {}
```

```php
$name = Member::first()->name; // Access to an undefined property App\Models\Member::$name.
```

Larastan resolves the table name before `initializeModelAttributes()` is called, so it looks up
the non-existent `members` table (derived from the class name) instead of `users` declared via
`#[Table]`.

Fix: after `newInstanceWithoutConstructor()`, call `initializeModelAttributes()` when available,
guarded with `method_exists()` for compatibility with Laravel 11/12. Applied in four locations:
`ModelPropertyHelper::hasDatabaseProperty()`, `ModelPropertyHelper::getDatabaseProperty()`,
`ModelCastHelper::getModelCasts()`, and `SchemaAggregator::getModelReferenceType()`.

Note: on Laravel 13 PHPStan knows `initializeModelAttributes()` always exists on `Model`, so
the `method_exists()` check is marked with `@phpstan-ignore function.alreadyNarrowedType` – the
guard is intentional for cross-version compatibility, not redundant logic.

New tests (skipped on Laravel 11/12 where the feature does not exist):

- `ModelPropertyHelperTest` — verifies that `hasDatabaseProperty()` resolves columns from the
  correct table declared via `#[Table]`.
- `ModelCastHelperTest` — verifies that `getCastForProperty()` uses the primary key and key type
  declared via `#[Table(key: '...', keyType: '...')]`.
- `MigrationHelperTest` — verifies that `foreignIdFor(Member::class)` resolves to
  `non-negative-int` (from the correct table's primary key) instead of falling back to `int`.

**Breaking changes**

None.

### DIFF
--- a/src/Properties/ModelCastHelper.php
+++ b/src/Properties/ModelCastHelper.php
@@ -61,6 +61,7 @@ use function array_map;
 use function array_merge;
 use function class_exists;
 use function explode;
+use function method_exists;
 use function str_replace;
 
 class ModelCastHelper
@@ -255,6 +256,10 @@ class ModelCastHelper
         try {
             /** @var Model $modelInstance */
             $modelInstance = $modelClassReflection->getNativeReflection()->newInstanceWithoutConstructor();
+            // @phpstan-ignore function.alreadyNarrowedType (method exists only since Laravel 13)
+            if (method_exists($modelInstance, 'initializeModelAttributes')) {
+                $modelInstance->initializeModelAttributes();
+            }
         } catch (ReflectionException) {
             throw new ShouldNotHappenException();
         }
--- a/src/Properties/ModelPropertyHelper.php
+++ b/src/Properties/ModelPropertyHelper.php
@@ -70,8 +70,7 @@ class ModelPropertyHelper
         }
 
         try {
-            /** @var Model $modelInstance */
-            $modelInstance = $classReflectionOrTable->getNativeReflection()->newInstanceWithoutConstructor();
+            $modelInstance = $this->newModelInstance($classReflectionOrTable);
         } catch (ReflectionException) {
             return false;
         }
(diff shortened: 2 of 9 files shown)


rajandangi:
Tested this patch against a Laravel 13.24 application with 111 Eloquent models using the new metadata attributes. After clearing the PHPStan result cache, errors dropped from 2,817 to 0 and the full lint command passed, including 100% type coverage. This would unblock our Laravel 13 model attribute migration. Thanks for the fix.

cyppe:
Looking forward for this.

canvural:
Thank you! This looks good, I'll make small changes later.

PR #2501, opened 2026-06-24
fix: Allow mixins to support Eloquent computed attributes

imliam:
- [x] Added or updated tests

**Changes**

Fixes a scenario where Eloquent computed attributes (e.g. `protected function isExpired(): Attribute` without generics are not resolved when accessed through a `@mixin` annotation - the typical pattern in `JsonResource` classes and other places.

This change treats them as existing but with a `mixed` type.

Here is a scenario that would have previously not worked, but does with these changes:

```php
/** @mixin User */
class UserResource extends JsonResource
{
    public function toArray($request)
    {
        return [
            'computed_value' => $this->computed_value,
        ];
    }
}

class User extends Authenticatable
{
    protected function computedValue(): Attribute
    {
        return Attribute::make(
            fn ($value) => $value,
            fn ($value) => $value,
        );
    }
}
```

**Breaking changes**

Computed attributes declared with a raw `Attribute` return type were previously treated as non-existent by Larastan. They are now visible as mixed-typed properties (unless they have generics). At higher levels (8+) this may surface new mixed-related errors for code that was previously hidden - the fix is to add `@return Attribute<TGet, TSet>` to the method.

### DIFF
--- a/src/Properties/ModelPropertyHelper.php
+++ b/src/Properties/ModelPropertyHelper.php
@@ -12,8 +12,6 @@ use PHPStan\PhpDoc\TypeStringResolver;
 use PHPStan\Reflection\ClassReflection;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Constant\ConstantStringType;
-use PHPStan\Type\Generic\GenericObjectType;
-use PHPStan\Type\MixedType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\TypeCombinator;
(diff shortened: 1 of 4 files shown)


calebdw:
Just add the generics to the method?

imliam:
> Just add the generics to the method?

Sure, that's the ideal thing (and what I've done in the project this came up on), but I'm proposing this upstream because it was confusing figuring out the actual problem here (some attributes worked, some didn't), especially as the the generics aren't _actually_ part of the Laravel Attribute class.

If we can infer the attribute exists but the type is unknown, why not provide that? It will give more useful errors to users and is technically more correct.

PR #1842, opened 2024-02-11
[3.x] fix: default date casting

calebdw:
**Changes**
Closes #1744

This has been released on my [fork](https://github.com/calebdw/larastan) if anyone would like to take advantage of this immediately.

Thanks!

PR #2506, opened 2026-07-01
fix: remove dead cases and handle nullableTimestampsTz in schema aggregator

### LINKED ISSUE #2504: dead code/bogus logic
please see

https://github.com/larastan/larastan/blob/ef28a38804063c0e26deb18680d964e723048f3d/src/Properties/SchemaAggregator.php#L265-L295

and some related phpstan errors (phpstan-dev version)

```
  273    Switch condition comparison between 'nullabletimestamps'|'nullabletimestampstz'|'remembertoken'|'timestamps'|'timestampstz' and 'droptimestamps' is always false.                                
         🪪  switch.alwaysFalse                                                                                                                                                                           
  274    Switch condition comparison between 'nullabletimestamps'|'nullabletimestampstz'|'remembertoken'|'timestamps'|'timestampstz' and 'droptimestampstz' is always false.                              
         🪪  switch.alwaysFalse                                                                                                                                                                           
  283    Switch condition comparison between 'nullabletimestamps'|'nullabletimestampstz'|'timestamps'|'timestampstz' and 'dropremembertoken' is always false.                                             
```

lazerg:
The `if` guarding this switch only lets through the non-drop timestamp and `rememberToken` methods, so the `dropTimestamps`, `dropTimestampsTz` and `dropRememberToken` cases can never match, which is why PHPStan reports them as `switch.alwaysFalse`. This removes those unreachable cases and adds the missing `nullableTimestampsTz` case, which the guard already allowed but was never handled. Added a test covering the `nullableTimestampsTz` mapping.

Fixes #2504

### DIFF
--- a/src/Properties/SchemaAggregator.php
+++ b/src/Properties/SchemaAggregator.php
@@ -262,26 +262,18 @@ final class SchemaAggregator
                     }
                 }
 
-                if (
-                    $firstMethodCall->name->name === 'timestamps'
-                    || $firstMethodCall->name->name === 'timestampsTz'
-                    || $firstMethodCall->name->name === 'nullableTimestamps'
-                    || $firstMethodCall->name->name === 'nullableTimestampsTz'
-                    || $firstMethodCall->name->name === 'rememberToken'
-                ) {
-                    switch (strtolower($firstMethodCall->name->name)) {
-                        case 'remembertoken':
-                            $table->setColumn(new SchemaColumn('remember_token', 'string', $nullable));
-                            continue 2;
+                switch (strtolower($firstMethodCall->name->name)) {
+                    case 'remembertoken':
+                        $table->setColumn(new SchemaColumn('remember_token', 'string', $nullable));
+                        continue 2;
 
-                        case 'timestamps':
-                        case 'timestampstz':
-                        case 'nullabletimestamps':
-                        case 'nullabletimestampstz':
-                            $table->setColumn(new SchemaColumn('created_at', 'string', true));
-                            $table->setColumn(new SchemaColumn('updated_at', 'string', true));
-                            continue 2;
-                    }
+                    case 'timestamps':
+                    case 'timestampstz':
+                    case 'nullabletimestamps':
+                    case 'nullabletimestampstz':
+                        $table->setColumn(new SchemaColumn('created_at', 'string', true));
+                        $table->setColumn(new SchemaColumn('updated_at', 'string', true));
+                        continue 2;
                 }
 
                 $defaultsMap = [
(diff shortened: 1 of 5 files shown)


src/Properties/SchemaAggregator.php
szepeviktor: How about simply this?
```php
switch (strtolower($firstMethodCall->name->name)) {
    case 'remembertoken':
        $table->setColumn(new SchemaColumn('remember_token', 'string', $nullable));
        continue 2;

    case 'timestamps':
    case 'timestampstz':
    case 'nullabletimestamps':
    case 'nullabletimestampstz':
        $table->setColumn(new SchemaColumn('created_at', 'string', true));
        $table->setColumn(new SchemaColumn('updated_at', 'string', true));
        continue 2;
}
```

szepeviktor: I mean without the guarding `if`!

lazerg: Done, pushed.

szepeviktor: Thank you.

src/Properties/SchemaAggregator.php
```diff
@@ -262,37 +262,18 @@ private function processColumnUpdates(string $tableName, string $argName, array
...
-                        case 'timestamps':
-                        case 'timestampstz':
-                        case 'nullabletimestamps':
-                            $table->setColumn(new SchemaColumn('created_at', 'string', true));
-                            $table->setColumn(new SchemaColumn('updated_at', 'string', true));
-                            break;
-                    }
-
-                    continue;
+                switch (strtolower($firstMethodCall->name->name)) {
```
canvural, line 265: What happened to `droptimestamps` and `droptimestampstz`? We should now support them. So adding them to the switch case should be fine, right?

lazerg: Good catch, added them back and pushed.

szepeviktor: @canvural The original guarding `if` did not allow drop* methods!

szepeviktor: Please handle `dropremembertoken` too

lazerg: Done, pushed.

lazerg:
Hi @canvural, just following up since it's been a couple weeks with no response. Could you take a look when you have a moment? Thanks.

calebdw:
> just following up since it's been a couple weeks with no response

Lol that's cute 🙃, my oldest PR (#1842) has been open 3 years with no response

canvural:
Thank you!

PR #2511, opened 2026-07-13
Implement strict Contract mode

### LINKED ISSUE #2216: Adding a "Strict Contract Mode"
## **Why This Matters**  

Laravel developers often rely on dependency injection and interfaces to define clear application contracts. However, a common issue in the community is the tendency to type-hint interfaces but then make assumptions about the underlying runtime implementation. This practice can lead to unexpected errors, especially when different implementations are used in production versus local development or test environments.  

Currently, Larastan attempts to infer return types by resolving bindings in the container, which can lead to incorrect assumptions. This is particularly problematic when an interface has multiple possible implementations, as Larastan may infer a return type that is only valid in some configurations but not others.  

By introducing a **"Strict Contract Mode"**, Larastan can help developers adhere more closely to proper interface-based design while reducing unintended misinterpretations of runtime behavior.  

## **The Core Idea**  

The proposal is to introduce an **optional setting** in the Larastan configuration:  

```yaml
parameters:
    checkStrictContracts: false
```
When enabled, Larastan will **only attempt to resolve dependencies when the provided argument is not a class or interface Fully Qualified Name (FQN)**.  

This means that:  
- If an argument is explicitly an interface FQN, Larastan will **not** attempt to resolve it to a runtime implementation.  
- Developers will be required to **adhere strictly to the contract** of the interface they are type-hinting, reducing the risk of incorrect assumptions.  
- The default behavior remains unchanged (`false`), ensuring **backward compatibility** while giving teams the option to enforce stricter rules.  

### **1. Encourages Proper Interface-Based Design**  
One of the core principles of good software architecture is adhering to defined contracts. If a method expects an interface, the caller should not assume additional methods exist beyond what the interface guarantees. By making developers more aware of these mistakes, Larastan can encourage better design practices rather than masking potential issues.  

### **2. Prevents Incorrect Type Inference Across Environments**  
As highlighted in previous discussions, current behavior can lead to incorrect return type assumptions, such as assuming a specific class is always returned when, in reality the implementation can vary based on the application’s configuration. A strict mode would prevent such incorrect inferences and make sure developers explicitly handle all possible cases.  

### **3. Provides Flexibility Without Forcing a Breaking Change**  
By making this an **opt-in** feature, Larastan can continue working as it does today for those who prefer the current behavior. However, teams that prioritize strict adherence to interfaces can enable the setting to enforce better coding practices in their projects.  

### **4. Aligns With Industry Best Practices**  
Strict contract enforcement is a common best practice in statically-typed languages and strongly encourages more predictable, maintainable code. By allowing developers to choose this mode, Larastan can cater to those who value stricter type safety while still accommodating Laravel’s flexible nature.  

## **Next Steps**  

I’m happy to implement this myself, but I wanted to share the proposal first to ensure that the idea aligns with the direction of Larastan before spending time on a PR. If this approach seems promising, I’d be excited to move forward with the implementation.

AJenbo:
- [x] Added or updated tests

Resolves #2216

**Changes**

Adds an opt-in `checkStrictContracts` config parameter (default `false`). When enabled, Larastan no longer resolves container bindings for arguments to `resolve()`, `app()`, `App::make()`/`App::makeWith()`, and `Container::make()`/`makeWith()`/`resolve()` when the argument is already a class or interface FQCN it infers the return type as that FQCN itself instead of digging into the container to find the concrete implementation currently bound to it. String aliases (e.g. `resolve('cache')`) are unaffected and continue to resolve to their concrete implementation as before.

This is implemented in a single shared helper (`AppMakeHelper::resolveTypeFromCall()`), which backs all of the call sites above, so enabling the flag consistently affects every resolution path in one place.

**Breaking changes**

None, the option is off by default.

### DIFF
--- a/docs/custom-config-parameters.md
+++ b/docs/custom-config-parameters.md
@@ -193,3 +193,25 @@ $foo = env('FOO', fn() => 'bar');
 will also dump `string` when the parameter is enabled.
 
 The reasoning behind it is that the default value is always used when the environment variable is not set. And for almost all the cases the default value should be the same type as the actual env variable. So it makes sense to generalize the return type to the type of the default value.
+
+## `checkStrictContracts`
+**default**: `false`
+
+By default, when Larastan sees a class or interface FQCN passed to `resolve()`, `app()`, `App::make()`/`App::makeWith()`, or `Container::make()`/`makeWith()`/`resolve()`, it resolves the argument through the container bindings and infers the return type as the concrete class that would be bound at runtime. This can lead to incorrect assumptions when the binding varies between environments (e.g. production vs. testing), since code may end up relying on methods that only exist on the concrete implementation and not on the interface it type-hinted.
+
+When this option is enabled, Larastan will not resolve container bindings for arguments that are already a class or interface FQCN, and will instead infer the type as the FQCN itself. Aliases (e.g. `resolve('cache')`) are unaffected and continue to resolve to their concrete implementation.
+
+### Example
+```php
+use Illuminate\Contracts\Config\Repository;
+
+$repository = resolve(Repository::class);
+\PHPStan\dumpType($repository);
+```
+
+With `checkStrictContracts` disabled (default), this dumps `Illuminate\Config\Repository` (the concrete implementation bound in the container). With it enabled, this dumps `Illuminate\Contracts\Config\Repository` (the interface that was type-hinted).
+
+```neon
+parameters:
+    checkStrictContracts: true
+```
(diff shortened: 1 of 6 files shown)


szepeviktor:
Thank you for the PR.

What is your use-case?

AJenbo:
1: my application dosent like being booted in the CI environment (this limits it to static analasis without the runtime probing)
2: I would like my application to stay compatible with the interface, larastan make developers code against the implementation
3: its faster

for more details see the linked issue

PR #2514, opened 2026-07-13
chore: remove `ext-json` dependency from `composer.json`

maks-oleksyuk:
- [ ] Added or updated tests

**Changes**

> For composer.json file that had ext-json in the require section, this requirement is no longer necessary if already requires php ^8.0.

https://php.watch/versions/8.0/ext-json

**Breaking changes**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,6 @@
     ],
     "require": {
         "php": "^8.2",
-        "ext-json": "*",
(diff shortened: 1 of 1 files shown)


szepeviktor:
https://github.com/php/php-src/commit/cc27781f40f1df6234b277d460d9bb20ec540a12

szepeviktor:
Thank you.

PR #2515, opened 2026-07-14
ci: bump GitHub Actions to latest major versions

maks-oleksyuk:
- `actions/checkout@v4`→`v7`, `actions/cache@v4`→`v6`, `actions/upload-artifact@v4`→`v7` (+ `upload-artifact/merge`) — clears the Node 20 deprecation warnings
- `phpunit/phpunit` floor bumped to `^13.3.0`
- test deprecation fixes: declared property instead of dynamic, drop redundant `ReflectionProperty::setAccessible(true)`, `expects($this->once())`

Rebased onto latest `3.x`; the earlier schema-aggregator and e2e-baseline churn is now covered upstream (#2506, #2528).

### DIFF
--- a/.github/workflows/coding-standard.yml
+++ b/.github/workflows/coding-standard.yml
@@ -29,10 +29,10 @@ jobs:
 
     steps:
       - name: "Checkout"
-        uses: "actions/checkout@v4"
+        uses: actions/checkout@v7
 
       - name: "Setup PHP"
-        uses: "shivammathur/setup-php@v2"
+        uses: shivammathur/setup-php@v2
         with:
           php-version: "${{ matrix.php }}"
           extensions: "dom, curl, libxml, mbstring, zip, fileinfo"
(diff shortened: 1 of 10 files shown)


calebdw:
What bugs are you trying to fix?

maks-oleksyuk:
> What bugs are you trying to fix?

@calebdw update description

calebdw:
This should be fixed by https://github.com/sebastianbergmann/phpunit/issues/6831 (might need to update phpunit version)

maks-oleksyuk:
In that case, this PR might have to wait for the new version, since this isn't the only fix that's been made; another e2e test has also been fixed, and outdated actions have been updated

szepeviktor:
@maks-oleksyuk Thank you for your PR.
Please change only 1 thing in a PR.

maks-oleksyuk:
@szepeviktor It seems that PHPUnit version 13.2.5 doesn't provide the expected fix; instead, a different error occurs, which you can see in CI
UPD:
see https://github.com/sebastianbergmann/phpunit/issues/6855

maks-oleksyuk:
@szepeviktor, check it out; it looks like all the checks need to be passed

Also, I'd rather not split this into smaller PRs, since it mainly involves CI fixes and infrastructure updates needed to unblock other PRs. Maybe we could just rename the PR if necessary

szepeviktor:
Thank you for your work.
Please do 1 thing in this PR.
No quote removal, no `--ansi` ...

szepeviktor:
Now it looks Very Nice!

What is the reason behind removing roave/security-advisories and that Composer security var?

maks-oleksyuk:
composer remove roave/security-advisories in e2e-tests.yml — the e2e job re-resolves dependencies on old pinned forks (monica, filament, etc.). If those repos carry roave/security-advisories as a dev dep, a newly-disclosed CVE in some transitive package can hard-fail the whole composer update, unrelated to what's actually being tested (larastan compatibility). Removing it first avoids that - same pattern already used for tomasvotruba/bladestan on monica.

COMPOSER_NO_SECURITY_BLOCKING=1 - Composer ≥2.9 blocks installing packages with unpatched security advisories. phpstan-dev.yml intentionally tests against older Laravel versions, so this flag disables that block for the compat test.

canvural (approved):

canvural:
Thank you!

PR #2519, opened 2026-07-25
feat: detect anonymous Blade component views in checkUnusedViews

G4Zz0L1:
`UnusedViewsRule` (`checkUnusedViews`) only traced `view()` / `@include` / `@extends`, so anonymous Blade components rendered via `<x-...>` tags were always reported as unused — a false positive for any component-driven app.

`UsedViewInAnotherViewCollector` now also scans view HTML for `<x-name>` / `<x-a.b />` / `</x-name>` tags and marks the corresponding view used, under both the short (`name`, when `resources/views/components` is a registered view path) and prefixed (`components.name`) namings, plus their `.index` variants. `<x-slot>` is skipped (it is a slot, not a component).

Adds a regression test: a component used only through `<x-alert>` is no longer flagged, while an unreferenced component still is.

### DIFF
--- a/src/Collectors/UsedViewInAnotherViewCollector.php
+++ b/src/Collectors/UsedViewInAnotherViewCollector.php
@@ -8,10 +8,7 @@ use Larastan\Larastan\Support\ViewFileHelper;
 use Larastan\Larastan\Support\ViewParser;
 use PhpParser\Node;
 
-use function array_filter;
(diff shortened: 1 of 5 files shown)


szepeviktor:
What a nice PR.
Thank you.

canvural:
Thank you!

PR #2520, opened 2026-07-30
fix: resolve typed migration table constants

riasvdv:
- [x] Added or updated tests

**Changes**

Fixes a bug with my earlier PR (#2315) where imported class constants aren’t recognized.

**Breaking changes**

None


riasvdv:
CI failures seem unrelated

canvural (changes requested):
I don't think this fixes it. The test passes on `3.x` without these changes. Can you add a test that fails before the fix and handle typed constant values?

riasvdv:
@canvural Seems like that was the wrong test (and as a result, fix). It worked in our project because the string based migration was still cached at the time.

I've updated the implementation

canvural (approved):

canvural:
Thank you!

PR #2523, opened 2026-08-09
feat: add `createQuietly()` to `EloquentBuilder.stub` for model-property checking

Mr-Offline:
- [x] Added or updated tests

Resolves part of #676

**Changes**

Adds a stub declaration for `Illuminate\Database\Eloquent\Builder::createQuietly()` so Larastan validates the `$attributes` array keys against the model's real properties, the same way it already does for `create()`.

`createQuietly()` was missing from `stubs/common/EloquentBuilder.stub`, even though it's functionally identical to `create()` (it just wraps the call in `Model::withoutEvents(...)`). Without a stub entry, PHPStan fell back to Laravel's native `array $attributes = []` signature, which accepts any string key and provides no typo protection.

Before this change:
`
User::query()->createQuietly([
    'emali' => 'test@example.com', // typo — not caught before this PR
]);
`
After this change, Larastan reports an error for invalid keys, while valid ones still pass:
`
User::query()->createQuietly(['foo' => 'bar']); // reported: `foo` is not a property on App\User
User::query()->createQuietly(['id' => 1]);       // passes
`

Specifically:
- `stubs/common/EloquentBuilder.stub` — added a `createQuietly()` method declaration with `@phpstan-param array<model-property<TModel>, mixed> $attributes`, following the same pattern as the existing `create()` method.
- `tests/Integration/data/model-property-builder.php` — added two integration test calls: one with an invalid attribute key (`foo`) and one with a valid one (`id`).
- `tests/Integration/IntegrationTest.php` — added the expected PHPStan error for the invalid `createQuietly()` call.

**Breaking changes**

None. This only adds stricter checking for a method that previously had no dedicated stub, so no existing passing code becomes invalid unless it was already passing genuinely incorrect attribute keys to `createQuietly()`.

### DIFF
--- a/stubs/common/EloquentBuilder.stub
+++ b/stubs/common/EloquentBuilder.stub
@@ -45,6 +45,14 @@ class Builder
     public function create(array $attributes = []);
 
     /**
+     * Save a new model and return the instance without raising model events.
+     *
+     * @phpstan-param array<model-property<TModel>, mixed> $attributes
+     * @phpstan-return TModel
+     */
+    public function createQuietly(array $attributes = []);
+
+    /**
      * Create a collection of models from plain arrays.
      *
      * @param  array<mixed>  $items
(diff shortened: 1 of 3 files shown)


canvural:
Thank you!

PR #2527, opened 2026-08-11
chore(docs): update the documentation concerning model type inference.

dantleech:
This PR updates the docuemntation:

- Do not suggest that relationships can be inferred automatically as that
  behavior was [removed in Larastan 3.0](https://github.com/larastan/larastan/discussions/2524#discussioncomment-17966367).
- Document the fact that Larastan will scan the migrations folder to infer
  property types.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -223,15 +223,15 @@ class User extends Model
 > }
 > ```
 
+## Model Properties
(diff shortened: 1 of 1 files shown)


canvural:
Thank you!

PR #2529, opened 2026-08-11
fix: respect model class attributes when resolving properties

### LINKED ISSUE #2463: #[Table] attribute not respected in L13

sinemacula-ben:
- [x] Added or updated tests

Resolves #2463

**Changes**

`ModelPropertyHelper` builds the model with `newInstanceWithoutConstructor()`, so `initializeModelAttributes()` never runs and the class attributes it applies are never resolved. `getTable()` then falls back to the name derived from the class name, which has no columns, and every property access on a model that sets its table with `#[Table]` reports `property.notFound`. The `key`, `keyType` and `incrementing` arguments, and `#[Connection]`, are lost the same way.

Calling `initializeModelAttributes()` on the instance restores them. The call is guarded, as the method only exists on Laravel 13 and above.

**Breaking changes**

None.

### DIFF
--- a/src/Properties/ModelPropertyHelper.php
+++ b/src/Properties/ModelPropertyHelper.php
@@ -76,6 +76,8 @@ class ModelPropertyHelper
             return false;
         }
 
+        $this->initializeModel($modelInstance);
+
         if ($propertyName === $modelInstance->getKeyName()) {
             return true;
         }
@@ -97,6 +99,8 @@ class ModelPropertyHelper
         } catch (ReflectionException) {
             throw new ShouldNotHappenException();
         }
+
+        $this->initializeModel($modelInstance);
 
         $tableName = $modelInstance->getTable();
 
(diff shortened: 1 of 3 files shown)


sinemacula-ben:
Closing as a duplicate of #2500, which fixes the same root cause in more places.

PR #2533, opened 2026-08-19
fix: sync BuilderHelper::$passthru with Eloquent Builder's $passthru

### LINKED ISSUE #2532: BuilderHelper::$passthru has drifted from Eloquent Builder::$passthru — explain(), aggregate(), getRawBindings() and six more are typed as the builder
- Larastan Version: 3.10.0
- Laravel Version: 13.25.0
- PHPStan Version: 2.2.8
- PHP Version: 8.5.4

### Description

`Larastan\Larastan\Methods\BuilderHelper::$passthru` is a hand-maintained copy of `Illuminate\Database\Eloquent\Builder::$passthru`, and the two have drifted. Comparing the lists (Laravel 13.25.0), **nine** methods are present in Laravel's list but missing from Larastan's:

`aggregate`, `doesntExistOr`, `existsOr`, `explain`, `getRawBindings`, `implode`, `insertOrIgnoreUsing`, `numericAggregate`, `rawValue`

At runtime these take the passthru branch of `Eloquent\Builder::__call()` — `return $this->toBase()->{$method}(...)` — so they return the *query builder's* result (`explain()` → `Illuminate\Support\Collection`, `getRawBindings()` → `array`, `aggregate()` → `mixed`, …), never `$this`. But because they're missing from Larastan's list, `EloquentBuilderForwardsCallsExtension::findMethod()` falls through to the fluent branch and types them as **the builder itself**.

Spot-checked three of the nine with `dumpType()` (level 5, plain Larastan config, no strict rules):

```
PassthruBuilder.php:30:Dumped type: LarastanRepro\PassthruBuilder [identifier=phpstan.dumpType]   // $this->explain()
PassthruBuilder.php:31:Dumped type: LarastanRepro\PassthruBuilder [identifier=phpstan.dumpType]   // $this->getRawBindings()
PassthruBuilder.php:32:Dumped type: LarastanRepro\PassthruBuilder [identifier=phpstan.dumpType]   // $this->aggregate('max', ['id'])
```

The user-visible consequence is a false positive on correct code:

```
PassthruBuilder.php:25:Method LarastanRepro\PassthruBuilder::explainQuery() should return
Illuminate\Support\Collection<int, mixed> but returns LarastanRepro\PassthruBuilder. [identifier=return.type]
```

```php
<?php

declare(strict_types=1);

namespace LarastanRepro;

class Post extends Model
{
}
```

```php
<?php

declare(strict_types=1);

namespace LarastanRepro;

use function PHPStan\dumpType;

/**
 * @extends Builder<Post>
 */
class PassthruBuilder extends Builder
{
    /**
     * explain() is in Eloquent Builder's $passthru: at runtime it returns
     * Illuminate\Support\Collection, so this method is correct — but it is reported.
     *
     * @return Collection<int, mixed>
     */
    public function explainQuery(): Collection
    {
        return $this->explain();
    }

    public function dumps(): void
    {
        dumpType($this->explain());
        dumpType($this->getRawBindings());
        dumpType($this->aggregate('max', ['id']));
    }
}
```

`phpstan.neon`:

```neon
includes:

parameters:
    level: 5
    paths:
        - src
```

### Expected behaviour

Passthru methods should be typed as what the underlying query-builder method returns — the same treatment `count()`, `exists()`, `toSql()`, etc. already get.

### Suggested fix

Either sync the list, or derive it from Laravel's own property so it cannot drift again:

```php
$passthru = (new \ReflectionClass(\Illuminate\Database\Eloquent\Builder::class))
    ->getDefaultProperties()['passthru'];
```

(Laravel's `__call()` compares `strtolower($method)` against that list.) Deriving it would also absorb per-Laravel-version differences automatically — `BuilderHelper`'s constructor already has to version-gate `getCountForPagination` by hand today. Happy to send a PR either way.

Distinct from #2531 (which is about the return-type model for *fluent* forwarded calls and involves `: static` declarations): this one mistypes results even when no custom scope or `static` return type is involved. Found while preparing that report.

alexkart:
- [x] Added or updated tests

Resolves #2532

**Changes**

`BuilderHelper::$passthru` had drifted from `Illuminate\Database\Eloquent\Builder::$passthru`: nine methods present in Laravel's list were missing — `aggregate`, `doesntExistOr`, `existsOr`, `explain`, `getRawBindings`, `implode`, `insertOrIgnoreUsing`, `numericAggregate`, `rawValue`. Because they were missing, `EloquentBuilderForwardsCallsExtension` typed calls to them as the builder itself, while at runtime they take `__call()`'s passthru branch and return the query builder's result (`explain()` → `Illuminate\Support\Collection`, `implode()` → `string`, `insertOrIgnoreUsing()` → `int`, …).

This PR synchronizes the list and extends the passthru type tests with assertions for all nine.

No `LARAVEL_VERSION` gating is needed (unlike `getCountForPagination`): all nine are present in Eloquent's `$passthru` in every supported Laravel version — verified against v11.44.2, v12.4.1, and v13.25.0.

**Breaking changes**

None intended. Code that relied on the previous (incorrect) builder return type for these nine methods will surface new PHPStan errors, but the old type was wrong — chaining builder methods on e.g. `explain()` would fail at runtime.

**Local test note**

`GeneralTypeTest`: all nine new assertions pass. The pre-existing `passthru-l12-15.php:10` case (`getCountForPagination`) fails when the class is run *in isolation* (`vendor/bin/phpunit tests/Type/GeneralTypeTest.php`) — on clean `3.x` HEAD as well, so it is unrelated to this change. Cause: `getCountForPagination` is appended in `BuilderHelper`'s constructor behind `defined('LARAVEL_VERSION')`, but PHPStan instantiates the service before `bootstrap.php` runs in a fresh process (verified by instrumenting the constructor: `defined === false` there, while `dumpType(LARAVEL_VERSION)` later in the same analysis shows `'13.26.1'`). Full-suite runs pass because an earlier test class's container has already executed the bootstrap, and `define()` is process-wide. Filed separately as #2534 — the nine entries added here live in the default property and are not affected by that ordering.


canvural:
Thank you!

PR #2535, opened 2026-08-19
fix: resolve getCountForPagination passthru at call time

### LINKED ISSUE #2534: getCountForPagination passthru handling never applies in real analyses — the LARAVEL_VERSION gate in BuilderHelper::__construct() runs before bootstrap.php defines it
- Larastan Version: 3.10.0
- Laravel Version: 13.25 / 13.26
- PHPStan Version: 2.2.8 (identical under 2.2.7)
- PHP Version: 8.5.4

### Description

`BuilderHelper::__construct()` appends `getCountForPagination` to `$passthru` behind a `defined('LARAVEL_VERSION')` gate. `LARAVEL_VERSION` is defined by Larastan's `bootstrap.php` (via `bootstrapFiles`) — but PHPStan instantiates the container services **before** the bootstrap files run in a fresh process, so the gate is always false at construction time and the append never happens.

Verified by instrumenting the constructor with a `file_put_contents(...)` probe: at construction time `defined('LARAVEL_VERSION') === false`, while `dumpType(LARAVEL_VERSION)` later **in the same analysis** prints `'13.26.1'` — the constant arrives after the service already exists.

The consequence is user-facing, not just a test quirk. In an ordinary `vendor/bin/phpstan analyse` run of a real Laravel 13.25 application (one process, one container):

```php
use function PHPStan\dumpType;

dumpType(User::query()->getCountForPagination()); // Illuminate\Database\Eloquent\Builder<App\User>  ← wrong
dumpType(User::query()->count());                 // int<0, max>                                      ← correct (default-property entry)
```

So the Laravel ≥ 12.15 passthru handling of `getCountForPagination` is effectively dead for end users, while every entry in the `$passthru` *default property* works fine.

### Why the test suite doesn't catch it

`tests/Type/data/passthru-l12-15.php:10` asserts `int<0, max>` and passes in CI — but only because the full PHPUnit run creates earlier test classes' PHPStan containers first, whose bootstrap files already executed; `define()` is process-wide, so by the time `GeneralTypeTest`'s container instantiates `BuilderHelper`, the constant exists. Running the class in isolation on clean `3.x` HEAD fails that assertion:

```
vendor/bin/phpunit tests/Type/GeneralTypeTest.php
...
Expected type int<0, max>, got type Illuminate\Database\Eloquent\Builder<App\User>
in tests/Type/data/passthru-l12-15.php on line 10.
```

The same mistype reproduces via the CLI against the test config (`vendor/bin/phpstan analyse -c tests/Type/data/config-with-migrations.neon --level 1 <file with the dumpType above>`).

Not a PHPStan regression: 2.2.7 and 2.2.8 behave identically.

### Possible directions

- Evaluate the version lazily where `$passthru` is consulted (call time), instead of at construction time.
- Or gate on `\Illuminate\Foundation\Application::VERSION` — a class constant that needs no bootstrap. (Caveat: Lumen's `version()` string differs; the current `define`-based approach shares that caveat, since `bootstrap.php` uses `$app->version()`.)

Found while preparing #2533, which is unaffected (its entries go into the default property); the failure is also described in that PR's "Local test note".

lazerg:
- [x] Added or updated tests

Resolves #2534

**Changes**

`BuilderHelper::__construct()` appended `getCountForPagination` to `$passthru` behind a `defined('LARAVEL_VERSION')` check, but PHPStan builds its container before `bootstrap.php` defines that constant, so the check was always false and the method never made it into the list. In a real `phpstan analyse` run `User::query()->getCountForPagination()` was therefore typed as `Illuminate\Database\Eloquent\Builder<App\User>` instead of `int<0, max>`.

The version check now runs in a new `getPassthru()` method, at the point where the list is actually consulted, so the constant is already there. Both call sites read through it.

The existing `tests/Type/data/passthru-l12-15.php` assertion only passed because earlier test classes in the same process had already booted their own bootstrap files. Running `vendor/bin/phpunit tests/Type/GeneralTypeTest.php` on its own fails on `3.x` and passes with this change.

**Breaking changes**

None. The `$passthru` property keeps its current contents; anything reading it directly behaves as before.

### DIFF
--- a/src/Methods/BuilderHelper.php
+++ b/src/Methods/BuilderHelper.php
@@ -76,13 +76,21 @@ class BuilderHelper
         private bool $checkProperties,
         private MacroMethodsClassReflectionExtension $macroMethodsClassReflectionExtension,
     ) {
+    }
+
+    /** @return string[] */
+    public function getPassthru(): array
+    {
         // @phpstan-ignore-next-line
         if (! defined('LARAVEL_VERSION') || version_compare(LARAVEL_VERSION, '12.15.0', '<')) {
-            return;
+            return $this->passthru;
         }
 
         // @phpstan-ignore-next-line
-        $this->passthru[] = 'getCountForPagination';
+        $passthru   = $this->passthru;
+        $passthru[] = 'getCountForPagination';
+
+        return $passthru;
     }
 
     public function dynamicWhere(
@@ -236,7 +244,7 @@ class BuilderHelper
 
         $queryBuilderReflection = $this->reflectionProvider->getClass(QueryBuilder::class);
 
-        if (in_array($methodName, $this->passthru, true)) {
+        if (in_array($methodName, $this->getPassthru(), true)) {
             return $queryBuilderReflection->getNativeMethod($methodName);
         }
 
--- a/src/Methods/EloquentBuilderForwardsCallsExtension.php
+++ b/src/Methods/EloquentBuilderForwardsCallsExtension.php
@@ -139,7 +139,7 @@ final class EloquentBuilderForwardsCallsExtension implements MethodsClassReflect
 
         $parametersAcceptor = $ref->getVariants()[0];
 
-        if (in_array($methodName, $this->builderHelper->passthru, true)) {
+        if (in_array($methodName, $this->builderHelper->getPassthru(), true)) {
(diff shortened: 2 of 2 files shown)


canvural:
Thank you!

PR #2536, opened 2026-08-20
feat: add throw types for facade calls

marickvantuil:
Facade calls now report the throws of the underlying method, not just the `RuntimeException` from `Facade::__callStatic()`. Both get unioned.

PHPStan previously only knew about the `__callStatic()` throw, so a catch for the method's real exception got flagged as dead:

```php
use Illuminate\Support\Facades\View;

try {
    return View::first(['home', 'welcome']);
} catch (InvalidArgumentException $e) {
    // Illuminate\View\Factory::first() has @throws InvalidArgumentException,
    // but PHPStan reported: Dead catch - InvalidArgumentException is never
    // thrown in the try block. (catch.neverThrown)
    return null;
}
```

Adds `FacadeThrowTypeExtension` and an integration test.

- [x] Added or updated tests

**Changes**

- Facade calls expose the underlying method's real throws, so a valid catch is no
  longer flagged as a dead catch.

**Breaking changes**

Nothing API-wise. Facade calls now carry real throw types, so PHPStan may flag things it missed before: a catch for an exception that's no longer thrown, or a `@throws` that's now too wide.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -134,6 +134,10 @@ services:
         tags:
             - phpstan.broker.methodsClassReflectionExtension
     -
+        class: Larastan\Larastan\Methods\FacadeThrowTypeExtension
+        tags:
+            - phpstan.dynamicStaticMethodThrowTypeExtension
+    -
         class: Larastan\Larastan\Methods\ManagersMethodsExtension
         tags:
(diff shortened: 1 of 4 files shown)


szepeviktor:
I like your fast retreat.

PR #2539, opened 2026-08-24
fix: skip redundant initial rewind when scanning directories

### LINKED ISSUE #2538: Migration/schema/view/config scans silently miss files on WSL2 / Docker Desktop bind mounts (9p `rewinddir` bug) — waves of false `property.notFound` errors
- Larastan Version: 3.10.0 (still present on current `3.x` HEAD)
- Laravel Version: 12.64.0

### Description

On projects that run PHPStan inside a Docker Desktop (WSL2 backend) container with the code bind-mounted from the Windows filesystem, Larastan silently reads only part of the `database/migrations` directory. In our case it saw 43 of 70 migration files — the dropped chunk included `…create_users_table…` — which produced 163 false `property.notFound` errors on Eloquent models that do not occur on CI (ext4) for the same commit.

Nothing fails loudly: no exception, no warning, just fewer files.

**Root cause**

It is not a Larastan logic bug — it is how PHP's SPL directory iterators interact with a long-standing 9p bug in WSL2 (also reachable through Docker Desktop's Windows file sharing):

1. `RecursiveDirectoryIterator` reads the first buffer of directory entries when the handle is constructed.
2. `foreach` then calls `rewind()` on the iterator.
3. On 9p/drvfs mounts, rewinding a partially read directory handle is silently ignored ([microsoft/WSL#5074](https://github.com/microsoft/WSL/issues/5074), [microsoft/WSL#12416](https://github.com/microsoft/WSL/issues/12416)), so iteration resumes *after* the already-buffered chunk — the first libc read buffer of dirents is lost.

This is exactly the failure mode reported (and closed as "not a PHP bug") in [php#80056](https://bugs.php.net/bug.php?id=80056) and [php#80227](https://bugs.php.net/bug.php?id=80227). Plain `readdir()`, `scandir()`, `glob()`, `ls`, `find` are all unaffected — only iterators that rewind a partially read handle lose entries.

**Evidence**

Real-world case — same container, same commit, same PHP binary:

| Location | Filesystem | Migration files seen | PHPStan errors |
| --- | --- | --- | --- |
| bind mount from `C:\` | 9p (drvfs) | 43 of 70 | 163 (all phantom) |
| byte-identical copy under `/tmp` | ext4/overlay | 70 of 70 | 0 (matches GitHub CI) |

Minimal repro (directory with N migration-style file names on a `C:\` bind mount, counted from inside the container):

| Reader | `php:8.3-cli-alpine` (musl), N=200 | `php:8.3-cli` (glibc), N=500 |
| --- | --- | --- |
| `scandir()` / `glob()` | 200 | 500 |
| `FilesystemIterator` | **179** | **157** |
| plain `readdir()` walk | 200 | 500 |

The dropped count is exactly one libc readdir buffer (musl reads directories in 2 KiB chunks ≈ 21 of these dirents; glibc in 32 KiB chunks ≈ 343). That means on Alpine-based images — a very common Laravel container base — the bug already triggers at ~21+ migrations; on Debian-based images at a few hundred files.

Environment: Windows 11 Pro (build 26200), WSL 2.6.3 (bug still present as of August 2026), Docker Desktop with WSL2 backend, project on NTFS bind-mounted into the container (`stat -f` inside the container reports `v9fs`).

**Affected code in Larastan**

All four directory scans use the same `iterator_to_array(new RegexIterator(new RecursiveIteratorIterator(new RecursiveDirectoryIterator(…))))` pattern:

- `Properties\MigrationHelper::getMigrationFiles()`
- `Properties\SquashedMigrationHelper::getSchemaFiles()`
- `Support\ViewFileHelper::getViews()`
- `Support\ConfigParser::getConfigFiles()`

So besides migrations, squashed schema dumps, Blade views and config files can silently go missing on such mounts (any directory with more than one read buffer of entries).

**Suggested fix**

Replace the SPL iterator chain with a plain recursive `readdir()` walk, which never rewinds a partially read handle and returns exactly the same files (same keys, same `SplFileInfo` values) on healthy filesystems. We have verified this end-to-end: with the walk in place, PHPStan over the 9p mount reports 0 errors, identical to CI.

Happy to submit a PR — the change is small, behavior-preserving, and unit-tested.

Docker Desktop on Windows is arguably the most common local dev setup for Laravel on Windows machines (Sail defaults to it), so this probably bites more people than the issue tracker suggests — the symptom ("hundreds of `property.notFound` errors locally, CI green") does not look like a filesystem problem at all.

```php
// Any Eloquent property access whose column comes from a migration in the
// dropped chunk, e.g.:
$user = \App\Models\User::query()->firstOrFail();
echo $user->email; // false positive: Access to an undefined property App\Models\User::$email.

// Standalone repro of the underlying failure (run inside a container with
// /demo bind-mounted from the Windows filesystem, 200 files in the directory):
count(scandir('/demo/tree'));                         // 202 — complete
iterator_count(new FilesystemIterator('/demo/tree')); // 179 — first readdir buffer lost
```

denkfabrik-li:
- [x] Added or updated tests

Resolves #2538

**Changes**

`MigrationHelper`, `SquashedMigrationHelper`, `ViewFileHelper` and `ConfigParser` all collect files via `iterator_to_array(new RegexIterator(new RecursiveIteratorIterator(new RecursiveDirectoryIterator(…))))`. SPL directory iterators read the first buffer of entries when the handle is constructed, and `foreach` then calls `rewind()`. On filesystems where rewinding a partially read directory handle is silently ignored — the 9p mounts used by WSL2 and Docker Desktop on Windows ([microsoft/WSL#5074](https://github.com/microsoft/WSL/issues/5074), [php#80056](https://bugs.php.net/bug.php?id=80056), [php#80227](https://bugs.php.net/bug.php?id=80227)) — the already-buffered chunk is lost, so migrations, schema dumps, views and config files silently go missing from the scan. The loss is exactly one libc directory read buffer — musl reads directories in 2 KiB chunks, so on Alpine-based images (a very common Laravel container base) the first ~21 files of a directory disappear; glibc's 32 KiB buffer pushes the threshold to a few hundred files. Measured on such a mount: `FilesystemIterator` returned 179 of 200 files (musl) and 157 of 500 (glibc), while `scandir()`/`glob()`/plain `readdir()` returned all of them.

Real-world impact: in a Laravel app analysed inside a Docker Desktop container with the project bind-mounted from `C:\`, Larastan saw only 43 of 70 migrations (the dropped chunk included `create_users_table`), producing 163 false `property.notFound` errors that do not occur on CI. A byte-identical copy of the tree on ext4 inside the same container analyses clean.

This PR replaces the four SPL scans with a shared `Support\DirectoryScanner::findFiles()` — a plain recursive `readdir()` walk that never rewinds a partially read handle:

- Same results on healthy filesystems: files matched by the same regex against the full pathname, returned as `array<pathname, SplFileInfo>` exactly like `iterator_to_array()` produced (a unit test asserts parity with the replaced SPL construct).
- On affected 9p mounts it sees every entry; verified end-to-end on the app above — with this walk, PHPStan over the 9p mount reports 0 errors, identical to CI.
- The bug itself is not portably reproducible in CI (it needs a 9p mount), hence the tests cover the walker's behavior (recursion, pattern matching, key/value shape, SPL parity) rather than the filesystem defect.

Note: `composer test:types` currently reports 3 pre-existing `switch.alwaysFalse` errors in `Properties/SchemaAggregator.php` on `3.x` HEAD with PHPStan 2.2.9 (the scheduled static-analysis workflow is failing for the same reason); unrelated to this change.

**Breaking changes**

None. On filesystems with working `rewinddir()` the scanner returns the same file set with the same array shape; on broken ones it returns the complete file set instead of a truncated one.

Note: rebased onto current `3.x` (8058c08). The pre-existing `switch.alwaysFalse` errors in `Properties/SchemaAggregator.php` that an earlier version of this note mentioned were fixed upstream by #2506 in the meantime — `composer test:types`, `test:unit` and `test:cs` are all clean on this branch now.


szepeviktor:
This is an interesting PR.
Thank you!

canvural (changes requested):
Thank you for the PR and for the detailed investigation into this issue.

Could we keep SPL and only skip the redundant initial `rewind()`? Moving down to `readdir()` means reimplementing behavior SPL already provides, including symlink handling, unreadable-directory errors, and iterator semantics.

Symfony Finder handles non-rewindable filesystems with a small `RecursiveDirectoryIterator` subclass that ignores only the first `rewind()` and restores normal behavior after `next()`: [Symfony fix](https://github.com/symfony/finder/commit/96c05af6da0fae519d3ffc57880d6b4ed1dc0d13).

We can do the same and use a subclass of the iterator instead of the iterator directly.

canvural (approved):
Thank you for the update. This approach looks good to me.

Have you test this change on the affected Windows/WSL2 or Docker Desktop setups?

canvural:
Thank you!

PR #2542, opened 2026-08-28
ci: refresh monicahq/monica e2e baseline

maks-oleksyuk:
The E2E workflow's `Test with monicahq/monica on PHP 8.3` job has been red because the checked-in baseline is stale against the pinned monica ref.

This commits the baseline verbatim from the workflow's own `Generate baseline` step (artifact from run [33120962148](https://github.com/larastan/larastan/actions/runs/33120962148)):

- **adds** `Call to static method route() on an unknown class Redirect.` (`class.notFound`) entries — `ManageContact/ContactMoveController`, `ManageJournals/{JournalController,PostController,SliceOfLifeController}` (7 occurrences total)
- **drops** the stale `Constant Collator::ON …` entry for `CollectionHelper.php` that no longer matches

No analyzer/src changes — baseline only. Unblocks E2E.

Note: the `Redirect` facade resolving to `class.notFound` looks like a real larastan gap (standard `Illuminate\\Support\\Facades\\Redirect` alias not being resolved for the monica app) and is worth a separate look; this PR just stops it from blocking the pipeline.

### DIFF
--- a/e2e/monicahq-monica.baseline.neon
+++ b/e2e/monicahq-monica.baseline.neon
@@ -13,6 +13,12 @@ parameters:
 			path: ../../e2e/app/Domains/Contact/Dav/Web/Backend/CardDAV/CardDAVBackend.php
 
 		-
+			message: '#^Call to static method route\(\) on an unknown class Redirect\.$#'
+			identifier: class.notFound
+			count: 1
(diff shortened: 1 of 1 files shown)


szepeviktor:
Щиро дякую, Максе!

Issue #2474, opened 2026-04-13
Builder<T>::get() loses abstract template type when T is an unresolved @template parameter — collection-of<TModel> collapses to Model bound


alexkart:
- Larastan Version: 3.9.5
- Laravel Version: 13.x
- PHPStan Version: 2.1.47
- PHP Version: 8.3

### Description

When a generic helper method accepts `class-string<T>` (where `T of Model`) and executes the Eloquent builder pipeline on it, the resulting `Collection` loses the abstract template type `T` and collapses to its bound `Model`. All methods stubbed with `collection-of<TModel>` are affected — `get()`, `findMany()`, `find($array)`, `hydrate()`, `fromQuery()` — so there is no template-preserving path for a caller holding `class-string<T>` to produce a `Collection<int, T>` through the normal API.

Historically this was worked around with `/** @var Collection<int, T> $models */` narrowing. **PHPStan 2.1.47 now rejects that workaround** under strict/bleeding-edge checks because `Collection`'s `TModel` is invariant — see PRs [phpstan-src#5400](https://github.com/phpstan/phpstan-src/pull/5400) and [phpstan-src#5437](https://github.com/phpstan/phpstan-src/pull/5437). Projects on `level: max` + `phpstan-strict-rules` + `bleedingEdge.neon` now have no clean escape hatch except a manual rewrap through `(new $modelClass)->newCollection($models)`.

### Minimal reproduction

Full standalone repro repo: **https://github.com/alexkart/larastan-builder-template-repro**

```
git clone https://github.com/alexkart/larastan-builder-template-repro
cd larastan-builder-template-repro
composer install
vendor/bin/phpstan analyse
```

Inline code:

```php
<?php

declare(strict_types=1);

namespace Repro;

use Illuminate\Database\Eloquent\Collection as EloquentCollection;
use Illuminate\Database\Eloquent\Model;

final class Loader
{
    /**
     * @template T of Model
     *
     * @param class-string<T> $modelClass
     * @return EloquentCollection<int, T>
     */
    public function loadAllBroken(string $modelClass): EloquentCollection
    {
        $instance = new $modelClass;

        return $instance->newQuery()->get();
    }

    /**
     * Known workaround: re-wrap via newCollection() which goes through
     * HasCollection's standard PHPStan generic resolution.
     *
     * @template T of Model
     *
     * @param class-string<T> $modelClass
     * @return EloquentCollection<int, T>
     */
[...]

PR #2544, opened 2026-09-08
fix: keep static in the newQuery() method family on non-final models

yeapea:
- [x] Added or updated tests

I couldn't find an existing issue for this. I searched this repo for `newQuery static`, for the extension class name and for `Builder<static>`; the hits are about generic builders in general, and #2474 is the closest in spirit (a template parameter lost through `Builder<T>::get()`) but it's a different path.

**Changes**

`NewModelQueryDynamicMethodReturnTypeExtension` parameterizes the returned builder with the class it was called on rather than with `static`. Laravel annotates all six of these methods as `@return Builder<static>`, and `Builder` is invariant in its model parameter, so on a model that isn't final nothing declaring `Builder<static>` can be satisfied by `$this->newQuery()`:

```php
class Page extends Model
{
    /** @param Builder<static> $query */
    public function gate(Builder $query): void {}

    public function run(): void
    {
        $this->gate($this->newQuery());
    }
}
```

```
Parameter #1 $query of method Page::gate() expects
Illuminate\Database\Eloquent\Builder<static(Page)>,
Illuminate\Database\Eloquent\Builder<Page> given.
```

Adding `final` to the class makes the error go away, and that's what makes the constellation easy to miss: a codebase whose models are all final never meets it, while Laravel's own generators produce models without `final`.

The patch keeps `static` when the call was made on it and the model isn't final. Everything else stays where it was:

| call | before | after |
| --- | --- | --- |
| `$this->newQuery()`, non-final model | `Builder<Page>` | `Builder<static(Page)>` |
| `$this->newQuery()`, final model | `Builder<Page>` | unchanged |
| `$page->newQuery()` | `Builder<Page>` | unchanged |
| `Page::query()` | `Builder<Page>` | unchanged |

The final-model guard is load-bearing rather than decorative: PHPStan doesn't collapse `static(FinalPage)` back to `FinalPage` by itself, so without that condition a final model would start failing `Builder<static>` **and** `Builder<FinalPage>` parameters, which is worse than what it fixes.

**Breaking changes**

One, and it's the flip side of the fix rather than a separate defect. Under invariance the two annotations can't both work, so code that declares the concrete model and returns `$this->newQuery()` from inside a non-final model starts erroring:

```php
class Comment extends Model
{
    /** @return Builder<self> */
    public function ownQuery(): Builder
    {
        // Method Comment::ownQuery() should return Builder<Comment>
        // but returns Builder<static(Comment)>.
        return $this->newQuery();
    }
}
```

The fix in that code is to write `Builder<static>`, which is what Eloquent declares anyway. I left `UPGRADE.md` alone because it's organized per major version and this isn't one — say the word if you'd like an entry and where.

**What I ran**

PHP 8.5.8, `laravel/framework` 13.30.1, on `3.x` at 8f878f2.

- `composer test:cs` on the changed file: clean.
- `composer test:types`: no errors.
- The whole `tests/Type` directory: 1455 tests, 1477 assertions, green. With the source change reverted and the new assertions kept, exactly those two data files fail and nothing else does, so the other 1447 assertions don't move.
- `tests/Rules`, `tests/Unit`, `tests/Reflection` and `tests/Integration`: green, except that `ConfigCollectionRuleTest::testRule` is reported risky here ("did not remove its own error handlers") and `failOnRisky` turns that into a non-zero exit. It does the same at 8f878f2 without the patch, so it isn't this change.

### DIFF
--- a/src/ReturnTypes/NewModelQueryDynamicMethodReturnTypeExtension.php
+++ b/src/ReturnTypes/NewModelQueryDynamicMethodReturnTypeExtension.php
@@ -11,6 +11,7 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\MethodReflection;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
 use PHPStan\Type\ObjectType;
+use PHPStan\Type\StaticType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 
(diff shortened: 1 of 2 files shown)


canvural:
Thank you!

Issue #2329, opened 2025-08-02
Relationship Issue


jkdobariya:
- Larastan Version: 3.6
- Laravel Version: 11.45.1

### Description

Relationship already exists even after it's appearing error.

Relation 'instance' is not found in App\Models\Campaign model.                                                
         🪪  larastan.relationExistence  

### Laravel code where the issue was found

```php

final class Campaign extends Model
{
    use HasFactory;

    public function instance()
    {
        return $this->hasOne(Instance::class, 'id', 'instance');
    }
}
```
<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/55a9a7a7-48b8-4899-a6c7-ef286cf5f3a6" />

szepeviktor:
Hello @jkdobariya 👋🏻 

Try adding a docblock
https://github.com/larastan/larastan/blob/6431d010dd383a9279eb8874a76ddb571738564a/tests/application/app/User.php#L185-L186

Issue #2178, opened 2025-01-24
Model relation is not found if relation doesn't have a return type


LordSimal:
- Larastan Version: 3.0.2
- Laravel Version: 11.37.0

### Description

If the relation inside a model doesn't have a return Type like `HasMany` larastan doesn't detect it correctly and therefore reports the following error when trying to eager load a query.

### Laravel code where the issue was found

```php
// MainProduct.php 

    public function product_types()
    {
        return $this->hasMany(ProductType::class);
    }

// MainProductController.php

    public function view(int $id): View
    {
        $mainProduct = MainProduct::whereId($id)->with('product_types')->firstOrFail();

        return view('main_product/view', compact('mainProduct'));
    }
```

```
 ------ ------------------------------------------------------------------------ 
  Line   Http/Controllers/MainProductController.php                              
 ------ ------------------------------------------------------------------------ 
  27     Relation 'product_types' is not found in App\Models\MainProduct model.  
         🪪  larastan.relationExistence                                          
 ------ ------------------------------------------------------------------------ 
```

but after adding the return type to the relation it works

```php
// MainProduct.php 

    public function product_types(): HasMany
    {
        return $this->hasMany(ProductType::class);
    }
```

calebdw:
Hello!

This is by design, the only way we know that a model method is a relationship is if the return is a child of Relation. Since not all model methods are accessible as properties we can't just assume that if a method exists and there's no return type then it's a relationship.

Thanks!

LordSimal:
Thanks for the quick answer! I can definitely understand, that this is the most straight forward way of narrowing the methods inside the model.

Just thought (technically) you could check the return statement inside the method to see if it uses a relation method (or that variable returned is an instance of relation) as a fallback if no return type is present.

calebdw:
> Just thought (technically) you could check the return statement inside the method to see if it uses a relation method (or that variable returned is an instance of relation) as a fallback if no return type is present.

It's not as easy as it might sound---you have to parse the file (io read) with PHP-Parser and walk through all of the nodes until you find the correct method / return statement node, then you have to convert the PHP-Parser node back into a PHPStan type. It's a slow and complicated process and it was [removed in 3.x for a reason](https://github.com/larastan/larastan/blob/3.x/UPGRADE.md#correct-return-types-for-model-relation-methods).

[...]

PR #2549, opened 2026-09-15
fix: don't fall back to the declaring model when a relation path can't be followed

MizouziE:
- [x] Added or updated tests

**Changes**

Since 3.12.0, a nested `whereHas()` over a dotted relation path reports a relation that definitely exists:

```
Relation 'schedules' is not found in App\Models\FtpConnection model.  larastan.relationExistence
```

All it takes is relation methods with a bare native return type and no generic `@return`:

```php
class FtpConnection extends Model
{
    public function transfers(): HasMany
    {
        return $this->hasMany(Transfer::class);
    }
}

FtpConnection::query()->whereHas('transfers.task', function (Builder $query): void {
    $query->whereHas('schedules', function (Builder $query): void {
        $query->where('monday', '=', true);
    });
});
```

That code is fine at runtime. `QueriesRelationships::hasNested()` only hands your callback to the last segment, and `has()` types it as `(\Closure(\Illuminate\Database\Eloquent\Builder<TRelatedModel>): mixed)|null`, so the callback is given a builder for `Task` and `Task::schedules()` is right there.

`stubs/common/EloquentBuilder.stub` types that callback as `builder-of<TModel, TRelation>`, so the path gets resolved by `BuilderOfType`. A relation without generics resolves `TRelatedModel` to its bound, `Model`. That is harmless as the last segment but fatal as an intermediate one, because the next name gets looked up on `Model`, is not found, and the whole path resolves to `null`. `getResult()` treated that exactly like a relation that is genuinely absent and fell back to `?? $this->type`, so the callback came out as `Builder<FtpConnection>`, which going by `has()` is the one model it can never be. Note that `transfers()` is still recognised as a relationship, since its return type is a child of `Relation`, which is [the rule Larastan goes by](https://github.com/larastan/larastan/issues/2178#issuecomment-2612393490). Only its related model is unknown.

`BuilderOfType` now tells the two failures apart:

* Absent from a model we do know: unchanged. `RelationExistenceRule` already reports it, and falling back to the declaring model is [documented behaviour](https://github.com/larastan/larastan/blob/79091ad281b47871ec2c0f7d1aa61b042a1e15fc/docs/custom-types.md#builder-of) pinned by `testRelationshipFallbacks()`.
* Failed on a model we do not know: now `Builder<Model>`. "Do not know" means abstract, because an abstract model is never the one being queried, so not finding a relationship on it tells you nothing about the concrete subclass, and `Model` is just the extreme case. That is the same line `ModelRuleHelper::findModelReflectionFromType()` already draws, so the two now agree. It also covers a relation annotated as `HasMany<SomeAbstractBase, $this>`, which had the same problem.

`Builder<Model>` is already what a single unannotated segment produces, so this closes an inconsistency rather than inventing a third behaviour. Adding a segment to a path cannot make Larastan more sure of the model than it was one segment earlier.

Digging `Transfer` back out of `$this->hasMany(Transfer::class)` is [off the table by design](https://github.com/larastan/larastan/blob/79091ad281b47871ec2c0f7d1aa61b042a1e15fc/UPGRADE.md#correct-return-types-for-model-relation-methods) and was [turned down](https://github.com/larastan/larastan/issues/2178#issuecomment-2612739232) when it came up again, so `Builder<Model>` is the honest answer. Annotating the relation is the standing advice on these reports, in [#2178](https://github.com/larastan/larastan/issues/2178#issuecomment-2612739232) and [#2329](https://github.com/larastan/larastan/issues/2329#issuecomment-3146261359), and [`AddGenericReturnTypeToRelationsRector`](https://github.com/driftingly/rector-laravel/blob/main/docs/rector_rules_overview.md#addgenericreturntypetorelationsrector) will do it for you, as also [suggested there](https://github.com/larastan/larastan/issues/2178#issuecomment-2612782764).

The path walk now lives in its own method, `followRelationPath()`, and the combined result is memoised, since `RelationshipQueryCallbackExtension::getEagerCallbackType()` asks one `BuilderOfType` for the relation type and then, for `withWhereHas()` and `withWhereRelation()`, the builder type as well. Same walk as before, it just hands back the unknown-model flag per path rather than accumulating it as it goes.

**Tests**

Four new fixture models in `App\BareRelations`. The three with relations have to keep the bare return type shape, because annotating them hides the bug.

* `RelationExistenceRuleTest::testBareRelationPaths()` covers the reproducer, which should report nothing, next to a positive control on annotated models that must still report.
* `tests/Type/data/relationship-query-callbacks.php` pins the callback parameter for the single segment case, the dotted path and a nested `whereHas()`, directly below the existing annotated `'posts.comments'` case.
* `tests/Type/data/builder-of-type.php` gets `testBareRelationshipBuilders()` and `testAbstractRelatedModels()`, which pin both failure modes on the same models and check a path still walks *through* an abstract model rather than stopping at one. `testRelationshipUnions()` gets `$partlyUnknown`, since an unknown alternative is discarded like a missing one once another alternative resolves.

One existing expectation changed. `builder-of<Comment, 'commentable.comments'>` was asserting `Builder<App\Comment>` on the line directly below `builder-of<Comment, 'commentable'>` asserting `Builder<Model>`, so it contradicted its own neighbour.

**Breaking changes**

Nothing changes in configuration, the public API or any stub signature, and I don't think this needs an `UPGRADE.md` entry.

`builder-of<TModel, TRelation>` where the relation names are known but the path cannot be followed now resolves to `Builder<Illuminate\Database\Eloquent\Model>` instead of `Builder<TModel>`. That covers paths through a relation with no generics, and through one annotated to an abstract model. Either way it only affects paths which used to resolve to a model the callback never receives.

Worth flagging: with `checkModelProperties: true`, column checks inside one of those callbacks will now name `Illuminate\Database\Eloquent\Model` rather than the declaring model. Neither class is the one being queried, and annotating the relation sorts out both.


canvural:
Thank you!

PR #2551, opened 2026-09-18
fix: flag Octane container injection in arrow functions

arpitjain099:
The rule gates on `$args[1]->value instanceof Node\Expr\Closure`, but in PHP-Parser `ArrowFunction` isn't a subclass of `Closure` - they're siblings under `Expr`. So `fn ($app) => new Service($app)` returns early and neither detection path runs.

Every case in `ContainerInjection.php` used `function`, so nothing covered this. I added six arrow-function equivalents and the test still passed with the expectations untouched: all six were silent. With the fix, four are reported, matching their closure twins.

`getParams()` and `getStmts()` exist on both node types, so only the gate and the `checkForThisAppUsage` signature needed widening.

`composer test:unit` 1892 pass, `test:cs` 74/74, `test:types` no errors.


canvural:
Thank you!

PR #2552, opened 2026-09-19
perf: cache Attribute ObjectType instances in ModelPropertyHelper

haratake22:
- [x] Added or updated tests
- [ ] Documented user facing changes

**Changes**

This PR optimizes `ModelPropertyHelper` by caching `ObjectType` and `GenericObjectType` instances for `Illuminate\Database\Eloquent\Casts\Attribute`. 

Previously, `new ObjectType(Attribute::class)` and `new GenericObjectType(Attribute::class, ...)` were instantiated repeatedly on every property accessor check (`hasAccessor` / `getAccessor`), triggering unnecessary class reflection lookups and object allocations.

**Performance Impact / Benchmark Results**

Running `phpstan analyse` on Larastan's codebase (169 files):

| Metric | Before | After | Improvement |
| :--- | :--- | :--- | :--- |
| **Execution Time** | 83.068s | 79.075s | **~4.0s faster (-4.81%)** ⚡ |
| **Analysis Errors** | 0 errors | 0 errors | All green ✅ |

**Breaking changes**
None. This is an internal performance optimization.

PR #2553, opened 2026-09-19
perf: cache ObjectType instances in helper extensions and BuilderHelper

haratake22:
- [x] Added or updated tests

**Changes**

Follow-up to #2552.

This PR optimizes helper return type extensions and `BuilderHelper` by caching type objects and reflection lookups:

1. **Helper Dynamic Return Type Extensions (`src/ReturnTypes/Helpers/`)**:
   - Avoids repeatedly instantiating `ObjectType` / `GenericObjectType` on every helper call by caching them:
     - `AppExtension`: Caches `Application` object type.
     - `AuthExtension`: Caches `Factory` and combined guard types.
     - `CollectExtension`: Caches empty collection generic object type.
     - `ResponseExtension`: Caches `ResponseFactory` and `Response` object types.
     - `StrExtension`: Caches `Stringable` object type and `MixedType`.
     - `ValidatorExtension`: Caches `Factory` and combined validator types.

2. **`BuilderHelper` (`src/Methods/BuilderHelper.php`)**:
   - Caches `determineBuilderName` results per model class to avoid repeated reflection lookups (`newEloquentBuilder`, attribute inspections, etc.).

**Performance Impact**

In Laravel applications with frequent helper calls and Eloquent model method calls, this reduces repeated object allocations, GC pressure, and reflection lookup overhead.

**Breaking changes**
None.


canvural:
Thank you!

PR #2555, opened 2026-09-20
fix: keep builder return types when model docblocks widen them

### LINKED ISSUE #2554: 3.12: model `@method` docblocks now override the builder return type, yielding `Model|Builder<Model>`
### Summary

Upgrading larastan from **3.10.0** to **3.12.x** changes the inferred return type of `Model::query()` / `Model::newQuery()` from `Builder<Model>` to the union `Model|Builder<Model>` when the model carries the `@method` docblocks that `barryvdh/laravel-ide-helper` generates.

In 3.10 larastan's own return type extension won over the docblock. In 3.12 the docblock wins, so the union leaks into every call site and produces `return.type` / `argument.type` errors at level 10 in code that was previously clean.

### Reproduction

A model with stock ide-helper output — note the union in the generated annotations, which is what `ide-helper:models` has always written:

```php
/**
 * @method static \Illuminate\Database\Eloquent\Builder<static>|Foo newModelQuery()
 * @method static \Illuminate\Database\Eloquent\Builder<static>|Foo newQuery()
 * @method static \Illuminate\Database\Eloquent\Builder<static>|Foo query()
 * @mixin \Eloquent
 */
class Foo extends Model {}
```

Then:

```php
\PHPStan\dumpType(Foo::query());
\PHPStan\dumpType((new Foo)->newQuery());
```

### Result

Same code, same PHPStan (**2.2.14**) in both runs — only larastan differs:

| larastan | Dumped type |
| --- | --- |
| 3.10.0 | `Illuminate\Database\Eloquent\Builder<Foo>` |
| 3.12.2 | `Foo\|Illuminate\Database\Eloquent\Builder<Foo>` |

### Expected

`Builder<Foo>`, as in 3.10 — `query()` and `newQuery()` return a builder, never the model, so the union is not a type either method can actually produce.

### Impact

On a level-10 codebase with no baseline this surfaced 21 errors across ordinary query helpers, all of the shape:

```
Method X::population() should return Builder<Foo> but returns Foo|Builder<Foo>.
Anonymous function should return Builder<Foo> but returns Foo|Builder<Foo>.
```

Chaining recovers (`newQuery()->where(...)` is `Builder<Foo>` again), so it only bites where a `query()` result is returned or passed directly.

Scope annotations are affected the same way, since ide-helper writes them in the same union form (`@method static Builder<static>|Foo someScope()`), so regenerating the docblocks is not a workaround.

### Versions

- larastan 3.12.2 (also reproduces on 3.12.0/3.12.1)
- phpstan 2.2.14
- Laravel 13.32.0, PHP 8.4
- barryvdh/laravel-ide-helper 3.7 (default config)

lazerg:
- [x] Added or updated tests

Resolves #2554

**Changes**

laravel-ide-helper writes `@method static Builder<static>|Foo query()` onto the models it annotates, and that docblock shadows Larastan's stub. Once these methods moved from return type extensions to `builder-of` stubs, `Foo::query()` and `$foo->newQuery()` on such a model started coming back as `Foo|Builder<Foo>`, where 3.10 gave `Builder<Foo>`. The model half of that union is not something either method can return, so it leaks into every call site that returns or passes the result directly.

`ModelDynamicStaticMethodReturnTypeExtension` now also accepts a return type that unions the builder with the model, and a new extension does the same for the `newQuery()` family on instance calls. Both back off when the docblock did not widen the type, so the stubs keep driving everything else.

**Breaking changes**

None.


canvural:
I'm not sure if I want to make a fix for a wrong docblock. Clearly the doc block type is wrong here.

lazerg:
You're right that the annotation is wrong. Nobody types it by hand, though — `ide-helper:models` writes it.

`ModelsCommand` emits `Builder<static>|Model` for `query`/`newQuery`/`newModelQuery` ([source](https://github.com/barryvdh/laravel-ide-helper/blob/master/src/Console/ModelsCommand.php#L728-L737)), for every scope, and for custom builder methods. It has done that for years, so regenerating the docblocks does not remove the union. Any project that runs the command carries it.

Until 3.12 the return type extension won and the union never reached the call site. The stub now wins instead, so level 10 code that returns a builder breaks on an upgrade.

Happy to close this if you would rather see it fixed in ide-helper. I will leave the branch as is until you decide.

PR #2558, opened 2026-09-22
fix: check model properties in forceCreateQuietly()

midhunmonachan:
- [x] Added or updated tests

Related to #676.

**Changes**

Adds the missing `forceCreateQuietly()` stub so Larastan checks attribute names, just as it does for `forceCreate()` and `createQuietly()`.

The regression test covers query builder and static model calls, valid and invalid attributes, and calls without arguments. It fails without the stub and passes with it.

**Breaking changes**

Invalid attribute names passed to `forceCreateQuietly()` now produce an error when model-property checking is enabled.

**Tests**

Passed on PHP 8.5 and Laravel 13.32:

- `composer test:cs`
- `composer test:types`
- `composer test:unit` (1,919 tests, 2 skipped)


canvural:
Thank you!

PR #2559, opened 2026-09-22
feat: add a rule to check the variadic types in throw_if helpers

wgriffioen:
- [X] Added or updated tests

**Changes**

Adds `CheckThrowIfArgumentTypesCompatibleWithClassConstructorRule`, which checks the variadic arguments of `throw_if()` / `throw_unless()` against the constructor of the exception class they name.

```php
throw_if($condition, ExampleException::class, 'bar', 1);
```

Laravel instantiates that class with the remaining arguments, so this is a constructor call, but nothing checks it today:

```
Parameter #1 $foo of exception class ExampleException constructor expects int in throw_if(), string given.
Parameter #2 $bar of exception class ExampleException constructor expects string in throw_if(), int given.
```

This mirrors `CheckDispatchArgumentTypesCompatibleWithClassConstructorRule`, which solves the same problem for `Job::dispatch()`.

*Why a rule and not a stub.* The helper's signature binds the variadic arguments to `TParams`, not to the constructor of the class-string in `$exception`. There is no PHPDoc syntax that expresses "these varargs must match the constructor of the class named by another argument", so this cannot be fixed in `stubs/`.

*The condition narrows the arguments.* The exception is only constructed when the condition holds, so the arguments are checked in `filterByTruthyValue()` of the condition (`filterByFalseyValue()` for `throw_unless()`). Without this, the very common shape below would be reported even though the constructor can never receive `null`:

```php
throw_if(null !== $action, MissingRequiredAttributesException::class, $action);
```

I ran the rule over an internal Laravel application with 117 `throw_if()` / `throw_unless()` call sites: **zero false positives**. Before the scope narrowing was added, that same codebase produced 2.

*Also reported*, because the parameters silently never reach a constructor:

- `larastan.throwIf.ignoredParameters` — the second argument is a string that is not a class name, so Laravel throws `RuntimeException($string)` and drops the parameters.
- `larastan.throwIf.notThrowable` — the named class does not implement`Throwable`, so `throw` fatals.

*Not checked*: closures, already instantiated exceptions, exception classes that are not a single known class-string, unpacked leading arguments, and interfaces or abstract classes (not instantiable at all, which is a different defect).
Named arguments are supported, including `condition:` / `exception:` in either order and named constructor parameters forwarded through the variadic.

The rule is **disabled by default** (`checkThrowIfArgumentTypes`), following `noImplicitQueryBuilderCall`, so it does not introduce new errors for existing users on a minor release.

**Breaking changes**

None, as the added rule is opt-in


src/Rules/CheckThrowIfArgumentTypesCompatibleWithClassConstructorRule.php
```diff
@@ -0,0 +1,236 @@
...
+ * relationship between the class-string and the variadic arguments cannot be
+ * expressed in PHPDoc, so the call is otherwise unchecked.
+ *
+ * @implements Rule<FuncCall>
+ */
+class CheckThrowIfArgumentTypesCompatibleWithClassConstructorRule implements Rule
+{
+    public function __construct(
+        private ReflectionProvider $reflectionProvider,
+        private FunctionCallParametersCheck $check,
```
canvural, line 41: We should use `NodeCallbackInvoker` instead. Read more about it [here](https://phpstan.org/developing-extensions/rules#invoking-rules-for-synthetic-nodes).

src/Rules/CheckThrowIfArgumentTypesCompatibleWithClassConstructorRule.php
```diff
@@ -0,0 +1,236 @@
...
+        $className = $constantStrings[0]->getValue();
+
+        if (! $this->reflectionProvider->hasClass($className)) {
+            if ($constructorArgs === []) {
+                return [];
+            }
+
+            return [
+                RuleErrorBuilder::message(sprintf(
+                    'Function %s() is called with a string that is not a class name, so a RuntimeException is thrown with that string as its message and the given parameters are ignored.',
```
canvural, line 94: This message is too long for my taste. And, it's already typed as `class-string`, so PHPStan would report it anyway, no?

src/Rules/CheckThrowIfArgumentTypesCompatibleWithClassConstructorRule.php
```diff
@@ -0,0 +1,236 @@
...
+            return [
+                RuleErrorBuilder::message(sprintf(
+                    'Class %s given to %s() does not implement Throwable.',
+                    $classReflection->getDisplayName(),
+                    $functionName,
+                ))
+                    ->identifier('larastan.throwIf.notThrowable')
+                    ->build(),
+            ];
+        }
```
canvural, line 114: Again, can't we type it as `class-string<\Throwable>` in stubs. Then PHPStan would take care of it.

src/Rules/CheckThrowIfArgumentTypesCompatibleWithClassConstructorRule.php
```diff
@@ -0,0 +1,236 @@
...
+
+        // An interface extending Throwable, or an abstract exception, cannot be
+        // instantiated at all. That is a different defect from the argument
+        // types this rule reports on, so it is left to other rules.
+        if ($classReflection->isInterface() || $classReflection->isAbstract() || ! $classReflection->hasConstructor()) {
```
canvural, line 119: `$classReflection->hasConstructor()` check is useless here, no? At this point it needs to be a Throwable class, and it'll always have constructor because every exception has a constructor.

src/Rules/CheckThrowIfArgumentTypesCompatibleWithClassConstructorRule.php
```diff
@@ -0,0 +1,236 @@
...
+     * Separates the exception argument from the parameters forwarded to its
+     * constructor, the way Laravel's helper does at runtime.
+     *
+     * Mirrors PHPStan's own handling of call_user_func(): the leading
+     * arguments may be named, and the remaining ones keep their names so they
+     * can be checked as named arguments of the constructor.
+     *
+     * @return array{Arg|null, Arg, list<Arg>}|null
+     */
+    private function splitArgs(FuncCall $node): array|null
```
canvural, line 200: Is this method needed? I think PHPStan exposes a shortcut for this. Like `$node->getArg('name')` or something like that. Can you check?

canvural (changes requested):

